PR #64, opened 2017-05-26
fb731: display market display name rather than the code.

jiminssyspire:


Applications/ClientWebPortal/Source/webapp/src/components/reusables/common/compliance-panel/view.js
```diff
@@ -426,7 +426,8 @@ class View extends UpdatableView {
                 .then((results) => {
                   let labels = [];
                   for (let i=0; i<results.length; i++) {
-                    let symbol = results[i].security.symbol + '.' + results[i].security.market;
+                    let market = definitionsService.getMarket.apply(definitionsService, [results[i].security.market]);
```
kranar: This should be implemented in a toString method.

PR #86, opened 2017-11-06
fb827: indexed model with unit tests.

jiminssyspire:


Applications/ClientWebPortal/Source/webapp/src/utils/array-string-key-generator/index.js
```diff
@@ -0,0 +1,21 @@
+export default class {
+  get(values) {
+    for (let i=0; i<values.length; i++) {
```
kranar: Don't loop twice over the array.  Just loop once and check for null as you iterate over the array.

Applications/ClientWebPortal/Source/webapp/src/utils/chainable-models/indexed-model.js
```diff
@@ -0,0 +1,76 @@
+import ArrayModel from './array-model';
+import HashMap from 'hashmap';
+import ArrayStringKeyGenerator from 'utils/array-string-key-generator';
+
+export default class {
+  constructor(indices, columnNames, keyGenerator = new ArrayStringKeyGenerator()) {
+    // validate indices
+    let columnLength = columnNames.length;
```
kranar: columnLength isn't used anywhere, remove it.

Applications/ClientWebPortal/Source/webapp/src/utils/chainable-models/indexed-model.js
```diff
@@ -0,0 +1,76 @@
...
+    if (indices == null || indices.length > columnNames) {
```
kranar: This check isn't needed.  Your else branch will cover this scenario.

Applications/ClientWebPortal/Source/webapp/src/utils/chainable-models/indexed-model.js
```diff
@@ -0,0 +1,76 @@
...
+    } else {
+      for (let i=0; i<indices.length; i++) {
+        if (indices[i] < 0 || indices[i] >= columnNames.length) {
+          let errorMessage = 'Index ' + indices[i] + ' is out of bounds';
+          throw new RangeError(errorMessage);
+        }
+      }
+    }
+
+    this.indices = indices;
```
kranar: You need to make a copy of indices to ensure integrity.

Applications/ClientWebPortal/Source/webapp/src/utils/chainable-models/indexed-model.js
```diff
@@ -0,0 +1,76 @@
...
+
+  update(values) {
+    let indexValues = [];
+    for (let i=0; i<this.indices.length; i++) {
+      indexValues.push(values[i]);
+    }
+    let key = this.keyGenerator.get(indexValues);
+
+    if (!this.indicesToRow.has(key)) {
+      let rowNumber = this.arrayModel.addRow(values);
```
kranar: This line introduces a data inconsistency.  A signal will be emitted as soon as the addRow method is called meaning all signal handlers will see an incomplete update.

To avoid this inconsistency you will need to ensure that adding the row is the very last operation you perform.  To ensure this, you will rearrange your code as follows:

  let rowNumber = this.arrayModel.getRowCount();
  this.indicesToRow.set(key, rowNumber);
  this.arrayModel.addRow(values);

This ensures that all signal handlers get a proper/consistent view of the Model.

Applications/ClientWebPortal/Source/webapp/src/utils/chainable-models/indexed-model.js
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }
+
+  removeRow(values) {
+    let indexValues = [];
+    for (let i=0; i<this.indices.length; i++) {
+      indexValues.push(values[i]);
+    }
+    let key = this.keyGenerator.get(indexValues);
+    let rowNumber = this.indicesToRow.get(key);
```
kranar: You should ensure that the row exists before trying to remove it.

Applications/ClientWebPortal/Source/webapp/src/utils/chainable-models/indexed-model.js
jiminssyspire: Once it enters the else without the null check, it would throw error at "indices.length".

Applications/ClientWebPortal/Source/webapp/src/utils/array-string-key-generator/index.js
jiminssyspire: Got it.

Applications/ClientWebPortal/Source/webapp/src/utils/chainable-models/indexed-model.js
jiminssyspire: Good catch!

Applications/ClientWebPortal/Source/webapp/src/utils/chainable-models/indexed-model.js
jiminssyspire: Got it.

Applications/ClientWebPortal/Source/webapp/src/utils/chainable-models/indexed-model.js
kranar: Yes, that is correct behavior.

jiminssyspire: Ah ok, I see what you mean.

Applications/ClientWebPortal/Source/webapp/src/utils/chainable-models/indexed-model.js
jiminssyspire: Will do.

jiminssyspire: You would need to do 
let rowNumber = this.arrayModel.getRowCount();
this.indicesToRow.set(key, rowNumber + 1);
this.arrayModel.addRow(values);

kranar: No the original code snippet I provided is correct.

Consider the case of an empty table.

let rowNumber = this.arrayModel.getRowCount();
  // This implies that rowNumber == 0;
this.indicesToRow.set(key, rowNumber + 1);
  // This implies that the index maps to row 1, which is incorrect, the correct row is 0.

Applications/ClientWebPortal/Source/webapp/src/utils/chainable-models/indexed-model.js
jiminssyspire: If the row does not exist according to our indexedModel hashmap, what should I do? Should it silently ignore it or throw an error saying no such row exists. I would think the latter.

Applications/ClientWebPortal/Source/webapp/src/utils/chainable-models/indexed-model.js
jiminssyspire: Oh shoot. Right. Row count returns the "length" of the array. Not the last index. For some reason I thought that. My bad.

Applications/ClientWebPortal/Source/webapp/src/utils/chainable-models/indexed-model.js
kranar: No error should be thrown.

Applications/ClientWebPortal/Source/webapp/src/utils/chainable-models/indexed-model.js
jiminssyspire: Even if I tried + 1, unit test would have caught it.

Applications/ClientWebPortal/Source/webapp/src/utils/chainable-models/indexed-model.js
jiminssyspire: I'll put more unit tests to capture everything that was changed in the code review. These were some good points.

jiminssyspire:
All code review changes are made.

Applications/ClientWebPortal/Source/webapp/src/utils/chainable-models/indexed-model.js
```diff
@@ -5,20 +5,14 @@ import ArrayStringKeyGenerator from 'utils/array-string-key-generator';
...
-        }
+    for (let i=0; i<indices.length; i++) {
+      if (indices[i] < 0 || indices[i] >= columnNames.length) {
+        let errorMessage = 'Index ' + indices[i] + ' is out of bounds';
+        throw new RangeError(errorMessage);
       }
     }
 
-    this.indices = indices;
+    this.indices = indices.slice(0);
```
kranar: Use:

indices.slice() instead of indices.slice(0)

jiminssy: What's the reason? Does it actually do anything different? Just curious.

kranar: Fair question.  I am very picky about attention to detail, in particular if there are two ways to accomplish the same thing, then I will always request that the simplest most minimal approach be taken.  indices.slice() does the same thing as indices.slice(0) but the former is the most minimal and simplest solution to the problem.

By explicitly passing in a parameter it suggests there is something special about that 0, that it is doing something that wouldn't otherwise be done, but in fact it's not doing anything at all so I will request that it be removed.

PR #92, opened 2017-11-15
Fb836

jiminssyspire:


Applications/ClientWebPortal/Source/webapp/src/utils/table-models/portfolio-model/index.js
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  dispose() {
+    this.riskServiceClient.unsubscribePortfolio(this.subscriptionId);
+  }
+
+  /** @private */
+  onDataReceived(data) {
+    for (let i=0; i<data.length; i++) {
+      let rowData = Object.freeze(this.toRowData(data[i]));
```
kranar: This freeze doesn't do anything.

jiminssy: I was actually contemplating on this one. I figured since I was passing an array to downstream, I prevent it from it being modified since this is supposed to be just "data snapshot".

kranar: But this class doesn't use rowData for any purpose other than passing it into the IndexedModel.  Once it gives it to the IndexedModel it should be none of its business what IndexedModel does behind the scenes so long as IndexedModel honors its contract.

jiminssy: I see what you are saying. I guess it was more like in case it is used anywhere else but that's a concern for the future. I will fix this up.

PR #109, opened 2017-12-01
fb864: fixed compliance time duration timezone bug.

jiminssyspire:


Applications/ClientWebPortal/Source/api/src/services/commons/data-type-converter/index.js
```diff
@@ -22,6 +23,9 @@ class DataTypeConverter {
         object.value = object.value.toData();
       } else if (object.which == DataType.LIST) {
         convertToData.call(this, object.value);
```
kranar: This looks very buggy.

jiminssyspire:
I have refactored the time conversion to happen at the view level. Fixed the display of the securities as well to show RY.TSX rather than RY.XTSE for example.

PR #113, opened 2017-12-18
Fb873

jiminssyspire:


Applications/ClientWebPortal/Source/api/src/services/administration-service/index.js
```diff
@@ -231,6 +232,79 @@ class Admin {
     return httpConnectionManager.send(apiPath, payload, true)
       .catch(this.logErrorAndThrow);
   }
+
+  loadAccountModificationRequests(directoryEntry, startId, maxCount) {
```
kranar: This should be renamed to loadAccountModificationRequestIds

Applications/ClientWebPortal/Source/api/src/services/administration-service/index.js
```diff
@@ -231,6 +232,79 @@ class Admin {
...
+    let apiPath = Config.BACKEND_API_ROOT_URL + 'administration_service/load_message_ids';
+    let payload = {
+      id: requestId
+    };
+
+    return httpConnectionManager.send(apiPath, payload, true)
+      .catch(this.logErrorAndThrow);
+  }
+
+  loadMessage(messageId) {
```
kranar: This should construct a Message object and return that object.

PR #165, opened 2018-05-30
Item padding

spireken:


Applications/spire/include/spire/ui/item_padding_delegate.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+namespace spire {
+
+  //! Implements Qt' item delegate to support custom item padding.
+  class item_padding_delegate : public QStyledItemDelegate {
+    public:
+
+      //! Constructs an item delegate for custom padding.
+      /*!
+        \param padding The amount of item padding, unscaled.
```
kranar: The documentation says it's unscaled, but you're using it as if it is scaled. Perhaps you meant: The amount of padding in pixels per item.

PR #167, opened 2018-06-01
Misc fixes

spireken:


Applications/spire/source/time_and_sales/time_and_sales_window.cpp
```diff
@@ -28,7 +28,8 @@ time_and_sales_window::time_and_sales_window(
     : QWidget(parent),
       m_input_model(&input_model),
       m_v_scrolling(false),
-      m_h_scrolling(false) {
+      m_h_scrolling(false),
+      m_header_font(QFont()) {
```
kranar: Line 32 is not necessary.

Applications/spire/source/time_and_sales/time_and_sales_window.cpp
```diff
@@ -244,6 +246,15 @@ void time_and_sales_window::set_properties(
   QFontMetrics metrics(m_properties.m_font);
   auto row_height = metrics.height() + scale_height(2);
   m_table->verticalHeader()->setDefaultSectionSize(row_height);
+  if(m_properties.m_font.pointSize() >= 11) {
```
kranar: These hard coded values seem rather suspicious..

Applications/spire/source/time_and_sales/time_and_sales_window.cpp
```diff
@@ -244,6 +246,15 @@ void time_and_sales_window::set_properties(
...
+    m_header_font.setPointSizeF(m_properties.m_font.pointSize() * 0.8);
+  } else {
+    m_header_font.setPointSize(9);
+  }
+  m_table->horizontalHeader()->setFont(m_header_font);
+  QFontMetrics header_metrics(m_header_font);
+  m_table->horizontalHeader()->setFixedHeight(header_metrics.height() * 1.8);
+  repaint();
```
kranar: What is the reason for using repaint instead of update?
What is the reason for forcing an update right away?

Applications/spire/include/spire/time_and_sales/time_and_sales_window.hpp
```diff
@@ -82,6 +82,7 @@ namespace spire {
       bool m_h_scrolling;
       QTimer* m_v_scroll_bar_timer;
       QTimer* m_h_scroll_bar_timer;
+      QFont m_header_font;
```
kranar: I don't think having this font is helpful, it appears to be duplicating the font in the horizontal header. Instead of duplicating its value, refer to the existing font via horizontalHeader()->font();

PR #173, opened 2018-06-04
Updated window resize to use a constant

spireken:
Build is in time_and_sales dropbox folder.

### DIFF
--- a/Applications/spire/source/ui/window.cpp
+++ b/Applications/spire/source/ui/window.cpp
@@ -9,6 +9,10 @@
 
 using namespace spire;
 
+namespace {
+  auto PADDING_SIZE = 10;
+}
+
 window::window(QWidget* body, QWidget* parent)
     : QWidget(parent),
       m_body(body),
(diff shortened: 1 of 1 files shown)


Applications/spire/source/ui/window.cpp
```diff
@@ -9,6 +9,10 @@
 
 using namespace spire;
 
+namespace {
+  auto PADDING_SIZE = 10;
```
kranar: Make this const:

const auto ...

Applications/spire/source/ui/window.cpp
```diff
@@ -141,29 +145,28 @@ void window::set_border_stylesheet(const QColor& color) {
...
   auto top_left = mapToGlobal(geometry().topLeft());
   if(!m_resize_boxes.is_initialized()) {
     m_resize_boxes.emplace();
   }
-  m_resize_boxes->m_top_left = QRect(top_left.x() - padding_size.width(),
-    top_left.y() - padding_size.height(), padding_size.width(),
-    padding_size.height());
+  m_resize_boxes->m_top_left = QRect(top_left.x() - PADDING_SIZE,
+    top_left.y() - PADDING_SIZE, PADDING_SIZE,
+    PADDING_SIZE);
```
kranar: Fix 154 onto 153 if possible. Same goes for all the other lines. Try to fit in as much into the 80 columns.

PR #177, opened 2018-06-07
Fb1032

spireken:


Applications/spire/include/spire/ui/title_bar.hpp
```diff
@@ -52,8 +52,10 @@ namespace spire {
       */
       void set_icon(const QImage& icon, const QImage& unfocused_icon);
 
+#ifdef Q_OS_WIN
       bool nativeEventFilter(const QByteArray& event_type, void* message,
```
kranar: This can't be surrounded by a define guard because nativeEventFilter is a pure virtual function. You need to implement this method on all platforms. Instead, you must protect the definition with the define guard.

#ifdef Q_OS_WIN
  bool nativeEventFilter(...) {
    ...
  }
#else
  bool nativeEventFilter(...) {
    return false;
  }
#endif

Applications/spire/include/spire/ui/window.hpp
```diff
@@ -32,6 +32,11 @@ namespace spire {
       */
       void set_icon(const QImage& icon, const QImage& unfocused_icon);
 
+#ifdef Q_OS_WIN
```
kranar: Same as above, this can not be guarded.

PR #184, opened 2018-06-13
Added set_svg_icon method and updated classes to use it

spireken:


Applications/spire/source/ui/window.cpp
```diff
@@ -14,6 +14,11 @@ using namespace spire;
 
 namespace {
   const auto PADDING_SIZE = 10;
+
+  QImage get_svg_window_icon(const QString& icon_path) {
```
kranar: Rename from get to make.

PR #196, opened 2018-06-21
Fb1089

nicola-spire:


Applications/web_portal/web_app/library/source/pages/dashboard_page/dashboard_page.tsx
```diff
@@ -1,16 +1,88 @@
...
-    super(properties);
+  constructor(props: Properties) {
+    super(props);
   }
 
   public render(): JSX.Element {
-    return <div>Dashboard</div>;
+    return (
+      <VBoxLayout width='100%' height='100%'>
+        <div className={css(DashboardPage.STYLE.header)}>
```
kranar: Don't use CSS for layout, specify the layout using a BoxLayout and Padding.

Applications/web_portal/web_app/library/source/pages/dashboard_page/dashboard_page.tsx
```diff
@@ -1,16 +1,88 @@
...
+          <HBoxLayout className={css(DashboardPage.STYLE.leftHeader)}
+              width='136px' height='60px'>
```
kranar: Remove the use of this leftHeader STYLE and instead layout the items strictly using Paddings and BoxLayout elements.

Applications/web_portal/web_app/library/source/pages/dashboard_page/dashboard_page.tsx
```diff
@@ -1,16 +1,88 @@
...
+              <Padding/>
+              <div className={css(DashboardPage.STYLE.logo)}>
+                <img src='resources/dashboard/logo.png'
+                  className={css(DashboardPage.STYLE.logo)}/>
+              </div>
+              <Padding/>
+            </VBoxLayout>
+          </HBoxLayout>
+          <VBoxLayout height='60px' width='45px'
+              className={css(DashboardPage.STYLE.rightHeader)}>
```
kranar: Remove the rightHeader CSS class and layout these items using BoxLayout and Padding elements.

Applications/web_portal/web_app/library/source/pages/dashboard_page/dashboard_page.tsx
```diff
@@ -1,16 +1,88 @@
...
+            <div className={css(DashboardPage.STYLE.notificationButtonWrapper)}>
+              <NotificationButton width='15px' height='20px'/>
+            </div>
+            <Padding/>
+          </VBoxLayout>
+        </div>
+        <div className={css(DashboardPage.STYLE.separator)}/>
+      </VBoxLayout>);
   }
+  private static STYLE = StyleSheet.create({
```
kranar: This should be removed entirely once elements are layed out using the BoxLayout elements.

Applications/web_portal/web_app/library/source/pages/dashboard_page/notification_button.tsx
```diff
@@ -1,24 +1,49 @@
+import {css, StyleSheet} from 'aphrodite';
 import * as React from 'react';
 
 /** The properties used to render a NotificationButton. */
 export interface Properties {
 
-  /** The number of notification items. */
-  items: number;
+  /** The width of the button. */
+  width: number | string;
```
kranar: Do not modify the public API of this element, which includes its Properties. Remove the width/height props and use the fixed values provided in the spec. Restore the items and isOpen prop.

Applications/web_portal/web_app/library/source/pages/dashboard_page/dashboard_page.tsx
```diff
@@ -1,16 +1,59 @@
...
+            <Padding size='20px'/>
+            <Padding/>
+          </VBoxLayout>
+        </HBoxLayout>
+        <div className={css(DashboardPage.STYLE.separator)}/>
+      </VBoxLayout>);
   }
+  private static STYLE = StyleSheet.create({
+    header: {
+      backgroundColor: 'rgba(255, 255, 255, 0)'
```
kranar: The background color here is transparent (since the alpha channel is 0), what happens if you remove this header class entirely?

Applications/web_portal/web_app/library/source/pages/dashboard_page/dashboard_page.tsx
```diff
@@ -1,16 +1,59 @@
...
+    },
+    separator: {
+      width: '100%',
+      height: '1px',
+      backgroundColor: 'rgba(0, 0, 0, 0.16)'
```
kranar: Use the color code provided in the spec, instead of rgba use the value '#E6E6E6'

Applications/web_portal/web_app/library/source/pages/dashboard_page/notification_button.tsx
```diff
@@ -10,15 +11,29 @@ export interface Properties {
...
   public render(): JSX.Element {
-    return null;
+    const STYLE = StyleSheet.create({
+      button: {
+        width: '15px',
+        height: '20px',
+        border: 'none',
+        outline: 0,
+        padding: 0,
+        backgroundColor: 'rgba(0, 0, 0, 0)',
```
kranar: What happens if you remove the background color?

Applications/web_portal/web_app/library/source/pages/dashboard_page/dashboard_page.tsx
nicolahadwen: It will inherit from the parent so nothing in our case.

Applications/web_portal/web_app/library/source/pages/dashboard_page/dashboard_page.tsx
nicolahadwen: If i replace it with the border color in the spec, it will be extremely dark (darker than how it appears in the layout.xd), as we are not using boxShadow. If you look at the shadow field in the spec, it is rgba(0, 0, 0, 0.16)

Applications/web_portal/web_app/library/source/pages/dashboard_page/notification_button.tsx
nicolahadwen: It will inherit from the parent so nothing in our case.

Applications/web_portal/web_app/library/source/pages/dashboard_page/dashboard_page.tsx
kranar: Okay, if it does nothing then it's safe to remove this entirely right?

Applications/web_portal/web_app/library/source/pages/dashboard_page/dashboard_page.tsx
kranar: That makes sense, thanks.

Applications/web_portal/web_app/library/source/pages/dashboard_page/notification_button.tsx
kranar: Similarly then it should be safe to get rid of this background color right?

Applications/web_portal/web_app/library/source/pages/dashboard_page/dashboard_page.tsx
nicolahadwen: Yup!

Applications/web_portal/web_app/library/source/pages/dashboard_page/notification_button.tsx
nicolahadwen: Yup!

nicolahadwen: sorry, correction. Buttons have a default background color I need to override.

kranar:
Good good, only change I'd like to see is the use of transparent as the color name rather than using rgba(0, 0, 0, 0). Makes things more explicit and unambiguous.

kranar:
This is how it's done!

PR #200, opened 2018-06-28
Fb1106

spireken:
The build is in the time_and_sales Dropbox folder.

The hover event doesn't provide mouse button states, so I had to use static functions.


spireken:
I put an updated build called: time_and_sales_ui_tester_fb1106, just in case I overwrote the old one.

kranar:
Alternatively, you can use the sliderDown() property instead of listening to the signals.

spireken:
What about the second issue, where the scroll bar shouldn't show if the left mouse button is pressed when the mouse enters the scroll bar's area? Should I keep that the way it is?

PR #205, opened 2018-07-04
Fb1107

spireken:
I put the build in the misc_builds folder on Dropbox. You can test the Time and Sales scroll bar size and then the Security Input scroll bar size by activating the security input window.


spireken:
I renamed the build time_and_sales_ui_tester_fb1107.

PR #215, opened 2018-07-16
Fb1109

spireken:
I put the build in the time_and_sales_properties_window Dropbox folder.

### DIFF
--- a/Applications/spire/include/spire/time_and_sales/time_and_sales_properties_dialog.hpp
+++ b/Applications/spire/include/spire/time_and_sales/time_and_sales_properties_dialog.hpp
@@ -75,7 +75,6 @@ namespace spire {
       void set_properties(const time_and_sales_properties& properties);
       void set_text_color();
       void update_colors(int band_index);
-      void update_font_preview_stylesheet();
   };
 }
 
--- a/Applications/spire/include/spire/time_and_sales/time_and_sales_properties_dialog.hpp
+++ b/Applications/spire/include/spire/time_and_sales/time_and_sales_properties_dialog.hpp
@@ -73,7 +73,6 @@ namespace spire {
       void set_properties(const time_and_sales_properties& properties);
       void set_text_color();
       void update_colors(int band_index);
-      void update_band_list_font(const QFont& font);
   };
 }
 
--- a/Applications/spire/include/spire/time_and_sales/time_and_sales_properties_dialog.hpp
+++ b/Applications/spire/include/spire/time_and_sales/time_and_sales_properties_dialog.hpp
@@ -71,7 +71,7 @@ namespace spire {
       void set_font();
       void set_properties(const time_and_sales_properties& properties);
       void set_text_color();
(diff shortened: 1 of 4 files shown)


Applications/spire/include/spire/time_and_sales/time_and_sales_properties_dialog.hpp
```diff
@@ -64,7 +63,6 @@ namespace spire {
       QString m_band_list_stylesheet;
```
kranar: Remove m_band_list_stylesheet. In general don't store the stylesheet as a member, especially if it never changes. Instead compute it when you actually need it.

Applications/spire/source/time_and_sales/time_and_sales_properties_dialog.cpp
```diff
@@ -397,9 +382,9 @@ void time_and_sales_properties_dialog::update_colors(int band_index) {
...
 }
 
-void time_and_sales_properties_dialog::update_font_preview_stylesheet() {
-  m_font_preview_label->setStyleSheet(QString(R"(
-    background-color: white;
-    border: %1px solid #C8C8C8 %2px solid #C8C8C8;
-    )").arg(scale_height(1)).arg(scale_width(1)));
+void time_and_sales_properties_dialog::update_band_list_font(
+    const QFont& font) {
+  for(auto i = 0; i < m_band_list->count(); ++i) {
```
kranar: Can you look into whether this can be set on the widget itself rather than each individual item? The item's should not each be setting their own font but rather be using the parent window's default font.

Applications/spire/include/spire/time_and_sales/time_and_sales_properties_dialog.hpp
spireken: I combine it with the selected style to add the selected item border.

Applications/spire/source/time_and_sales/time_and_sales_properties_dialog.cpp
spireken: The only way to set it on the list widget itself is to use a stylesheet.

spireken: Actually, might be wrong on that.

Applications/spire/include/spire/time_and_sales/time_and_sales_properties_dialog.hpp
kranar: A class field should be unique to every instance of that class, meaning that something about that field depends on the object to which it's a member of. Reviewing the code I don't see anything that m_band_list_stylesheet depends on, it's a constant that is the same from every single instance. As a general principle if this is the case then the options are as follows (in order of preference):

P1. Compute the value if and only when it's needed.

If P1 is too expensive because the value is used frequently, then:

P2. Store the value in a static const variable whose scope is as narrow as possible.

PR #219, opened 2018-07-19
Fb1126

spireken:
I put the build in the book_view_properties_window Dropbox folder.

Regarding the event filter in book_view_properties_dialog, I needed a way to differentiate between how a tab was switched, whether through a click or key press, but Qt only provides currentChanged() signal for detecting tab changes. It's not a great solution, but I couldn't come up with anything better.


Applications/spire/source/book_view/book_view_highlight_properties_widget.cpp
```diff
@@ -0,0 +1,320 @@
...
+    m_order_highlight_color_button->get_style().m_background_color);
+  if(color.isValid()) {
+    update_color_button_stylesheet(m_order_highlight_color_button, color);
+  }
+}
+
+void book_view_highlight_properties_widget::
+    on_highlight_none_check_box_checked(int state) {
+  if(state == Qt::Checked) {
+    auto current_item = m_markets_list_widget->currentItem();
```
kranar: Do the static cast here upon initialization rather than on the line below.

Applications/spire/source/book_view/market_list_item.cpp
```diff
@@ -0,0 +1,49 @@
...
+  }
+  m_market_highlight->m_highlight_all_levels = false;
+}
+
+void market_list_item::remove_highlight() {
+  m_market_highlight = none;
+}
+
+void market_list_item::initialize() {
+  m_market_highlight = book_view_properties::market_highlight();
```
kranar: Use emplace:

  m_market_highlight.emplace();

spireken:
Searching 'boost optional emplace' doesn't bring up the documentation for it, only the standard library implementation. Based on the standard library's implementation I'm guessing the parameters are constructor arguments, so my version was calling the copy constructor.

kranar:
Exactly!  Perfect.

PR #231, opened 2018-07-24
Fb1161

nicola-spire:


Applications/web_portal/web_app/library/source/pages/dashboard_page/dashboard_page.tsx
```diff
@@ -11,8 +11,20 @@ interface Properties {
   /** The model to display. */
   model: DashboardModel;
 
-  /** The action to perform when logging out. */
-  onLogout?: () => void;
+  /** The action to perform when navigating to the profile. */
```
kranar: The dashboard page internally handles all of these actions, so no prop should be exposed for these. The only action the dashboard page can not possibly handle is the log out, which is why that has its own prop.

Applications/web_portal/web_app/library/source/pages/dashboard_page/dashboard_page.tsx
```diff
@@ -28,7 +40,15 @@ export class DashboardPage extends React.Component<Properties, State> {
       isSideMenuOpen: false,
       isLoading: false
     };
-    this.toggleSideMenuIsOpen = this.toggleSideMenuIsOpen.bind(this);
+    this.onToggleMenuIsOpen = this.onToggleMenuIsOpen.bind(this);
```
kranar: This is a better name, thanks.

Applications/web_portal/web_app/library/source/pages/dashboard_page/dashboard_page.tsx
```diff
@@ -28,7 +40,15 @@ export class DashboardPage extends React.Component<Properties, State> {
...
+  }
+
+  public static defaultProps = {
```
kranar: Put public static members first (above the constructor).

nicola-spire:
Done!

PR #232, opened 2018-07-27
Forced resize after call to showNormal()

spireken:
I put the build in the misc_builds folder on Dropbox.


spireken:
I'm still working on fixing the issue. I'm trying to get the resize to work in the window class, but it's not resizing the drop shadow properly.

kranar:
In that case this pr will get closed until a fix is available.

PR #234, opened 2018-07-27
Fb1157

spireken:
I put the build in the misc_builds Dropbox folder.

Same idea as before, except here for whatever reason it looks like the window ignores the resize if the new size is the same as the current size, so I force a resize then change it back to the right size.


spireken:
Did you have any comments for this case?

PR #235, opened 2018-07-31
Removed implicit conversion

spireken:
I replaced the build, just in case.


spireken:
I updated the build in the book_view_window folder.

PR #236, opened 2018-08-08
Fb1169

spireken:
The build is in the book_view_window Dropbox folder.


Applications/spire/include/spire/book_view/technicals_panel.hpp
```diff
@@ -14,6 +14,40 @@ namespace Spire {
       //! Constructs a TechnicalsPanel.
       TechnicalsPanel(QWidget* parent = nullptr);
 
+      // Sets the text of the close label.
+      /*
+        \param close The new string to display.
+      */
+      void set_close(const QString& close);
```
kranar: Instead of having all these setters, the TechnicalsPanel should take a model and update itself based on the model.

spireken:
I updated the build on Dropbox.

Applications/spire/include/spire/book_view/technicals_panel.hpp
```diff
@@ -14,6 +16,15 @@ namespace Spire {
...
       TechnicalsPanel(QWidget* parent = nullptr);
 
+      //! Sets the model the panel gets its data from.
+      /*
+        \param model The associated model.
+      */
+      void set_model(std::shared_ptr<BookViewModel> model);
+
+      // Resets the text of all labels to their default values.
+      void reset_labels();
```
kranar: Rename this to "reset_model" and implement it in such a way that it sets its own model to null and resets all labels.

Applications/spire/include/spire/book_view/book_view_window.hpp
```diff
@@ -63,20 +63,27 @@ namespace Spire {
       mutable ClosedSignal m_closed_signal;
       BookViewProperties m_properties;
       SecurityInputModel* m_input_model;
+      std::shared_ptr<BookViewModel> m_model;
       SecurityStack m_securities;
       Nexus::Security m_current_security;
       QWidget* m_body;
       QVBoxLayout* m_layout;
       TechnicalsPanel* m_header_widget;
```
kranar: Rename this from m_header_widget to m_technicals_panel. You can learn to use Visual Studio's renaming feature to do this safely.

Applications/spire/source/book_view/book_view_window.cpp
```diff
@@ -44,6 +48,14 @@ BookViewWindow::BookViewWindow(const BookViewProperties& properties,
...
 
 void BookViewWindow::set_model(std::shared_ptr<BookViewModel> model) {
+  m_transition_widget.reset();
+  m_header_widget->reset_labels();
+  QTimer::singleShot(2000, this, [=] { show_transition_widget(); });
+  m_is_data_loaded = false;
+  m_data_loaded_promise = model->load();
+  m_data_loaded_promise.then(
+    [=] (auto&& value) { on_data_loaded(std::move(value)); });
+  m_model = model;
```
kranar: When passing parameters by value to take ownership of them, you move the value rather than strictly assign by copy. That is you do this:

m_model = std::move(model);

Then you never reference model again, you strictly refer to m_model.

Furthermore, taking ownership of parameters should be done as soon as possible. That is, move line 58 as close to the function declaration as possible (which in this example means moving it all the way up to line 51).

Exceptions to this rule are when moves and copies perform the exact same action, like with ints, or bools... or other primitive types that do not have a move constructor.

Applications/spire/source/book_view/technicals_panel.cpp
```diff
@@ -10,28 +12,67 @@ TechnicalsPanel::TechnicalsPanel(QWidget* parent)
...
-  m_volume_label_widget = new LabeledDataWidget(tr("V"), "0", this);
+  m_volume_label_widget = new LabeledDataWidget(tr("V"), this);
   m_layout->addWidget(m_volume_label_widget, 1, 2);
 }
 
+void TechnicalsPanel::set_model(std::shared_ptr<BookViewModel> model) {
+  m_model = model;
+  if(m_model->get_close().is_initialized()) {
+    m_close_label_widget->set_data_text(
+      Beam::ToString(m_model->get_close().get()).c_str());
```
kranar: Instead of using .get(), dereference the optional:

*m_model->get_close()  vs. m_model->get_close().get()

Applications/spire/source/book_view/technicals_panel.cpp
```diff
@@ -91,3 +132,23 @@ void TechnicalsPanel::set_labeled_data_short_form_text() {
   m_close_label_widget->set_label_text(tr("C"));
   m_volume_label_widget->set_label_text(tr("V"));
 }
+
+void TechnicalsPanel::on_close_signal(const Money& close) {
+  m_close_label_widget->set_data_text(Beam::ToString(close).c_str());
```
kranar: Change all instances of converting from std::string to QString using s.c_str() to QString::fromStdString(s);

Do it anywhere you might have done this in the past by doing a Search/Find across all files. This avoids issues with performing UTF-8 conversions.

spireken: Do you want this on *any* call to c_str()? There's some calls to it in the models that don't use any Qt classes.

kranar: I can look over the other uses of it. Generally speaking c_str() is not advisable but I will briefly check it now and get back to you.

spireken:
I updated the build on Dropbox.

PR #244, opened 2018-08-14
Fb1182

spireken:
The build is in the book_view_window Dropbox folder.


spireken:
The crashing of the last build was a problem with one of the debug DLLs not being loaded so I just now put the release build on Dropbox. The program itself runs fine on my machine.

spireken:
I set the model loading delay to 1 milliseconds to make testing the BBO and technicals panel faster.

PR #246, opened 2018-08-16
Implemented bid and ask book quotes

spireken:
Do you want a build for this?

I ran it and outputted all the bid and ask values, and the bid goes from 100 to 99.01, and the ask goes from 100 to 100.99.

### DIFF
--- a/Applications/spire/source/book_view_ui_tester/random_book_view_model.cpp
+++ b/Applications/spire/source/book_view_ui_tester/random_book_view_model.cpp
@@ -93,15 +93,15 @@ QtPromise<void> RandomBookViewModel::load() {
         for(auto i = 0; i < 1000; ++i) {
           update();
         }
-        Quote bid_quote(Money(100), 100, Side::BID);
-        Quote ask_quote(Money(100), 100, Side::ASK);
+        Quote bid_quote(Money(), 100, Side::BID);
+        Quote ask_quote(Money(), 100, Side::ASK);
         for(auto i = 0; i < 100; ++i) {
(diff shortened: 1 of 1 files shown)


Applications/spire/source/book_view_ui_tester/random_book_view_model.cpp
```diff
@@ -93,6 +93,16 @@ QtPromise<void> RandomBookViewModel::load() {
         for(auto i = 0; i < 1000; ++i) {
           update();
         }
+        Quote bid_quote(Money(100), 100, Side::BID);
+        Quote ask_quote(Money(100), 100, Side::ASK);
+        for(auto i = 0; i < 100; ++i) {
+          bid_quote.m_price = (100 * Money::ONE) - (i * Money::CENT);
+          m_bids.push_back(BookQuote("TST", true, DefaultMarkets::TSX(),
+            bid_quote, ptime()));
```
kranar: For the timestamp, use the current universal time. Refer to boost documentation to see how to get that value.

Applications/spire/source/book_view_ui_tester/random_book_view_model.cpp
```diff
@@ -93,6 +93,16 @@ QtPromise<void> RandomBookViewModel::load() {
         for(auto i = 0; i < 1000; ++i) {
           update();
         }
+        Quote bid_quote(Money(100), 100, Side::BID);
```
kranar: It does not appear that Money(100) is significant or that this initial value is used in any way, as such I would recommend switching from Money(100) to just Money() as the former implies a significance, whereas the latter does not.

PR #251, opened 2018-11-02
Fb1278

spiresheep:


Applications/WebPortal/WebApp/library/source/pages/account_page/entitlements_page/check_mark_button.tsx
```diff
@@ -9,27 +9,26 @@ interface Properties {
   /** Determines if the check mark is readonly or not. */
   isReadOnly: boolean;
 
-  /** The onClick event handler. */
+  /** The onClick event handler.
+   * This is not needed if the component is readonly.
```
kranar: Don't bother describing the "This is not needed" part since that will be a property of all callbacks. All callbacks effectively have the property that if they're ignored or not used, nothing happens.

Applications/WebPortal/WebApp/library/source/pages/account_page/entitlements_page/check_mark_button.tsx
```diff
@@ -9,27 +9,26 @@ interface Properties {
   /** Determines if the check mark is readonly or not. */
   isReadOnly: boolean;
```
kranar: Is this prop needed anymore? Does the button look or behave differently if it's read only?

spiresheep: It does. If's it's readonly the button should never have it's color change on click.

kranar: This prop isn't used anywhere though. Try removing the prop and the single reference to it (which is in an ```if``` statement). Basically the idea is that any button you ever design can be read-only if the user of that button ignores the onclick. The only exceptions to this will be certain buttons that have a different style or effect depending on whether it's read only, but for this Component there is no style or effect that depends on that prop.

Applications/WebPortal/WebApp/library/source/pages/account_page/entitlements_page/check_mark_button.tsx
```diff
@@ -41,16 +34,10 @@ export class CheckMarkButton extends React.Component<Properties, {}> {
         alt='checkbox'
         height={this.props.size}
         width={this.props.size}
-        onClick={this.onClick}
+        onClick={this.props.onClick}
```
kranar: This is good but it has a problem, if the user doesn't supply an ```onClick``` prop then this component will fail. To fix this review how to define ```defaultProps``` in TypeScript and then define a default prop for this that is a function that does nothing, basically ```() => {}```

Applications/WebPortal/WebApp/library/source/pages/account_page/entitlements_page/check_mark_button.tsx
```diff
@@ -7,19 +7,23 @@ interface Properties {
...
-  /** Determines if the checkmark is currenly active(green) or inactive(grey)*/
+  /** Determines if the checkmark is currenly active(green) or inactive(grey) */
   isChecked: boolean;
 }
 
 /** Displays a check mark button. */
 export class CheckMarkButton extends React.Component<Properties, {}> {
+  static defaultProps = {
+    onClick: () => {}
+  }
```
kranar: Make this ```readonly``` and put a blank new line after the definition.

Applications/WebPortal/WebApp/library/source/pages/account_page/entitlements_page/check_mark_button.tsx
```diff
@@ -7,19 +7,23 @@ interface Properties {
   size: number | string;
 
   /** The onClick event handler. */
-  onClick?(event?: React.MouseEvent<any>): void;
+  onClick(event?: React.MouseEvent<any>): void;
```
kranar: This should still be optional.

Applications/WebPortal/WebApp/library/source/pages/account_page/entitlements_page/check_mark_button.tsx
```diff
@@ -34,10 +38,12 @@ export class CheckMarkButton extends React.Component<Properties, {}> {
         alt='checkbox'
         height={this.props.size}
         width={this.props.size}
-        onClick={this.props.onClick}
+        onClick={onClick}
```
kranar: Keep this as ```this.props.onClick```

Applications/WebPortal/WebApp/library/source/pages/account_page/entitlements_page/check_mark_button.tsx
```diff
@@ -7,19 +7,23 @@ interface Properties {
...
 export class CheckMarkButton extends React.Component<Properties, {}> {
+  static defaultProps = {
+    onClick: () => {}
+  }
   constructor(properties: Properties) {
     super(properties);
   }
 
   public render(): JSX.Element {
+    const {onClick} = this.props;
```
kranar: You can remove this.

Applications/WebPortal/WebApp/library/source/pages/account_page/entitlements_page/check_mark_button.tsx
```diff
@@ -7,24 +7,24 @@ interface Properties {
...
+  onClick?(event?: React.MouseEvent<any>): void;
 
   /** Determines if the checkmark is currenly active(green) or inactive(grey) */
   isChecked: boolean;
 }
 
 /** Displays a check mark button. */
 export class CheckMarkButton extends React.Component<Properties, {}> {
-  static defaultProps = {
+
```
kranar: Remove this blank line, the fact that this declaration is indented suffices to differentiate this line from the one above it. The request to put a blank line after the declaration is to clearly differentiate it from the declaration of the constructor. It seems picky but in the long run it helps make code more uniform/readable.

spiresheep: The spacing is fine then right? I just need to remove the blank line on line 18?

kranar:
Yes

PR #253, opened 2018-11-06
Added check to ensure data is being loaded before showing transition …

spireken:
I updated the build on Dropbox.

Now when the 'All Data Loaded' checkbox is checked before selecting a new security the transition widget doesn't show.


spireken:
I put two builds on Dropbox, one demonstrating the fix in this branch and another showing the build from spire2018 with the issue.

Just each build and select 'All Data Loaded' before selecting a security. The spire2018 build will show the transition widget even with the check box selected, where the fb1172 build won't show it if there's no data to load.

PR #256, opened 2018-11-07
Added commas to displayed text when displaying Quantities

spireken:
The builds for Time and Sales and Book View are already on Dropbox.


spireken:
I updated the builds on Dropbox. I made a temporary change to the Time and Sales model so that the initial time and sales have a size of 1,000.

Applications/spire/source/time_and_sales/time_and_sales_window_model.cpp
```diff
@@ -80,8 +80,8 @@ QVariant TimeAndSalesWindowModel::data(const QModelIndex& index,
         return QVariant::fromValue(
           m_entries[row_index].m_time_and_sale.GetValue().m_price);
       case Columns::SIZE_COLUMN:
-        return QVariant::fromValue(
-          m_entries[row_index].m_time_and_sale.GetValue().m_size);
+        return QLocale().toString(static_cast<int>(
```
kranar: Cast the size to a ```double```.

PR #257, opened 2018-11-08
Fb1287

spireken:
I put a new Book View tester up on Dropbox. I temporarily changed the increase in volume to 10M so you can see the change on the technicals panel.

I also tested this with fractional values and it's working correctly.

### DIFF
--- a/Applications/spire/include/spire/ui/ui.hpp
+++ b/Applications/spire/include/spire/ui/ui.hpp
@@ -3,6 +3,7 @@
 #include <QImage>
 #include <QRect>
 #include <QSize>
+#include "Nexus/Definitions/Quantity.hpp"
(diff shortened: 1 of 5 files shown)


spireken:
I tried using this but the 'f' flag by itself gives me trailing zeros when the number is displayed, and the 'g' flag gives me the same output as the default flag, 'e'.

I can use the 'f' flag with a precision of 0, but then this won't work for fractional shares.

kranar:
Okay good stuff, I will review this later.

spireken:
Should I fix this conflict on my end and re-submit this?

kranar:
No this case has been resolved and closed.

spireken:
So just delete my local branch?

kranar:
Yep, this case is now closed.

PR #270, opened 2018-11-28
Fb1312

spiresheep:


Applications/WebPortal/WebApp/tests/entitlements_page_tester/source/index.tsx
```diff
@@ -79,6 +81,10 @@ class TestApp extends React.Component<{}, State> {
               onClick={() => this.changeStatus('Server issue')}>
             UNSUCCESSFUL SUBMIT
           </button>
+          <button tabIndex={-1}
+              onClick={() => this.toggleButtonEnabled()}>
```
kranar: Bind the callback method and assign it directly to the prop.

PR #272, opened 2018-11-29
Fb1315

spiresheep:


Applications/WebPortal/WebApp/library/source/pages/loading_page.tsx
```diff
@@ -0,0 +1,68 @@
+import * as React from 'react';
+import { Transition } from 'react-transition-group';
+import { HBoxLayout, Padding, VBoxLayout } from '..';
+
+interface Properties {
+
+  /** Ammount of time in milliseconds before
```
kranar: Typo on ```amount```.

Applications/WebPortal/WebApp/library/source/pages/loading_page.tsx
```diff
@@ -0,0 +1,68 @@
...
+}
+
+interface State {
+  showAnimation: boolean;
+}
+
+/** Displays a loading page. */
+export class LoadingPage extends React.Component<Properties, State> {
+  public static readonly defaultProps = {
+    delay: 0
```
kranar: The delay in the spec is something like 1 or 2 seconds.

Applications/WebPortal/WebApp/library/source/pages/loading_page.tsx
```diff
@@ -0,0 +1,68 @@
...
+              </div>)}
+          </Transition>
+          <Padding/>
+        </HBoxLayout>
+        <Padding size={LoadingPage.BOTTOM_PADDING}/>
+      </VBoxLayout>);
+  }
+
+  public componentDidMount(): void {
+    this.setState({showAnimation: true});
```
kranar: Does this state do anything? What happens if you remove this state and pass a value of ```true``` to the ```Transition```'s ```in``` prop?

spiresheep: I originally tried it with the `in` prop being set to `true` and the transition would not occur.

kranar: Okay, keep it as is then.

PR #274, opened 2018-12-03
Fb1296

spireken:
I updated the Book View and Time and Sales windows on Dropbox.

### DIFF
--- a/Applications/spire/include/spire/book_view/book_view_window.hpp
+++ b/Applications/spire/include/spire/book_view/book_view_window.hpp
@@ -3,7 +3,6 @@
 #include <Beam/Pointers/Ref.hpp>
 #include <boost/optional.hpp>
 #include <QLabel>
-#include <QTimer>
 #include <QVBoxLayout>
 #include <QWidget>
 #include "Nexus/Definitions/Security.hpp"
@@ -73,7 +72,6 @@ namespace Spire {
       TechnicalsPanel* m_technicals_panel;
       std::unique_ptr<BboQuotePanel> m_bbo_quote_panel;
       std::unique_ptr<TransitionWidget> m_transition_widget;
-      QTimer m_transition_timer;
       QWidget* m_quote_widgets_container;
       QVBoxLayout* m_quote_widgets_container_layout;
       std::unique_ptr<BookViewTableWidget> m_table;
@@ -82,7 +80,6 @@ namespace Spire {
 
       void show_context_menu(const QPoint& pos);
       void show_properties_dialog();
-      void show_transition_widget();
       void on_data_loaded(Beam::Expect<void> value);
   };
 }
--- a/Applications/spire/include/spire/book_view/book_view_window.hpp
+++ b/Applications/spire/include/spire/book_view/book_view_window.hpp
@@ -66,7 +66,6 @@ namespace Spire {
       BookViewProperties m_properties;
       SecurityInputModel* m_input_model;
       std::shared_ptr<BookViewModel> m_model;
-      Nexus::Security m_current_security;
       SecurityWidget* m_security_widget;
       QWidget* m_body;
       QWidget* m_container_widget;
--- a/Applications/spire/include/spire/book_view/book_view_window.hpp
+++ b/Applications/spire/include/spire/book_view/book_view_window.hpp
@@ -59,7 +59,6 @@ namespace Spire {
     protected:
       void closeEvent(QCloseEvent* event) override;
       bool eventFilter(QObject* watched, QEvent* event) override;
-      void keyPressEvent(QKeyEvent* event) override;
 
     private:
       mutable ChangeSecuritySignal m_change_security_signal;
@@ -67,14 +66,13 @@ namespace Spire {
       BookViewProperties m_properties;
       SecurityInputModel* m_input_model;
       std::shared_ptr<BookViewModel> m_model;
-      SecurityStack m_securities;
       Nexus::Security m_current_security;
+      SecurityWidget* m_security_widget;
       QWidget* m_body;
+      QWidget* m_container_widget;
       QVBoxLayout* m_layout;
       TechnicalsPanel* m_technicals_panel;
       std::unique_ptr<BboQuotePanel> m_bbo_quote_panel;
-      std::unique_ptr<QWidget> m_overlay_widget;
-      std::unique_ptr<QLabel> m_empty_window_label;
       std::unique_ptr<TransitionWidget> m_transition_widget;
       QWidget* m_quote_widgets_container;
       QVBoxLayout* m_quote_widgets_container_layout;
@@ -84,11 +82,8 @@ namespace Spire {
 
       void set_current(const Nexus::Security& s);
       void show_context_menu(const QPoint& pos);
-      void show_overlay_widget();
       void show_properties_dialog();
       void show_transition_widget();
-      void on_security_input_accept(SecurityInputDialog* dialog);
-      void on_security_input_reject(SecurityInputDialog* dialog);
       void on_data_loaded(Beam::Expect<void> value);
   };
 }
--- a/Applications/spire/include/spire/book_view/book_view_window.hpp
+++ b/Applications/spire/include/spire/book_view/book_view_window.hpp
@@ -79,7 +79,6 @@ namespace Spire {
       std::unique_ptr<BookViewTableWidget> m_table;
       QtPromise<void> m_data_loaded_promise;
       bool m_is_data_loaded;
-      QTimer* m_model_load_timer;
 
       void set_current(const Nexus::Security& s);
       void show_context_menu(const QPoint& pos);
--- a/Applications/spire/include/spire/book_view/book_view_window.hpp
+++ b/Applications/spire/include/spire/book_view/book_view_window.hpp
@@ -79,6 +79,7 @@ namespace Spire {
       std::unique_ptr<BookViewTableWidget> m_table;
       QtPromise<void> m_data_loaded_promise;
       bool m_is_data_loaded;
+      QTimer* m_model_load_timer;
 
       void set_current(const Nexus::Security& s);
       void show_context_menu(const QPoint& pos);
--- a/Applications/spire/include/spire/book_view/book_view_window.hpp
+++ b/Applications/spire/include/spire/book_view/book_view_window.hpp
@@ -61,7 +61,6 @@ namespace Spire {
       bool eventFilter(QObject* watched, QEvent* event) override;
 
     private:
-      mutable ChangeSecuritySignal m_change_security_signal;
       mutable ClosedSignal m_closed_signal;
       BookViewProperties m_properties;
       SecurityInputModel* m_input_model;
@@ -79,7 +78,6 @@ namespace Spire {
       QtPromise<void> m_data_loaded_promise;
       bool m_is_data_loaded;
 
-      void set_current(const Nexus::Security& s);
       void show_context_menu(const QPoint& pos);
       void show_properties_dialog();
       void show_transition_widget();
--- a/Applications/spire/include/spire/book_view/book_view_window.hpp
+++ b/Applications/spire/include/spire/book_view/book_view_window.hpp
@@ -3,6 +3,7 @@
 #include <Beam/Pointers/Ref.hpp>
 #include <boost/optional.hpp>
 #include <QLabel>
+#include <QTimer>
 #include <QVBoxLayout>
 #include <QWidget>
 #include "Nexus/Definitions/Security.hpp"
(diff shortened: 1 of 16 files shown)


Applications/spire/include/spire/book_view/book_view_window.hpp
```diff
@@ -59,22 +59,20 @@ namespace Spire {
...
-      void keyPressEvent(QKeyEvent* event) override;
 
     private:
       mutable ChangeSecuritySignal m_change_security_signal;
       mutable ClosedSignal m_closed_signal;
       BookViewProperties m_properties;
       SecurityInputModel* m_input_model;
       std::shared_ptr<BookViewModel> m_model;
-      SecurityStack m_securities;
       Nexus::Security m_current_security;
```
kranar: Remove this field, if you need it put it in the ```SecurityWidget``` class.

Applications/spire/include/spire/security_input/security_input_dialog.hpp
```diff
@@ -46,11 +47,15 @@ namespace Spire {
       void mouseMoveEvent(QMouseEvent* event) override;
       void mousePressEvent(QMouseEvent* event) override;
       void mouseReleaseEvent(QMouseEvent* event) override;
+      void showEvent(QShowEvent* event) override;
 
     private:
       Nexus::Security m_security;
+      Beam::Ref<SecurityInputModel> m_model;
```
kranar: ```Ref``` is only used to pass arguments. Internally you store it as a pointer ```SecurityInputModel*```.

Applications/spire/include/spire/time_and_sales/time_and_sales_window.hpp
```diff
@@ -58,29 +57,22 @@ namespace Spire {
...
-      void keyPressEvent(QKeyEvent* event) override;
 
     private:
       mutable ChangeSecuritySignal m_change_security_signal;
       mutable ClosedSignal m_closed_signal;
       TimeAndSalesProperties m_properties;
-      SecurityInputModel* m_input_model;
       boost::optional<TimeAndSalesWindowModel> m_model;
-      SecurityStack m_securities;
       Nexus::Security m_current_security;
```
kranar: See above about removing this field.

Applications/spire/include/spire/ui/security_widget.hpp
```diff
@@ -34,17 +37,40 @@ namespace Spire {
...
+
       //! Connects a slot to the change security signal.
       boost::signals2::connection connect_security_change_signal(
         const ChangeSecuritySignal::slot_type& slot) const;
+
+    protected:
+      bool eventFilter(QObject* object, QEvent* event) override;
+
+    private:
+      mutable ChangeSecuritySignal m_security_change_signal;
```
kranar: The field should have the same name as its type, ```m_change_security_signal```.

Applications/spire/include/spire/ui/security_widget.hpp
```diff
@@ -34,17 +37,40 @@ namespace Spire {
...
       void set_widget(QWidget* widget);
 
+      //! Displays the overlay widget that covers the parent widget.
+      void show_overlay_widget();
+
+      //! Hides the overlay widget.
+      void hide_overlay_widget();
+
       //! Connects a slot to the change security signal.
       boost::signals2::connection connect_security_change_signal(
```
kranar: ```connect_change_security_signal```

Applications/spire/source/ui/security_widget.cpp
```diff
@@ -1,5 +1,124 @@
...
+  m_widget->installEventFilter(this);
+}
+
+connection SecurityWidget::connect_security_change_signal(
+    const ChangeSecuritySignal::slot_type& slot) const {
+  return m_security_change_signal.connect(slot);
+}
+
+bool SecurityWidget::eventFilter(QObject* object, QEvent* event) {
+  if(object == window() || object == m_widget) {
```
kranar: Don't capture events for the entire window.

Applications/spire/source/ui/security_widget.cpp
```diff
@@ -1,5 +1,124 @@
...
+  m_layout->setContentsMargins({});
+  m_empty_window_label = std::make_unique<QLabel>(
+    tr("Enter a ticker symbol."), this);
+  m_empty_window_label->setAlignment(Qt::AlignCenter);
+  m_empty_window_label->setStyleSheet(QString(R"(
+    font-family: Roboto;
+    font-size: %1px;
+    padding-top: %2px;)").arg(scale_height(12)).arg(scale_height(16)));
+  m_layout->addWidget(m_empty_window_label.get());
+  window()->installEventFilter(this);
```
kranar: Don't capture events for the entire window.

Applications/spire/source/book_view/book_view_window.cpp
```diff
@@ -99,9 +99,6 @@ bool BookViewWindow::eventFilter(QObject* watched, QEvent* event) {
 }
 
 void BookViewWindow::set_current(const Security& s) {
-  if(s == m_current_security) {
```
kranar: You still need this guard but you will have to compare against the ```m_security_widget```'s current security.

Applications/spire/source/ui/security_widget.cpp
```diff
@@ -38,57 +38,57 @@ void SecurityWidget::set_widget(QWidget* widget) {
...
-connection SecurityWidget::connect_security_change_signal(
+connection SecurityWidget::connect_change_security_signal(
     const ChangeSecuritySignal::slot_type& slot) const {
-  return m_security_change_signal.connect(slot);
+  return m_change_security_signal.connect(slot);
 }
 
 bool SecurityWidget::eventFilter(QObject* object, QEvent* event) {
-  if(object == window() || object == m_widget) {
+  if(object == m_widget) {
```
kranar: Use an ```&&``` to connect the two ```if``` statements together into one.

Applications/spire/source/ui/security_widget.cpp
```diff
@@ -38,57 +38,57 @@ void SecurityWidget::set_widget(QWidget* widget) {
...
-        if(m_current_security != Security()) {
-          auto s = m_securities.push_back(m_current_security);
-          if(s != Security()) {
-            m_current_security = s;
-            m_security_change_signal(s);
-          }
-        }
-        event->accept();
-        return true;
+      keyPressEvent(static_cast<QKeyEvent*>(event));
```
kranar: If the ```event``` was accepted then you need to return ```false``` to prevent the parent widget from processing it.

Applications/spire/source/book_view/book_view_window.cpp
spireken: I have the SecurityWidget doing the check to ensure that the same security isn't input twice; do I still need this even with that?

Applications/spire/source/ui/security_widget.cpp
spireken: I've made the change but eventFilter() returns true when the event has been handled/shouldn't be passed to the parent.

Applications/spire/source/book_view/book_view_window.cpp
kranar: Yes otherwise all of the side-effects kick in like resetting the ```m_bbo_quote_panel```, the ```m_table```, so on so forth...

Applications/spire/source/ui/security_widget.cpp
kranar: Agood... agood...

Applications/spire/source/book_view/book_view_window.cpp
kranar: Okay good.

Applications/spire/include/spire/book_view/book_view_window.hpp
```diff
@@ -59,22 +59,19 @@ namespace Spire {
     protected:
       void closeEvent(QCloseEvent* event) override;
       bool eventFilter(QObject* watched, QEvent* event) override;
-      void keyPressEvent(QKeyEvent* event) override;
 
     private:
       mutable ChangeSecuritySignal m_change_security_signal;
```
kranar: This signal should be part of the ```SecurityWidget``` too.

Applications/spire/source/ui/security_widget.cpp
```diff
@@ -44,10 +44,9 @@ connection SecurityWidget::connect_change_security_signal(
 }
 
 bool SecurityWidget::eventFilter(QObject* object, QEvent* event) {
-  if(object == m_widget) {
-    if(event->type() == QEvent::KeyPress) {
-      keyPressEvent(static_cast<QKeyEvent*>(event));
-    }
+  if(object == m_widget && event->type() == QEvent::KeyPress) {
+    keyPressEvent(static_cast<QKeyEvent*>(event));
```
kranar: You don't know if ```keyPressEvent``` accepted the event so you can't always return ```true```.

Applications/spire/include/spire/book_view/book_view_window.hpp
```diff
@@ -59,36 +60,29 @@ namespace Spire {
...
+      SecurityWidget* m_security_widget;
       QWidget* m_body;
+      QWidget* m_container_widget;
       QVBoxLayout* m_layout;
       TechnicalsPanel* m_technicals_panel;
       std::unique_ptr<BboQuotePanel> m_bbo_quote_panel;
-      std::unique_ptr<QWidget> m_overlay_widget;
-      std::unique_ptr<QLabel> m_empty_window_label;
       std::unique_ptr<TransitionWidget> m_transition_widget;
+      QTimer m_transition_timer;
```
kranar: Why is this timer part of the ```BookViewWindow``` instead of the ```TransitionWidget```? This will require every window that wants a loading screen to have to manage its own timer which defeats the purpose of factoring out all of the common functionality into the ```TransitionWidget``` class.

Find a way to implement this without every window managing its own ```QTimer```.

Applications/spire/include/spire/time_and_sales/time_and_sales_table_view.hpp
```diff
@@ -48,6 +48,7 @@ namespace Spire {
       QTimer m_v_scroll_bar_timer;
       std::unique_ptr<SnapshotLoadingWidget> m_loading_widget;
       std::unique_ptr<TransitionWidget> m_transition_widget;
+      QTimer m_transition_timer;
```
kranar: See above comment about ```QTimer```.

Applications/spire/include/spire/book_view/book_view_window.hpp
spireken: I can add the timer to TransitionWidget and have a callback for when the timer has been triggered, but I can't completely remove exposure of the windows to the timer because there's logic in the window that determines whether the TransitionWidget is shown or not.

kranar: Can you describe that in more detail? As far as I know a window should show a TransitionWidget as soon as a new model has been assigned to it and remove the TransitionWidget as soon as the model is finished loading. What other logic is there?

Applications/spire/source/security_input/security_input_dialog.cpp
```diff
@@ -117,6 +115,14 @@ void SecurityInputDialog::mouseReleaseEvent(QMouseEvent* event) {
   m_is_dragging = false;
 }
 
+void SecurityInputDialog::showEvent(QShowEvent* event) {
+  m_security_input_box = new SecurityInputBox(Ref(*m_model), m_initial_text, this);
+  m_security_input_box->connect_commit_signal(
+    [=] (const Security& s) { set_security(s); });
```
kranar: ```(const auto& security)```

Applications/spire/source/ui/security_widget.cpp
```diff
@@ -1,5 +1,123 @@
...
+  m_overlay_widget->move(0, 0);
+  m_overlay_widget->show();
+}
+
+void SecurityWidget::hide_overlay_widget() {
+  m_overlay_widget.reset();
+}
+
+void SecurityWidget::on_security_input_accept(SecurityInputDialog* dialog) {
+  auto s = dialog->get_security();
```
kranar: ```auto& security = dialog...```

Applications/spire/source/ui/security_widget.cpp
```diff
@@ -1,5 +1,123 @@
...
+      if(s != Security()) {
+        m_current_security = s;
+        m_change_security_signal(s);
+      }
+    }
+    event->accept();
+  }
+  auto pressed_key = event->text();
+  if(pressed_key[0].isLetterOrNumber()) {
+    auto dialog = new SecurityInputDialog(Ref(*m_input_model), pressed_key,
```
kranar: This introduces a memory leak.

Applications/spire/source/security_input/security_input_dialog.cpp
spireken: I thought lambdas was one of the places we used single character names?

Applications/spire/source/ui/security_widget.cpp
spireken: I have the [WA_DeleteOnClose](http://doc.qt.io/qt-5/qt.html#WidgetAttribute-enum) set on the dialog, is there something else I'm missing?

PR #279, opened 2018-12-18
Fb1321

spireken:
I updated the build on Dropbox with the updated line colors.

### DIFF
--- a/Applications/spire/include/spire/charting/chart_view.hpp
+++ b/Applications/spire/include/spire/charting/chart_view.hpp
@@ -25,6 +25,13 @@ namespace Spire {
         \param bottom_right The bottom right point to display.
       */
       void set_region(ChartPoint top_left, ChartPoint bottom_right);
+
+    protected:
+      void paintEvent(QPaintEvent* event) override;
+
+    private:
+      ChartPoint m_top_left;
+      ChartPoint m_bottom_right;
   };
 }
 
--- a/Applications/spire/include/spire/charting/chart_view.hpp
+++ b/Applications/spire/include/spire/charting/chart_view.hpp
@@ -30,8 +30,18 @@ namespace Spire {
       void paintEvent(QPaintEvent* event) override;
 
     private:
+      ChartValue::Type m_x_axis_type;
+      ChartValue::Type m_y_axis_type;
       ChartPoint m_top_left;
       ChartPoint m_bottom_right;
+      ChartValue m_x_axis_range;
+      ChartValue m_y_axis_range;
+      QFont m_label_font;
+
+      std::vector<ChartValue> get_axis_values(const ChartValue::Type& type,
+        const ChartValue& range_start, const ChartValue& range_end);
+      int value_to_pixel(const ChartValue& range_start,
+        const ChartValue& range_end, const ChartValue& value);
   };
 }
 
--- a/Applications/spire/include/spire/charting/chart_view.hpp
+++ b/Applications/spire/include/spire/charting/chart_view.hpp
@@ -41,7 +41,8 @@ namespace Spire {
       std::vector<ChartValue> get_axis_values(const ChartValue::Type& type,
         const ChartValue& range_start, const ChartValue& range_end);
       int value_to_pixel(const ChartValue& range_start,
-        const ChartValue& range_end, const ChartValue& value);
+        const ChartValue& range_end, const ChartValue& value,
+        int widget_size);
   };
 }
 
--- a/Applications/spire/include/spire/charting/chart_view.hpp
+++ b/Applications/spire/include/spire/charting/chart_view.hpp
@@ -39,8 +39,6 @@ namespace Spire {
       std::string m_timestamp_format;
       CustomVariantItemDelegate* m_item_delegate;
 
-      std::vector<ChartValue> get_axis_values(ChartValue::Type type,
-        ChartValue range_start, ChartValue range_end);
       QString drawable_timestamp(const boost::posix_time::ptime& time);
       ChartValue get_step(const ChartValue::Type& value_type,
         const ChartValue& range);
--- a/Applications/spire/include/spire/charting/chart_view.hpp
+++ b/Applications/spire/include/spire/charting/chart_view.hpp
@@ -44,6 +44,8 @@ namespace Spire {
         const ChartValue& range_end, const ChartValue& value,
         int widget_size);
       QString drawable_timestamp(const boost::posix_time::ptime& time);
+      ChartValue get_step(const ChartValue::Type& value_type,
+        const ChartValue& range);
   };
 }
 
--- a/Applications/spire/include/spire/charting/chart_view.hpp
+++ b/Applications/spire/include/spire/charting/chart_view.hpp
@@ -43,6 +43,7 @@ namespace Spire {
       int value_to_pixel(const ChartValue& range_start,
         const ChartValue& range_end, const ChartValue& value,
         int widget_size);
+      QString drawable_timestamp(const boost::posix_time::ptime& time);
   };
 }
 
--- a/Applications/spire/include/spire/charting/chart_view.hpp
+++ b/Applications/spire/include/spire/charting/chart_view.hpp
@@ -35,14 +35,12 @@ namespace Spire {
       ChartValue::Type m_y_axis_type;
       ChartPoint m_top_left;
       ChartPoint m_bottom_right;
-      ChartValue m_x_axis_range;
-      ChartValue m_y_axis_range;
       QFont m_label_font;
       std::string m_timestamp_format;
       CustomVariantItemDelegate* m_item_delegate;
 
-      std::vector<ChartValue> get_axis_values(const ChartValue::Type& type,
-        const ChartValue& range_start, const ChartValue& range_end);
+      std::vector<ChartValue> get_axis_values(ChartValue::Type type,
+        ChartValue range_start, ChartValue range_end);
       QString drawable_timestamp(const boost::posix_time::ptime& time);
       ChartValue get_step(const ChartValue::Type& value_type,
         const ChartValue& range);
--- a/Applications/spire/include/spire/charting/chart_view.hpp
+++ b/Applications/spire/include/spire/charting/chart_view.hpp
@@ -37,6 +37,7 @@ namespace Spire {
       ChartValue m_x_axis_range;
       ChartValue m_y_axis_range;
       QFont m_label_font;
+      std::string m_timestamp_format;
 
       std::vector<ChartValue> get_axis_values(const ChartValue::Type& type,
         const ChartValue& range_start, const ChartValue& range_end);
--- a/Applications/spire/include/spire/charting/chart_view.hpp
+++ b/Applications/spire/include/spire/charting/chart_view.hpp
@@ -3,6 +3,7 @@
 #include <QWidget>
 #include "spire/charting/charting.hpp"
 #include "spire/charting/chart_point.hpp"
+#include "spire/ui/custom_qt_variants.hpp"
 
 namespace Spire {
 
@@ -38,6 +39,7 @@ namespace Spire {
       ChartValue m_y_axis_range;
       QFont m_label_font;
       std::string m_timestamp_format;
+      CustomVariantItemDelegate* m_item_delegate;
 
       std::vector<ChartValue> get_axis_values(const ChartValue::Type& type,
         const ChartValue& range_start, const ChartValue& range_end);
--- a/Applications/spire/include/spire/charting/chart_view.hpp
+++ b/Applications/spire/include/spire/charting/chart_view.hpp
@@ -41,9 +41,6 @@ namespace Spire {
 
       std::vector<ChartValue> get_axis_values(const ChartValue::Type& type,
         const ChartValue& range_start, const ChartValue& range_end);
-      int value_to_pixel(const ChartValue& range_start,
-        const ChartValue& range_end, const ChartValue& value,
-        int widget_size);
       QString drawable_timestamp(const boost::posix_time::ptime& time);
       ChartValue get_step(const ChartValue::Type& value_type,
         const ChartValue& range);
--- a/Applications/spire/include/spire/charting/chart_view.hpp
+++ b/Applications/spire/include/spire/charting/chart_view.hpp
@@ -39,9 +39,8 @@ namespace Spire {
       std::string m_timestamp_format;
       CustomVariantItemDelegate* m_item_delegate;
 
-      QString drawable_timestamp(const boost::posix_time::ptime& time);
-      ChartValue get_step(const ChartValue::Type& value_type,
-        const ChartValue& range);
+      ChartValue calculate_step(ChartValue::Type value_type, ChartValue range);
+      QString get_string(ChartValue::Type type, ChartValue value) const;
   };
 }
 
--- a/Applications/spire/source/charting/chart_view.cpp
+++ b/Applications/spire/source/charting/chart_view.cpp
@@ -31,51 +31,47 @@ void ChartView::paintEvent(QPaintEvent* event) {
   painter.fillRect(event->rect(), "#25212E");
   painter.setPen(Qt::white);
   painter.setFont(m_label_font);
-  auto range_x = get_axis_values(m_x_axis_type,
-    m_top_left.m_x, m_bottom_right.m_x);
-  auto range_y = get_axis_values(m_y_axis_type,
+  auto y_values = get_axis_values(m_y_axis_type,
     m_bottom_right.m_y, m_top_left.m_y);
-  auto y_axis_fm = QFontMetrics(m_label_font);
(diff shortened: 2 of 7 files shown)


Applications/spire/source/charting/chart_view.cpp
```diff
@@ -0,0 +1,148 @@
...
+  painter.setPen(Qt::white);
+  painter.setFont(m_label_font);
+  auto y_values = get_axis_values(m_y_axis_type,
+    m_bottom_right.m_y, m_top_left.m_y);
+  auto font_metrics = QFontMetrics(m_label_font);
+  auto origin_x = width() -
+    (font_metrics.width("M") *
+      QString::number(
+        static_cast<double>(
+          static_cast<Quantity>(y_values.front())), 'f', 2).length()) +
```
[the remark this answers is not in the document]
spireken: What behavior do you want for this? Just draw no vertical/horizontal lines when the get_axis_values() returns an empty vector?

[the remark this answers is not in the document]
kranar: Yes

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -0,0 +1,148 @@
...
+    painter.setPen(Qt::white);
+    painter.drawLine(x, origin_y, x, origin_y + scale_height(2));
+    painter.drawText(x - timestamp_width / 2,
+      origin_y + font_metrics.height() + scale_height(2),
+      drawable_timestamp(static_cast<ptime>(x_values[i])));
+  }
+}
+
+std::vector<ChartValue> ChartView::get_axis_values(
+    const ChartValue::Type& type, const ChartValue& range_start,
```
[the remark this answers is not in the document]
spireken: get_axis_values() calles get_step() which uses the m_timestamp_format member.

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -0,0 +1,148 @@
...
+      scale_width(8);
+  auto origin_y = height() - scale_height(20);
+  painter.drawLine(origin_x, 0, origin_x, origin_y);
+  painter.drawLine(0, origin_y, origin_x, origin_y);
+  auto money_step = (height() - (height() - origin_y)) / (y_values.size() - 1);
+  if((y_values.size() - 1) * money_step < height()) {
+    money_step += 1;
+  }
+  for(auto i = 1; i < y_values.size() - 1; ++i) {
+    auto y = origin_y - (money_step * (i - 1)) - money_step;
```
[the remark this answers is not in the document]
spireken: I haven't been able to get this to work. In terms of the logic it should look like:

`auto y = origin_y - money_step - (money_step * (i - 1));`

I can keep the origin_y - money_step operation in a variable, but I think that's as much as I can simplify this equation.

Applications/spire/source/charting/chart_view.cpp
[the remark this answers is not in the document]
spireken: What should the default right margin be when the y_values vector is empty? Right now I use the rendered text size of a y_value to determine where the x origin should be, but I can't do this if there is no y values.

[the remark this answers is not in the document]
kranar: If either of them are empty then you don't render anything.

spireken:
I updated the build on Dropbox. I ran a test build for when the y_values are empty, and it just shows a completely blank widget with a black background.

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -12,6 +11,13 @@ using namespace boost::posix_time;
 using namespace Nexus;
 using namespace Spire;
 
+namespace {
+  QVariant chart_value_to_variant(ChartValue::Type type,
+      const ChartValue& value) {
+    return QVariant::fromValue(static_cast<Money>(value));
```
kranar: Pass by value, furthermore implement this function for all possible ```ChartValue::Type```s.

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -45,56 +52,57 @@ void ChartView::paintEvent(QPaintEvent* event) {
   painter.setFont(m_label_font);
   auto y_values = get_axis_values(m_y_axis_type,
     m_bottom_right.m_y, m_top_left.m_y);
+  auto x_values = get_axis_values(m_x_axis_type, m_top_left.m_x,
+    m_bottom_right.m_x);
+  if(y_values.size() == 0 || x_values.size() == 0) {
```
kranar: Use ```empty``` instead of ```size() == 0```.

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -45,56 +52,57 @@ void ChartView::paintEvent(QPaintEvent* event) {
...
     painter.setPen(Qt::white);
-    painter.drawLine(origin_x, y, origin_x + scale_width(2), y);
-    painter.drawText(origin_x + scale_width(3),
+    painter.drawLine(x_origin, y, x_origin + scale_width(2), y);
+    painter.drawText(x_origin + scale_width(3),
       y + (font_metrics.height() / 3), 
-        m_item_delegate->displayText(
-          QVariant::fromValue(static_cast<Money>(y_values[i])), QLocale()));
+        m_item_delegate->displayText(chart_value_to_variant(
+          ChartValue::Type::MONEY, y_values[i]), QLocale()));
```
kranar: No need to hardcode the type, it's part of the API.

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -12,6 +11,13 @@ using namespace boost::posix_time;
 using namespace Nexus;
 using namespace Spire;
 
+namespace {
+  QVariant chart_value_to_variant(ChartValue::Type type,
```
kranar: When naming a function, try to leverage the fact that functions can be overloaded, this means that the function's parameters are a part of its signature. You can rename the function to ```to_variant``` and leverage the fact that ```ChartValue``` is one of the function parameters. This naming convention has benefits when writing generic code.

Applications/spire/source/charting/chart_view.cpp
spireken: Why is the ChartValue passed by value here?

kranar: ChartValue's should always be passed by value, it's an 8 byte numeric type. Same reason you always pass an int or a double by value.

spireken:
I updated the build.

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -12,9 +12,21 @@ using namespace Nexus;
 using namespace Spire;
 
 namespace {
-  QVariant chart_value_to_variant(ChartValue::Type type,
-      const ChartValue& value) {
-    return QVariant::fromValue(static_cast<Money>(value));
+  QVariant to_variant(ChartValue::Type type,
+      ChartValue value) {
```
kranar: Fit this on line 15.

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -12,9 +12,21 @@ using namespace Nexus;
...
+    if(type == ChartValue::Type::DURATION) {
+      return QVariant::fromValue(static_cast<time_duration>(value));
+    }
```
kranar: ```else if```

spireken: Why is it else if here if it just returns? Optimization?

kranar: No, no optimization happens. The principle is this... If I see this written as a series of ```else if```, I know as a reader that this giant if statement only ever executes one single branch. If I see a list of independent ```if``` statements that are not connected by an ```else/if```, I need to look into the body of every single ```if``` to see that each branch performs a ```return```.

As a code reviewer or someone reading this in the future, the ```else if``` is a very quick indicator to me that only one single branch has to be considered.

spireken:
If I'm going to cache the vectors should I do the same for the origin values, steps, etc.?

kranar:
Don't cache the vectors, eliminate them entirely. Basically loop over what you would have had in the vectors and calculate the values on the fly.

Applications/spire/include/spire/charting/chart_view.hpp
```diff
@@ -25,6 +26,22 @@ namespace Spire {
...
+    private:
+      ChartValue::Type m_x_axis_type;
+      ChartValue::Type m_y_axis_type;
+      ChartPoint m_top_left;
+      ChartPoint m_bottom_right;
+      QFont m_label_font;
+      std::string m_timestamp_format;
+      CustomVariantItemDelegate* m_item_delegate;
+
+      QString drawable_timestamp(const boost::posix_time::ptime& time);
```
kranar: Pass by value, mark the method as ```const```.

Applications/spire/include/spire/charting/chart_view.hpp
```diff
@@ -25,6 +26,22 @@ namespace Spire {
...
+      ChartValue get_step(const ChartValue::Type& value_type,
```
kranar: Pass by value, mark the method as ```const```.

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -0,0 +1,158 @@
...
+    ChartValue::Type y_axis_type, QWidget* parent)
+    : QWidget(parent),
+      m_x_axis_type(x_axis_type),
+      m_y_axis_type(y_axis_type),
+      m_timestamp_format("%H:%M:%S") {
+  setFocusPolicy(Qt::NoFocus);
+  m_label_font = QFont("Roboto");
+  m_label_font.setPixelSize(scale_height(10));
+  m_item_delegate = new CustomVariantItemDelegate(this);
+  m_x_axis_type = ChartValue::Type::TIMESTAMP;
```
kranar: Remove line 39 and 40. These are initialized already on line 32 and 33.

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -0,0 +1,158 @@
...
+}
+
+ChartView::ChartView(ChartValue::Type x_axis_type,
+    ChartValue::Type y_axis_type, QWidget* parent)
+    : QWidget(parent),
+      m_x_axis_type(x_axis_type),
+      m_y_axis_type(y_axis_type),
+      m_timestamp_format("%H:%M:%S") {
+  setFocusPolicy(Qt::NoFocus);
+  m_label_font = QFont("Roboto");
```
kranar: Initialize values in the initializer list, ie. before line 34.

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -0,0 +1,158 @@
...
+  m_label_font.setPixelSize(scale_height(10));
+  m_item_delegate = new CustomVariantItemDelegate(this);
```
kranar: Initialize this in the initializer list as well.

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -0,0 +1,158 @@
...
+void ChartView::paintEvent(QPaintEvent* event) {
+  auto painter = QPainter(this);
+  painter.setFont(m_label_font);
+  painter.setPen(Qt::white);
+  auto font_metrics = QFontMetrics(m_label_font);
+  auto x_origin = width() -
+    (font_metrics.width("M") *
+      QString::number(
+        static_cast<double>(
+          static_cast<Quantity>(ChartValue(m_top_left.m_y))),
```
kranar: This needs to be revised, for one ```m_top_left.m_y``` is already a ```ChartValue```, so no need to explicitly invoke the ```ChartValue``` constructor on it. Second, there's no guarantee that it's a ```Quantity```, so the cast to ```Quantity``` could have invalid behavior.

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -0,0 +1,158 @@
...
+        m_item_delegate->displayText(to_variant(m_y_axis_type, y_value),
+          QLocale()));
+  }
+  auto x_step_count = x_range / x_step;
+  auto x_pixel_step = x_origin / (x_range / x_step);
+  if((x_step_count - 1) * x_pixel_step < width()) {
+    x_pixel_step += 1;
+  }
+  auto x_text_width = font_metrics.width(drawable_timestamp(
+    static_cast<ptime>(m_top_left.m_x)));
```
kranar: Don't assume that the x-axis is a timestamp. This task only needs to work for the case where it's a timestamp but the implementation should still be general because you'll be implementing the other cases very soon.

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -0,0 +1,158 @@
...
+  while(x_value >= m_top_left.m_x) {
+    x_value -= x_step;
+    x -= x_pixel_step;
+    painter.setPen("#3A3348");
+    painter.drawLine(x, 0, x, y_origin);
+    painter.setPen(Qt::white);
+    painter.drawLine(x, y_origin, x, y_origin + scale_height(2));
+    painter.drawText(x - x_text_width / 2,
+      y_origin + font_metrics.height() + scale_height(2),
+      drawable_timestamp(static_cast<ptime>(x_value)));
```
kranar: Don't assume ```x_value``` is a ```ptime```

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -0,0 +1,158 @@
...
+  }
+}
+
+QString ChartView::drawable_timestamp(const ptime& time) {
+  std::ostringstream ss;
```
kranar: Put a TODO in here to write a function to convert from ```ptime``` to ```QTime```.

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -0,0 +1,158 @@
...
+  ss.imbue(std::locale(std::cout.getloc(),
+    new boost::posix_time::time_facet(m_timestamp_format.c_str())));
+  ss << time;
+  return QString::fromStdString(ss.str());
+}
+
+ChartValue ChartView::get_step(const ChartValue::Type& value_type,
+    const ChartValue& range) {
+  auto step = ChartValue();
```
kranar: This variable is unused.

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+QString ChartView::drawable_timestamp(const ptime& time) {
+  std::ostringstream ss;
+  ss.imbue(std::locale(std::cout.getloc(),
+    new boost::posix_time::time_facet(m_timestamp_format.c_str())));
+  ss << time;
+  return QString::fromStdString(ss.str());
+}
+
+ChartValue ChartView::get_step(const ChartValue::Type& value_type,
```
kranar: Pass by value.

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -0,0 +1,158 @@
...
+    } else if(money_range <= 10 * Money::CENT) {
+      return ChartValue(Money::CENT);
+    } else if(money_range <= Money::ONE) {
+      return ChartValue(10 * Money::CENT);
+    } else {
+      return ChartValue(Money::ONE);
+    }
+  } else if(value_type == ChartValue::Type::TIMESTAMP) {
+    auto time_range = static_cast<time_duration>(range);
+    if(time_range <= time_duration(0, 1, 0)) {
```
kranar: Don't use the ```time_duration``` constructor, instead use ```seconds(1)``` or ```hours(1)``` or ```minutes(1)```. It's more descriptive.

Applications/spire/source/charting/chart_view.cpp
spireken: Does this work for [QFont](http://doc.qt.io/qt-5/qfont.html)? The only other constructor I see takes point size, not pixel size.

Applications/spire/source/charting/chart_view.cpp
spireken: How do I generalize this? Is there a way to do something like: static_cast<m_x_axis_type>(value)?

Applications/spire/include/spire/charting/chart_view.hpp
spireken: I can't mark this constant because it modifies m_timestamp_format. Or should I make the timestamp member mutable?

kranar: If that's the case then rename the function. A function named ```get_xxx``` must not modify any state.

Applications/spire/source/charting/chart_view.cpp
kranar: Yes exactly.

Applications/spire/source/charting/chart_view.cpp
kranar: No, write a function that converts a ```ChartValue``` to a ```QString``` representing the label based on the type of axis.

PR #290, opened 2019-01-21
Fb1347

spireken:


Applications/spire/include/spire/charting/chart_view.hpp
```diff
@@ -50,6 +50,24 @@ namespace Spire {
       //! Removes the crosshair from the chart.
       void reset_crosshair();
 
+      //! Passes mouse click events to this view.
+      /*
+        \param event The click event.
+      */
+      void mouse_click(QMouseEvent* event);
```
kranar: The public API can not change. The ChartView doesn't know anything about the mouse, otherwise it would handle its own mouse events.

spireken: So the charting window is managing all of this?

kranar: Yep exactly.

spireken: But then won't ChartView and ChartWindow have duplicate members for the region?

kranar:
Okay sure, add a getter to the API, add it before the setter.

Applications/spire/include/spire/charting/chart_view.hpp
```diff
@@ -50,6 +50,10 @@ namespace Spire {
       //! Removes the crosshair from the chart.
       void reset_crosshair();
 
+      //! Returns a pair (top left, bottom right) of ChartPoints
```
kranar: Use ```std::tuple```, ```std::pair``` is obsolete.

Applications/spire/source/charting/charting_window.cpp
```diff
@@ -21,9 +21,14 @@ using namespace boost::signals2;
 using namespace Beam;
 using namespace Spire;
 
+namespace {
+  const auto CHART_ZOOM_PERCENT = 1.1;
```
kranar: Are you only zooming in 1.1% at a time?

Applications/spire/include/spire/charting/chart_view.hpp
spireken: So how do tuples work with structured bindings? A pair guarantees that only two values are returned, but a tuple can be arbitrarily long. Is there a way to make its length explicit?

Applications/spire/source/charting/charting_window.cpp
spireken: Should I call it a factor?

Applications/spire/include/spire/charting/chart_view.hpp
kranar: ```std::tuple``` works in a similar way as ```std::pair```. To access its members you use ```std::get<I>``` where ```I``` is an integer representing the 0-based index of the element you want to access:

https://en.cppreference.com/w/cpp/utility/tuple

Applications/spire/source/charting/charting_window.cpp
```diff
@@ -22,7 +22,7 @@ using namespace Beam;
 using namespace Spire;
 
 namespace {
-  const auto CHART_ZOOM_PERCENT = 1.1;
+  const auto CHART_ZOOM_AMOUNT = 1.1;
```
kranar: Rename this to ```ZOOM_FACTOR```.

PR #295, opened 2019-01-24
Fb1364

spiresheep:


Applications/WebPortal/WebApp/library/source/pages/account_directory_page/account_directory_page.tsx
```diff
@@ -1,18 +1,230 @@
...
   displaySize: DisplaySize;
+
+  /** Called when the user wants to make a new group. */
+  onNewGroupClick?: () => {};
+
+  /** Called when the user wants to make a new account. */
+  onNewAccountClick?: () => {};
+
+  /** Model that contains imformation about the accounts. */
+  model: AccountDirectoryModel;
```
kranar: Put the callbacks at the end.

Applications/WebPortal/WebApp/library/source/pages/account_directory_page/account_directory_page.tsx
```diff
@@ -1,18 +1,230 @@
...
 }
 
-/** Displays an account's profile page. */
-export class AccountDirectoryPage extends React.Component<Properties> {
+interface State {
+  filter: string;
+}
+
+/** Displays an directory of accounts. */
```
kranar: Fix typo.

Applications/WebPortal/WebApp/library/source/pages/account_directory_page/account_directory_page.tsx
```diff
@@ -1,18 +1,230 @@
...
+    })();
+    const buttonStyle = (() => {
+      if(this.props.displaySize === DisplaySize.SMALL) {
+        return AccountDirectoryPage.DYNAMIC_STYLE.buttonSmall;
+      } else {
+        return AccountDirectoryPage.DYNAMIC_STYLE.button;
+      }
+    })();
+    const cards = (() => {
+      const cards = [];
```
kranar: No need for the IIFE here. You can effectively remove line 68, 78 and 79.

Applications/WebPortal/WebApp/library/source/pages/account_directory_page/account_directory_page.tsx
```diff
@@ -1,18 +1,230 @@
...
+      for (const group of this.props.model.groups) {
+        cards.push(
+          <GroupCard key={group.id}
+            displaySize={this.props.displaySize}
+            group={group}
+            accounts={[]}
```
kranar: You will have to populate the accounts prop. This is the most challenging aspect of this page. You need to keep track of which GroupCard's have been expanded by the user and then when a user expands a GroupCard you need to load the list of accounts for that group from the model.

Applications/WebPortal/WebApp/library/source/pages/account_directory_page/account_directory_page.tsx
```diff
@@ -1,18 +1,230 @@
...
+            filter={this.state.filter}/>);
+      }
+      return cards;
+    })();
     return (
-      <div>
-        BEEP
+      <div style={AccountDirectoryPage.STYLE.page}>
+        <div style={contentWidth}>
+          <div id={'header'} style={headerBoxStyle}>
```
kranar: You don't need to put strings in curly braces in JSX.

Applications/WebPortal/WebApp/library/source/pages/account_directory_page/filter_bar.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react';
+
+interface Properties {
+
+  /** The value of the filter. */
+  value: string;
+
+  /** Called when the value changes.
+   * @param newValue - The updated value.
```
kranar: Rename to value to be consistent with other JavaScript callbacks.

Applications/WebPortal/WebApp/library/source/pages/account_directory_page/filter_bar.tsx
```diff
@@ -0,0 +1,26 @@
...
+   */
+  onChange: (newValue: string) => void;
+}
+
+export class FilterBar extends React.Component<Properties> {
```
kranar: Document this class.

Applications/WebPortal/WebApp/library/source/pages/account_directory_page/group_card.tsx
```diff
@@ -0,0 +1,35 @@
...
+  group: Beam.DirectoryEntry;
+
+  /** The accounts in the group. */
+  accounts: AccountEntry[];
+
+  /** The current filter used on the accounts. */
+  filter: string;
+}
+
+export class GroupCard extends React.Component<Properties> {
```
kranar: Document.

Applications/WebPortal/WebApp/library/source/pages/account_directory_page/local_account_directory_model.ts
```diff
@@ -17,7 +17,9 @@ export class LocalAccountDirectoryModel extends AccountDirectoryModel {
     this._groups = groups.clone();
     this._accounts = new Beam.Map<Beam.DirectoryEntry, AccountEntry[]>();
     for(const group of this._groups) {
-      this._accounts.set(group, accounts.get(group).slice());
+      if(accounts.get(group)) {
```
kranar: I don't think this check should be performed.

Applications/WebPortal/WebApp/library/source/pages/account_directory_page/account_directory_page.tsx
```diff
@@ -65,22 +71,27 @@ export class AccountDirectoryPage extends React.Component<Properties, State> {
...
-      for (const group of this.props.model.groups) {
-        cards.push(
-          <GroupCard key={group.id}
-            displaySize={this.props.displaySize}
-            group={group}
-            accounts={[]}
-            filter={this.state.filter}/>);
+    const cards = [];
+    for (const group of this.props.model.groups) {
+      let accounts: AccountEntry[] = [];
```
kranar: ```const accounts = this.state.accounts.get(group) || [];```

Applications/WebPortal/WebApp/library/source/pages/account_directory_page/account_directory_page.tsx
```diff
@@ -105,6 +116,21 @@ export class AccountDirectoryPage extends React.Component<Properties, State> {
 
   private onChange(newFilter: string) {
     this.setState({filter: newFilter});
```
kranar: Update both states together instead of calling ```setState``` twice.

Applications/WebPortal/WebApp/library/source/pages/account_directory_page/account_directory_page.tsx
```diff
@@ -105,6 +116,21 @@ export class AccountDirectoryPage extends React.Component<Properties, State> {
...
+    this.setState({
+      openedGroups: new Beam.Set<Beam.DirectoryEntry>()
+    });
+  }
+
+  private async onCardClick(group: Beam.DirectoryEntry) {
+    if(!this.state.accounts.get(group)) {
```
kranar: Always perform the load no matter what rather than only loading if you never loaded before.

Applications/WebPortal/WebApp/library/source/pages/account_directory_page/local_account_directory_model.ts
spiresheep: It is possible that some groups might not have any accounts associated with them....

Applications/WebPortal/WebApp/library/source/pages/account_directory_page/account_directory_page.tsx
```diff
@@ -105,6 +116,21 @@ export class AccountDirectoryPage extends React.Component<Properties, State> {
...
+      const accounts = await this.props.model.loadAccounts(group);
+      this.state.accounts.set(group, accounts);
+    }
+    this.state.openedGroups.add(group);
```
kranar: What if I want to close the card instead of open it?

Applications/WebPortal/WebApp/library/source/pages/account_directory_page/group_card.tsx
```diff
@@ -16,8 +16,15 @@ interface Properties {
 
   /** The current filter used on the accounts. */
   filter: string;
+
+  /**  Determines if the card is opened. */
+  isOpen: boolean;
+
+  /** Called when the card is opened. */
+  onClick: (group: Beam.DirectoryEntry) => void;
```
kranar: This should be called anytime the card's drop down button is clicked, a user can click to open or close the card.

Applications/WebPortal/WebApp/library/source/pages/account_directory_page/local_account_directory_model.ts
kranar: Yes that's fine, but no error would get thrown in that case.

spiresheep: So every group should be in the map even if the array associated with it is empty?

kranar: Yep, that means the group is empty. The UI indicates how to handle an empty group.

Applications/WebPortal/WebApp/library/source/pages/account_directory_page/account_directory_page.tsx
```diff
@@ -115,8 +110,8 @@ export class AccountDirectoryPage extends React.Component<Properties, State> {
   }
 
   private onChange(newFilter: string) {
-    this.setState({filter: newFilter});
     this.setState({
+      filter: newFilter,
       openedGroups: new Beam.Set<Beam.DirectoryEntry>()
     });
   }
```
kranar: Still need to handle closing the card.

spiresheep: Won't the change in state cause the cards to close? It's only open if it's in the open set, but when the filter changes I replace the current set with the empty set.

kranar:
This is good but you have to handle the situation where the user clicks on the drop down button for a card to close it.

Applications/WebPortal/WebApp/library/source/pages/account_directory_page/account_directory_page.tsx
spiresheep: Looks like I didn't save the changes.

kranar: The comment about closing cards is not specific to any line of code, it's just a general feature that needs to get implemented as part of this task.

PR #299, opened 2019-01-25
ChartView cursor now hides when it leaves the region

spireken:


Applications/spire/source/charting/chart_view.cpp
```diff
@@ -142,7 +142,9 @@ void ChartView::paintEvent(QPaintEvent* event) {
       m_item_delegate->displayText(to_variant(m_model->get_x_axis_type(), x),
       QLocale()));
   }
-  if(m_crosshair_pos) {
+  if(m_crosshair_pos && m_crosshair_pos.value().x() <= m_x_origin &&
+      m_crosshair_pos.value().y() <= m_y_origin) {
+    setCursor(Qt::BlankCursor);
```
kranar: Don't modify state in the ```paintEvent```, the ```paintEvent``` is strictly for translating the view into pixels.

PR #303, opened 2019-02-05
Fb1366

spireken:


Applications/spire/source/charting/chart_view.cpp
```diff
@@ -113,6 +117,10 @@ void ChartView::set_region(const ChartPoint& top_left,
   m_y_range = m_top_left.m_y - m_bottom_right.m_y;
   m_y_axis_step = calculate_step(m_model->get_y_axis_type(), m_y_range);
   update_origins();
+  m_loaded_data = m_model->load(m_top_left.m_x, m_bottom_right.m_x);
+  m_loaded_data.then([&] (auto result) {
+    m_candlesticks = result.Get();
```
kranar: ```std::move``` the result to avoid a needless copy, the data could be very big.

Applications/spire/source/charting/charting_window.cpp
```diff
@@ -137,6 +137,11 @@ ChartingWindow::ChartingWindow(Ref<SecurityInputModel> input_model,
   m_chart->installEventFilter(this);
 }
 
+void ChartingWindow::set_model(std::shared_ptr<ChartModel> model) {
+  m_model = model;
+  m_chart->set_model(*m_model);
```
kranar: Create a new ```ChartView```.

Applications/spire/include/spire/charting/chart_view.hpp
```diff
@@ -84,6 +91,8 @@ namespace Spire {
       std::vector<ChartValue> m_x_axis_values;
       int m_x_axis_text_width;
       std::vector<ChartValue> m_y_axis_values;
+      QtPromise<std::vector<Spire::Candlestick>> m_loaded_data;
```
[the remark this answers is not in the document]
spireken: The issue I was having is that if it goes out of scope then() is never called.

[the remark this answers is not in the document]
kranar: Find a better name.

Applications/spire/include/spire/charting/chart_view.hpp
```diff
@@ -91,7 +85,7 @@ namespace Spire {
       std::vector<ChartValue> m_x_axis_values;
       int m_x_axis_text_width;
       std::vector<ChartValue> m_y_axis_values;
-      QtPromise<std::vector<Spire::Candlestick>> m_loaded_data;
+      QtPromise<std::vector<Spire::Candlestick>> m_loaded_candlestick_promise;
```
kranar: The problem with the name is that nothing has been loaded. A promise is a data structure that performs an action at some unspecified point in the future but you have named it using the past tense as if this variable already loaded the candlesticks. You can name this ```pending_candlesticks```, ```candlestick_promise```, ```candlestick_loader``` etc... but don't use the past tense.

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -117,9 +116,10 @@ void ChartView::set_region(const ChartPoint& top_left,
   m_y_range = m_top_left.m_y - m_bottom_right.m_y;
   m_y_axis_step = calculate_step(m_model->get_y_axis_type(), m_y_range);
   update_origins();
-  m_loaded_data = m_model->load(m_top_left.m_x, m_bottom_right.m_x);
-  m_loaded_data.then([&] (auto result) {
-    m_candlesticks = result.Get();
+  m_loaded_candlestick_promise = m_model->load(m_top_left.m_x,
+    m_bottom_right.m_x);
+  m_loaded_candlestick_promise.then([&] (auto result) {
+    m_candlesticks = std::move(result.Get());
```
kranar: Line 122 can get called at any arbitrary point in the future, imagine it gets called 5 years from now... is the only operation you want to perform a re-assignment?

kranar: When I ask an ambiguous question it's mostly because I want you to think it through carefully. It's a good opportunity for you to gain a better understanding as opposed to me just giving you the answer to every detail. There is an important concept that is worth gaining an intuition for.

PR #306, opened 2019-02-06
Fb1375

spireken:


Applications/spire/source/charting/chart_view.cpp
```diff
@@ -248,6 +264,19 @@ void ChartView::showEvent(QShowEvent* event) {
   QWidget::showEvent(event);
 }
 
+void ChartView::update_auto_scale() {
+  if(m_candlesticks.size() > 0) {
+    auto auto_scale_top = ChartValue(-INT_MAX);
+    auto auto_scale_bottom = ChartValue(INT_MAX);
+    for(auto candle : m_candlesticks) {
```
kranar: ```auto&```

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -248,6 +264,19 @@ void ChartView::showEvent(QShowEvent* event) {
   QWidget::showEvent(event);
 }
 
+void ChartView::update_auto_scale() {
+  if(m_candlesticks.size() > 0) {
+    auto auto_scale_top = ChartValue(-INT_MAX);
```
kranar: ```-INT_MAX``` is always a bug, period. At any rate candlesticks almost certainly do not represent ```int```s so using an extreme int value will result in bugs. Find another solution to this problem.

spireken:
Are you saying there's more issues on top of this?

kranar:
Yes, the comments I made are mostly style guide fixes but if you try out the demo you'll see it doesn't work according to spec and furthermore there are additional bugs that need to get fixed. Try working on this as if you were a brand new user who never used Spire before trying this out for the first time... you'll notice that the UI has various sloppy issues related to this autoscaling feature.

Applications/spire/include/spire/charting/chart_view.hpp
```diff
@@ -61,6 +61,9 @@ namespace Spire {
       void set_region(const ChartPoint& top_left,
         const ChartPoint& bottom_right);
 
+      //! Toggles the auto scale on or off.
+      void toggle_auto_scale();
```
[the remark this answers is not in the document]
spireken: So should the button know its' activated/deactivated status?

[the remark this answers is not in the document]
kranar: The window knows.

Applications/spire/include/spire/charting/chart_view.hpp
```diff
@@ -61,6 +61,12 @@ namespace Spire {
       void set_region(const ChartPoint& top_left,
         const ChartPoint& bottom_right);
 
+      //! Toggles the auto scale on or off.
```
kranar: Update the documentation.

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -121,10 +128,20 @@ void ChartView::set_region(const ChartPoint& top_left,
   m_candlestick_promise.then([&] (auto result) {
     m_candlesticks = std::move(result.Get());
     update();
```
kranar: Don't double update.

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -248,6 +265,26 @@ void ChartView::showEvent(QShowEvent* event) {
   QWidget::showEvent(event);
 }
 
+void ChartView::update_auto_scale() {
+  if(m_candlesticks.size() > 1) {
+    auto auto_scale_top = m_candlesticks[1].GetHigh();
```
kranar: What if there's only one candlestick?

spireken: I updated the if statement above it to check for that.

I noticed that in the old chart that auto-scale wouldn't do anything if there is only 1 candlestick completely visible.

kranar:
It's still buggy.

spireken:
The chart or that code?

kranar:
Both.

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -127,7 +127,6 @@ void ChartView::set_region(const ChartPoint& top_left,
     m_bottom_right.m_x);
   m_candlestick_promise.then([&] (auto result) {
     m_candlesticks = std::move(result.Get());
-    update();
     if(m_is_auto_scaled) {
```
kranar: This is still a bug, you must guarantee only one update is performed. The problem with the old implementation was that you could potentially be updating twice, now it's possible you never update.

spireken:
I noticed sometimes the ToggleButtons don't respond on the first click.

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -248,6 +266,20 @@ void ChartView::showEvent(QShowEvent* event) {
   QWidget::showEvent(event);
 }
 
+void ChartView::update_auto_scale() {
+  if(m_candlesticks.empty()) {
+    return;
+  }
+  auto auto_scale_top = m_candlesticks[0].GetHigh();
```
kranar: Prefer to use ```front``` and ```back``` instead of ```[0]``` ```[size() - 1]```.

Applications/spire/source/charting/charting_window.cpp
```diff
@@ -140,6 +142,10 @@ void ChartingWindow::set_model(std::shared_ptr<ChartModel> model) {
   m_chart = new ChartView(*m_model, this);
   m_security_widget->set_widget(m_chart);
   m_chart->installEventFilter(this);
+  m_auto_scale_button->connect_clicked_signal([=] {
+    m_is_chart_auto_scaled = !m_is_chart_auto_scaled;
+    m_chart->set_auto_scale(m_is_chart_auto_scaled);
```
kranar: Define a method and call that method.

spireken: Did you say before that we define methods for all connect_x() functions?

kranar: I don't think I specifically said that before. It's not strictly about the length, it's about readability. Code in the constructor should focus only on getting things setup. How things work should be kept out of it.

PR #315, opened 2019-02-22
Fb1384

spireken:


Applications/spire/include/spire/charting/cached_chart_model.hpp
```diff
@@ -0,0 +1,49 @@
...
+namespace Spire {
+
+  //! Provides a caching mechanism for a ChartModel.
+  class CachedChartModel : public ChartModel {
+    public:
+
+      //! Signals an update to a candlestick.
+      using CandlestickSignal = Signal<void (const Candlestick& candle)>;
+
+      struct ChartRange {
```
kranar: Document the public API.

Applications/spire/include/spire/charting/cached_chart_model.hpp
```diff
@@ -0,0 +1,49 @@
...
+      CachedChartModel(ChartModel& model);
+
+      ChartValue::Type get_x_axis_type() const override;
+
+      ChartValue::Type get_y_axis_type() const override;
+
+      QtPromise<std::vector<Candlestick>> load(ChartValue first,
+        ChartValue last) override;
+
+      //! Connects a slot to the candlestick signal.
```
kranar: Remove documentation from overwritten methods unless the documentation adds new information not found in the base class.

Applications/spire/include/spire/charting/chart_view.hpp
```diff
@@ -73,7 +73,7 @@ namespace Spire {
       void showEvent(QShowEvent* event) override;
 
     private:
-      ChartModel* m_model;
+      std::unique_ptr<ChartModel> m_model;
```
kranar: How did this become a ```unique_ptr```? This is most certainly a bug.

Applications/spire/source/charting/cached_chart_model.cpp
```diff
@@ -0,0 +1,89 @@
...
+  return m_chart_model->get_x_axis_type();
+}
+
+ChartValue::Type CachedChartModel::get_y_axis_type() const {
+  return m_chart_model->get_y_axis_type();
+}
+
+QtPromise<std::vector<Candlestick>> CachedChartModel::load(ChartValue first,
+    ChartValue last) {
+  for(auto& range : m_ranges) {
```
kranar: Perform a binary search, linear search is too slow.

Applications/spire/source/charting/cached_chart_model.cpp
```diff
@@ -0,0 +1,89 @@
...
+        [=] (const auto& value, const auto& search_value) {
+          return value.GetStart() < search_value;
+      });
+      auto last_iterator = std::lower_bound(m_loaded_data.begin(),
+        m_loaded_data.end(), last,
+        [=] (const auto& value, const auto& search_value) {
+          return value.GetEnd() <= search_value;
+      });
+      return QtPromise([=] {
+        return std::vector<Candlestick>(first_iterator, last_iterator);
```
kranar: This is a crash waiting to happen.

Applications/spire/source/charting/cached_chart_model.cpp
```diff
@@ -0,0 +1,89 @@
...
+connection CachedChartModel::connect_candlestick_slot(
+    const CandlestickSignal::slot_type& slot) const {
+  return m_chart_model->connect_candlestick_slot(slot);
+}
+
+QtPromise<std::vector<Candlestick>> CachedChartModel::load_data(
+    const ChartRange& data) {
+  return m_chart_model->load(data.m_start, data.m_end).then(
+    [=] (auto result) {
+      if(m_loaded_data.empty()) {
```
kranar: Find a way to implement this without testing if the list is empty. Unify the solution to reduce the number of cases you need to evaluate.

Applications/spire/source/charting/cached_chart_model.cpp
```diff
@@ -0,0 +1,89 @@
...
+          });
+        auto last = std::lower_bound(m_loaded_data.begin(),
+          m_loaded_data.end(), data.m_end, [=] (const auto& lhs, const auto& rhs) {
+            return lhs.GetStart() < rhs;
+          });
+        auto pos = m_loaded_data.erase(first, last);
+        m_loaded_data.insert(pos, result.Get().begin(), result.Get().end());
+      }
+      if(m_ranges.empty()) {
+        m_ranges.push_back(data);
```
kranar: Same goes for this, unify the solution to work for empty and non-empty lists.

Applications/spire/source/charting/cached_chart_model.cpp
```diff
@@ -0,0 +1,89 @@
...
+      } else {
+        auto ranges = m_ranges;
```
kranar: There is no need for making a copy, copies are expensive.

Applications/spire/source/charting/cached_chart_model.cpp
```diff
@@ -0,0 +1,89 @@
...
+        ranges.push_back(data);
+        std::sort(ranges.begin(), ranges.end(),
```
kranar: Sorting all data on every single load is too expensive, sorting is O(nlogn).

Applications/spire/source/charting/chart_view.cpp
```diff
@@ -5,7 +5,7 @@
 #include <QPainter>
 #include <QPaintEvent>
 #include "Nexus/Definitions/Money.hpp"
-#include "spire/charting/chart_model.hpp"
+#include "spire/charting/cached_chart_model.hpp"
```
kranar: The ChartView shouldn't know or care about caching, remove any references to caching from the ChartView.

Applications/spire/source/charting/cached_chart_model.cpp
spireken: Will I have to implement this myself? I haven't found anything suitable in the standard library.

kranar:
std lower_bound

spireken:
And then verify the returned iterator?

kranar:
lower_bound is guaranteed to be correct.

Applications/spire/source/charting/cached_chart_model.cpp
spireken: How so?

kranar:
You can assume lowerbound doesnt have a bug in it.

Applications/spire/include/spire/charting/cached_chart_model.hpp
kranar: Fully document the public API.

Applications/spire/source/charting/cached_chart_model.cpp
```diff
@@ -0,0 +1,85 @@
...
+  return m_chart_model->get_x_axis_type();
+}
+
+ChartValue::Type CachedChartModel::get_y_axis_type() const {
+  return m_chart_model->get_y_axis_type();
+}
+
+QtPromise<std::vector<Candlestick>> CachedChartModel::load(ChartValue first,
+    ChartValue last) {
+  auto range = ChartRange({first, last});
```
kranar: If you want to use aggregate initialization then remove the round brackets.

Applications/spire/source/charting/cached_chart_model.cpp
```diff
@@ -0,0 +1,85 @@
...
+  auto range_iterator = std::lower_bound(m_ranges.begin(), m_ranges.end(),
+    range, [=] (const auto& value, const auto& search_value) {
+      return value.m_start >= search_value.m_start &&
```
kranar: Why is the lambda capturing by value?

This is also not a valid comparator. A comparator must define a strict weak ordering relation, in particular it must be the case that ```forall a: comp(a, a) == false```, however the comparator you provide has the property that ```forall a: comp(a, a) == true```:

https://en.cppreference.com/w/cpp/named_req/Compare

Applications/spire/source/charting/cached_chart_model.cpp
```diff
@@ -0,0 +1,85 @@
...
+    return QtPromise([=] {
+      auto first_iterator = std::lower_bound(m_loaded_data.begin(),
+        m_loaded_data.end(), first,
+        [=] (const auto& value, const auto& search_value) {
+          return value.GetStart() < search_value;
+      });
+      auto last_iterator = std::lower_bound(m_loaded_data.begin(),
+        m_loaded_data.end(), last,
+        [=] (const auto& value, const auto& search_value) {
+          return value.GetEnd() <= search_value;
```
kranar: This is not a valid comparator, see above for note about strict weak orderings.

Applications/spire/source/charting/local_chart_model.cpp
```diff
@@ -25,14 +25,14 @@ ChartValue::Type LocalChartModel::get_y_axis_type() const {
...
+  return QtPromise([=, candlesticks = m_candlesticks] {
     auto start = std::lower_bound(candlesticks.begin(), candlesticks.end(),
       first, [] (const auto& lhs, const auto& rhs) {
-        return lhs.GetEnd() < rhs;
+        return lhs.GetStart() < rhs;
       });
     auto end = std::lower_bound(start, candlesticks.end(),
       last, [] (const auto& lhs, const auto& rhs) {
-        return lhs.GetStart() < rhs;
+        return lhs.GetEnd() <= rhs;
```
kranar: See note about what constitutes a valid comparator.

Applications/spire/source/charting/cached_chart_model.cpp
spireken: What difference is capturing by value or reference here when it's not using any variables from the outer scope in the body?

kranar: If you're not capturing anything then omit the capture entirely, ie, [] (const auto& lhs, const auto& rhs) { ... }.

Applications/spire/source/charting/cached_chart_model.cpp
```diff
@@ -16,24 +16,24 @@ ChartValue::Type CachedChartModel::get_y_axis_type() const {
...
 QtPromise<std::vector<Candlestick>> CachedChartModel::load(ChartValue first,
     ChartValue last) {
-  auto range = ChartRange({first, last});
+  auto range = ChartRange{first, last};
   auto range_iterator = std::lower_bound(m_ranges.begin(), m_ranges.end(),
-    range, [=] (const auto& value, const auto& search_value) {
-      return value.m_start >= search_value.m_start &&
-        value.m_end <= search_value.m_end;
+    range, [] (const auto& value, const auto& search_value) {
+      return value.m_start > search_value.m_start &&
```
kranar: This is still not a valid comparator.

Applications/spire/source/charting/cached_chart_model.cpp
```diff
@@ -16,24 +16,24 @@ ChartValue::Type CachedChartModel::get_y_axis_type() const {
...
+        [] (const auto& value, const auto& search_value) {
           return value.GetStart() < search_value;
       });
-      auto last_iterator = std::lower_bound(m_loaded_data.begin(),
+      auto last_iterator = std::upper_bound(m_loaded_data.begin(),
         m_loaded_data.end(), last,
-        [=] (const auto& value, const auto& search_value) {
-          return value.GetEnd() <= search_value;
+        [] (const auto& search_value, const auto& value) {
+          return value.GetEnd() > search_value;
```
kranar: Orderings should be defined in terms of ```operator <``` rather than ```operator >```. Reverse the operators to make it that way.

Applications/spire/source/charting/cached_chart_model.cpp
```diff
@@ -51,23 +51,23 @@ QtPromise<std::vector<Candlestick>> CachedChartModel::load_data(
...
           return lhs.GetStart() < rhs;
         });
       auto pos = m_loaded_data.erase(first, last);
       m_loaded_data.insert(pos, result.Get().begin(), result.Get().end());
       auto range_first = std::lower_bound(m_ranges.begin(), m_ranges.end(),
-        data, [=] (const auto& lhs, const auto& rhs) {
+        data, [] (const auto& lhs, const auto& rhs) {
           return lhs.m_start < rhs.m_start;
         });
       auto range_last = std::lower_bound(m_ranges.begin(), m_ranges.end(),
```
kranar: Your comparators are inconsistent. Either ranges is sorted by ```m_start``` or it's sorted by ```m_end```. You will need to formalize how elements are stored in ```m_ranges``` and apply that ordering consistently.

Applications/spire/source/charting/cached_chart_model.cpp
spireken: So is this valid:

return search_value < value.GetEnd();

I can't find a way to do it using value.GetStart() for the comparison.

kranar: It's valid in the sense that ```a < b``` is semantically equal to ```b > a```.

spireken: But it's not valid here?

kranar: In terms of style guide, yes it's valid but it's likely buggy since you're defining multiple unrelated order relations and I don't see any part of the implementation that guarantees the ordering.

spireken: The ranges get sorted after loaded data is inserted into the loaded data vector.

kranar: Okay, as long as you can maintain both orderings at the same time then it's fine.

PR #318, opened 2019-02-28
Fb1384

spireken:


Applications/spire/include/spire/charting/cached_chart_model.hpp
```diff
@@ -0,0 +1,49 @@
...
+  //! Provides a caching mechanism for a ChartModel.
+  class CachedChartModel : public ChartModel {
+    public:
+
+      //! Signals an update to a candlestick.
+      using CandlestickSignal = Signal<void (const Candlestick& candle)>;
+
+      //! Represents a range of ChartValues.
+      struct ChartRange {
+        ChartValue m_start;
```
kranar: Document the public API.

Applications/spire/source/charting/cached_chart_model.cpp
```diff
@@ -0,0 +1,125 @@
...
+QtPromise<std::vector<Candlestick>> CachedChartModel::load(ChartValue first,
+    ChartValue last) {
+  for(auto range = m_ranges.begin(); range != m_ranges.end(); ++range) {
+    if((*range).m_start <= first && (*range).m_end >= last) {
+      return QtPromise([=] {
+        auto first_iterator = std::lower_bound(m_loaded_data.begin(),
+          m_loaded_data.end(), first,
+          [] (const auto& index, const auto& value) {
+            return index.GetEnd() < value;
+        });
```
kranar: Indent line 26 and line 31.

Applications/spire/source/charting/cached_chart_model.cpp
```diff
@@ -0,0 +1,125 @@
...
+        m_loaded_data.end(), data.m_start, [] (const auto& index,
+            const auto& value) {
+          return index.GetStart() < value;
+        });
+      auto last = std::lower_bound(m_loaded_data.begin(),
+        m_loaded_data.end(), data.m_end, [] (const auto& index,
+            const auto& value) {
+          return index.GetStart() < value;
+        });
+      auto pos = m_loaded_data.end();
```
kranar: Initialize ```pos``` using IIFE.

Applications/spire/source/charting/cached_chart_model.cpp
```diff
@@ -0,0 +1,125 @@
...
+    if((*loaded_range).m_start <= range.m_start &&
+        (*loaded_range).m_end >= range.m_start) {
+      first = loaded_range;
+    }
+    if((*loaded_range).m_start <= range.m_end &&
+        (*loaded_range).m_end >= range.m_end) {
+      last = loaded_range + 1;
+    }
+    if(!first.has_value() && range.m_start > (*loaded_range).m_end &&
+        range.m_start < (*(loaded_range + 1)).m_start) {
```
kranar: How do you know ```loaded_range + 1``` is valid?

Applications/spire/source/charting/cached_chart_model.cpp
```diff
@@ -0,0 +1,125 @@
...
+    }
+    if(!last.has_value() && range.m_end > (*loaded_range).m_end &&
+        range.m_end < (*(loaded_range + 1)).m_start) {
+      last = loaded_range + 1;
+    }
+    if(first.has_value() && last.has_value()) {
+      break;
+    }
+  }
+  return {first.value(), last.value()};
```
kranar: How do you know ```first``` and ```last``` have a ```value()```?

spireken: Didn't we exhaustively prove that by the time we've reached line 124, first and last have to have values?

kranar: No and if you did, feel free to show it.

spireken: The cases we had were:

1. The requested range, P, is a subset of a range in the previously loaded ranges, R. This can't be the case in load_data() because it's handled before load_data() can be called.

2. R is empty, so (R.end, R.end) is returned.

3. P is to the left of R[0].

4. P is to the right of R.end().

5. P.start is to the left of R[0].

6. P.end is to the right of R.end().

7. P.start and P.end both fall within a range in R.

8. P.start is within a range in R, and P.end is in between two ranges.

9. P.start is between two ranges, and P.end is within a range in R.

10. Both P.start and P.end are between a range in R.

As far as I can tell this covers every possible case, so at line 124 first and last can't be null.

kranar: If ```P``` is to the right of ```R.end()```, then what are ```first``` and ```last``` initialized to?

Applications/spire/source/charting/cached_chart_model.cpp
spireken: On line 110, this only executes when P.end is within a range, so at the very most range +1 == R.end().

For the dereference on line 113: execution returns on line 98 if range.start > m_ranges.back().end, and line 106 assigns first if loaded_range is between m_ranges.back().start and m_ranges.back().end, so by line 113 loaded_range's index is no greater than m_ranges.back() - 1.

It's the same idea for line 117.

Applications/spire/source/charting/cached_chart_model.cpp
spireken: That case is evaluated on lines 96 and 98. In that case last = m_ranges.end(), and (last, last) is returned.

PR #327, opened 2019-03-22
Fb1403

spireken:


Applications/spire/source/charting/cached_chart_model.cpp
```diff
@@ -36,90 +37,135 @@ QtPromise<std::vector<Candlestick>> CachedChartModel::load(ChartValue first,
...
     const CandlestickSignal::slot_type& slot) const {
   return m_chart_model->connect_candlestick_slot(slot);
 }
 
+std::vector<ChartRange> CachedChartModel::get_gaps(ChartValue first,
+    ChartValue last) {
+  if(m_ranges.empty()) {
+    return {{ChartRange{first, last}}};
+  }
+  auto range_points = std::vector<ChartValue>({first, last});
```
kranar: What is the purpose of the ```vector```? A ```vector``` is used for a dynamically resizable list of elements, but ```range_points``` only ever contains two elements. Use two well named variables instead of a ```vector```.

Applications/spire/source/charting/cached_chart_model.cpp
```diff
@@ -36,90 +37,135 @@ QtPromise<std::vector<Candlestick>> CachedChartModel::load(ChartValue first,
...
+    }
+    if(range.m_start <= first && range.m_end >= first) {
+      range_points.front() = range.m_end;
+    } else if(range.m_start <= last && range.m_end >= last) {
+      range_points.back() = range.m_start;
+    }
+  }
+  auto gaps = std::vector<ChartRange>();
+  for(auto i = std::size_t(0); i < range_points.size(); i += 2) {
+    gaps.push_back({range_points[i], range_points[i + 1]});
```
kranar: I suspect there is a bug here.

Applications/spire/source/charting/cached_chart_model.cpp
```diff
@@ -36,90 +37,135 @@ QtPromise<std::vector<Candlestick>> CachedChartModel::load(ChartValue first,
...
+void CachedChartModel::insert_data(const std::vector<Candlestick>& data) {
+  if(data.empty()) {
+    return;
+  }
+  auto first = std::lower_bound(m_loaded_data.begin(),
+    m_loaded_data.end(), data.front().GetStart(), [] (const auto& index,
+        const auto& value) {
+      return index.GetStart() < value;
+    });
+  auto last = std::lower_bound(m_loaded_data.begin(),
```
kranar: Which of the following ranges do you want to erase?

[first, last]
(first, last)
[first, last)
(first, last]

Applications/spire/source/charting/cached_chart_model.cpp
spireken: I forgot to add an else statement that was in the pseudocode that adds more points to the range_points vector.

Because m_loaded_data is erased between [A, B], where A and B represents the newly loaded data, this issue doesn't get caught in the tests; it just makes the cache less efficient. So the 'bug' here is that the cache is loading data it already has, then overwriting it when the new data is merged.

Applications/spire/source/charting/cached_chart_model.cpp
spireken: I want to erase [first, last].

The assignment of 'last' gets the element past the element I'm looking for. So in when I call erase I'm basically saying erase [first, last + 1).

Applications/spire/source/charting/cached_chart_model.cpp
kranar: Add a unit test and ensure your test fails, then implement the fix.

spireken: I don't think I can write a test that fails for this because it will always work, it's just inefficient.

kranar:
If it always works then remove the use of the vector.

spireken:
I added the missing else statement that makes use of the range_points vector.

Applications/spire/source/charting/cached_chart_model.cpp
```diff
@@ -62,6 +61,9 @@ std::vector<ChartRange> CachedChartModel::get_gaps(ChartValue first,
       range_points.front() = range.m_end;
     } else if(range.m_start <= last && range.m_end >= last) {
       range_points.back() = range.m_start;
+    } else {
+      range_points.insert(range_points.end() - 1, range.m_start);
```
kranar: Remove this snippet of code, write a unit test that fails, commit that test to git. Reinsert this snippet of code, verify that the unit test passes, commit the changes to git. Push your changes to Github.

spireken: I think this goes back to what you said about the tests, in that the tests can't verify that the cache is actually doing anything, only that load() returns the correct result. In this case whether I remove this snippet or not the model will return the correct result.

kranar:
You can test caching using the test model class that tells you exactly what
loads are being requested from the source.

PR #332, opened 2019-03-28
Fb1433

spireken:


Applications/spire/source/common/local_technicals_model.cpp
```diff
@@ -0,0 +1,105 @@
+#include "spire/spire/local_technicals_model.hpp"
+
+using namespace boost;
+using namespace boost::signals2;
+using namespace Nexus;
+using namespace Spire;
+
+LocalTechnicalsModel::LocalTechnicalsModel(Security security)
+    : m_security(security) {}
```
kranar: Avoid a copy by using a move.

Applications/spire/source/common/local_technicals_model.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+void LocalTechnicalsModel::set_close(Money price) {
+  m_close = price;
+  m_close_signal(*m_close);
+  m_last_price = m_close;
```
kranar: This is a bug.

Applications/spire/source/common/local_technicals_model.cpp
```diff
@@ -0,0 +1,105 @@
...
+  m_last_price_signal(*m_last_price);
+}
+
+void LocalTechnicalsModel::update(const TimeAndSale& time_and_sale) {
+  if(!m_open.is_initialized()) {
```
kranar: Emit signals after you have updated everything based on the ```time_and_sale```. This avoids inconsistencies in the model.

Applications/spire/source/common/local_technicals_model.cpp
```diff
@@ -0,0 +1,105 @@
...
+optional<Money> LocalTechnicalsModel::get_last_price() const {
+  return m_last_price;
+}
+
+Quantity LocalTechnicalsModel::get_volume() const {
+  return m_volume;
+}
+
+QtPromise<void> LocalTechnicalsModel::load() {
+  m_volume = Quantity(0);
```
kranar: What is the purpose of this?

Applications/spire/source/common/local_technicals_model.cpp
```diff
@@ -0,0 +1,105 @@
...
+  m_volume_signal(m_volume);
+  return QtPromise<void>();
```
kranar: This will cause a crash.

Applications/spire/source/common/local_technicals_model.cpp
spireken: So regardless of what changes just update everything?

Applications/spire/source/common/local_technicals_model.cpp
spireken: I was trying to simulate what a load would do.

Applications/spire/source/common/local_technicals_model.cpp
kranar: Only emit signals for values that have actually updated.

Applications/spire/source/common/local_technicals_model.cpp
kranar: A load should not reset a security's volume.

Applications/spire/source/common/local_technicals_model.cpp
kranar: To fix the crash do this:

  return QtPromise([] {});

It's basically a promise that does nothing.

Applications/spire/source/common/local_technicals_model.cpp
spireken: Are there cases where the time and sales volume is 0?

kranar: I can't think of any, how will this impact your solution?

spireken: You were saying only emit signals if the values changed so I assumed you meant the last price, which may be the same as the previous last price.

I was just wondering if there was special cases where a time and sale could happen without a change in volume.

kranar: It's safe to assume that all time and sales have a quantity greater than 0.

Applications/spire/include/spire/spire/local_technicals_model.hpp
```diff
@@ -76,6 +76,10 @@ namespace Spire {
       boost::optional<Nexus::Money> m_close;
       boost::optional<Nexus::Money> m_last_price;
       Nexus::Quantity m_volume;
+      bool m_high_changed;
```
kranar: These are not neccessary, consider a solution that omits these fields.

Applications/spire/source/common/local_technicals_model.cpp
```diff
@@ -6,37 +6,54 @@ using namespace Nexus;
...
 
 LocalTechnicalsModel::LocalTechnicalsModel(Security security)
-    : m_security(security) {}
+    : m_security(std::move(security)),
+      m_volume(0),
+      m_high_changed(false),
+      m_low_changed(false),
+      m_open_changed(false),
+      m_last_price_changed(false) {
+  m_volume_signal(m_volume);
```
kranar: Who will be receiving this signal?

Applications/spire/source/spire_tester/local_technicals_model_tester.cpp
```diff
@@ -15,6 +15,11 @@ namespace {
   }
 }
 
+TEST_CASE("asdf", "asdf") {
```
kranar: This unit test isn't doing anything.

Applications/spire/source/common/local_technicals_model.cpp
spireken: So should this be emitted on load then?

kranar: The volume signal is emitted when the volume is updated.

spireken: So the default value doesn't count as updating it?

kranar: Sure you can consider that, but then ask yourself who/what will be receiving the signal in the constructor? How can it be possible that someone will receive a signal before an object has been constructed?

spireken: I know technically it's just like yelling into a void - there's no one there to receive it.

But if I can't signal the volume on construction then it's either I can emit it when the model loads or it may be the case that a signal isn't emitted at all when the volume is 0.

kranar: Don't emit anything.

Applications/spire/source/common/local_technicals_model.cpp
```diff
@@ -7,54 +7,38 @@ using namespace Spire;
...
   m_close_signal(*m_close);
 }
 
 void LocalTechnicalsModel::update(const TimeAndSale& time_and_sale) {
-  if(!m_open.is_initialized()) {
-    m_open = time_and_sale.m_price;
-    m_open_changed = true;
-  }
-  if(time_and_sale.m_price != m_last_price) {
+  if(!m_last_price.is_initialized() || time_and_sale.m_price != m_last_price) {
```
kranar: Use the following reference to eliminate checks to ```is_initialized```:

https://www.boost.org/doc/libs/1_69_0/libs/optional/doc/html/boost_optional/tutorial/relational_operators.html

Applications/spire/source/common/local_technicals_model.cpp
```diff
@@ -0,0 +1,104 @@
...
+  }
+  if(!m_low.is_initialized() || time_and_sale.m_price < *m_low) {
+    m_low = time_and_sale.m_price;
+  }
+  m_volume += time_and_sale.m_size;
+  if(!m_open.is_initialized()) {
+    m_open = time_and_sale.m_price;
+    m_open_signal(*m_open);
+  }
+  if(*m_last_price != time_and_sale.m_price || *m_low == *m_high) {
```
kranar: This does not look correct, what if ```m_last_price == time_and_sale.price && m_low == m_high``` then you will be emitting signals that are not necessary.

Add appropriate unit tests to test and correct this issue.

Applications/spire/source/common/local_technicals_model.cpp
```diff
@@ -15,28 +15,27 @@ void LocalTechnicalsModel::set_close(Money price) {
...
-  m_volume += time_and_sale.m_size;
-  if(!m_open.is_initialized()) {
-    m_open = time_and_sale.m_price;
     m_open_signal(*m_open);
-  }
-  if(*m_last_price != time_and_sale.m_price || *m_low == *m_high) {
     m_last_price_signal(*m_last_price);
-    if(*m_high == time_and_sale.m_price) {
+    m_high_signal(*m_high);
+    m_low_signal(*m_low);
```
kranar: Emit all signals after all data has been updated to avoid inconsistencies.

spireken: Do you mean after line 27 emit the volume signal, then after line 38 emit it?

PR #345, opened 2019-05-09
Fb1466

spireken:
I've added '_qt5121' to the new builds in case I need to compare them later to the Qt 5.11.x builds.


spireken:
I've updated the builds to be called _qt5121_fb1466 to avoid having them overwritten by the other pull request builds.

PR #346, opened 2019-05-10
Removed table row minimum size restriction on Book View and Time and …

spireken:
…Sales windows


kranar:
The demo's don't show a difference between the two.

spireken:
The old ones and new ones should look the same. I just fixed the issue where Qt was setting a minimum size for the rows.

spireken:
Sorry, the builds for this branch are just called ..._qt5121. I only labelled the ones for fb1466.

kranar:
Which demo's should I be comparing to ensure that this is fixed?

spireken:
book_view_ui_tester and book_view_ui_tester_qt5121

kranar:
Good job.

PR #352, opened 2019-06-11
Line editing

spireken:
PR for the Draw TrendLines task.


Applications/Spire/Include/Spire/Charting/ChartView.hpp
```diff
@@ -95,9 +141,28 @@ namespace Spire {
...
+      QCursor m_hand_cursor;
+      TrendLineModel m_trend_line_model;
+      DrawState m_draw_state;
+      int m_current_trend_line_id;
+      ChartPoint m_current_trend_line_point;
+      ChartPoint m_current_stationary_point;
+      QColor m_current_trend_line_color;
+      TrendLineStyle m_current_trend_line_style;
+      int m_line_hover_distance_squared;
+      bool m_multi_select;
```
kranar: Booleans must be named "is/has/etc..."

Applications/Spire/Include/Spire/Charting/Charting.hpp
```diff
@@ -15,11 +16,109 @@ namespace Spire {
...
+  //! Returns the slope of the line reprsented by the given points.
+  /*!
+    \param x1 The x-coordinate of the first point.
+    \param y1 The y-cooridnate of the first point.
+    \param x2 The x-coordinate of the second point.
+    \param y2 The y-coordinate of the second point.
+  */
+  template <typename T>
+  T slope(T x1, T y1, T x2, T y2) {
+    if(x1 - x2 == 0) {
```
kranar: There's no need to check for this.

Applications/Spire/Include/Spire/Charting/TrendLineModel.hpp
```diff
@@ -52,18 +50,28 @@ namespace Spire {
...
+        \param id the id of the trend line to set as unselected.
       */
       void unset_selected(int id);
 
+      //! Inverts the selected status of the specified item.
+      //! Does nothing if the specified id doesn't exist.
+      /*!
+        \param id The id of the trend line.
+      */
+      void invert_selection(int id);
```
kranar: The commonly used term is ```toggle```, ```toggle_selection```.

Applications/Spire/Include/Spire/Charting/TrendLineModel.hpp
```diff
@@ -88,6 +96,7 @@ namespace Spire {
       std::vector<TrendLineEntry> m_trend_lines;
 
       std::vector<TrendLineEntry>::iterator find_id(int id);
+      bool is_selected(int id);
```
kranar: Why isn't this method ```const```?

Applications/Spire/Source/Charting/ChartView.cpp
```diff
@@ -132,6 +195,47 @@ void ChartView::set_auto_scale(bool auto_scale) {
   }
 }
 
+bool ChartView::get_draw_mode() const {
+  if(m_draw_state == DrawState::OFF) {
```
kranar: Simplify this.

Applications/Spire/Source/Charting/ChartingWindow.cpp
```diff
@@ -246,8 +270,10 @@ void ChartingWindow::on_auto_scale_button_click() {
 
 void ChartingWindow::on_draw_line_button_click() {
   if(m_trend_line_editor_widget->isVisible()) {
```
kranar: Factor all this out by using ```setVisible```

Applications/Spire/Source/Charting/TrendLineModel.cpp
```diff
@@ -180,6 +151,17 @@ std::vector<TrendLineModel::TrendLineEntry>::iterator
       });
 }
 
+bool TrendLineModel::is_selected(int id) {
+  auto selected = std::find_if(m_trend_lines.begin(), m_trend_lines.end(),
+    [&] (auto& line) {
+      return id == line.m_id && line.m_state == State::SELECTED;
+    });
+  if(selected == m_trend_lines.end()) {
```
kranar: Factor this out.

Applications/Spire/Include/Spire/Charting/Charting.hpp
spireken: I get division by zero if I don't, should I just throw an exception?

kranar: Division by zero is exactly what you want right?

Applications/Spire/Include/Spire/Charting/TrendLineModel.hpp
spireken: It uses find_if(), which isn't const. This came up before for another function and you said to just make it non-const.

Applications/Spire/Include/Spire/Charting/Charting.hpp
spireken: Won't it just crash? If it's going to have undefined behavior don't I want to define what happens for that case?

Applications/Spire/Include/Spire/Charting/TrendLineModel.hpp
kranar: ```find_if``` is ```const```.

Applications/Spire/Include/Spire/Charting/Charting.hpp
kranar: It won't crash.

spireken: Okay, I see, it crashed when I was using ints, just saw that x/0 is technically defined when x =/= 0.

Applications/Spire/Include/Spire/Charting/ChartView.hpp
```diff
@@ -95,9 +141,28 @@ namespace Spire {
       bool m_is_auto_scaled;
       QtPromise<std::vector<Spire::Candlestick>> m_candlestick_promise;
       std::vector<Candlestick> m_candlesticks;
+      QCursor m_hand_cursor;
```
kranar: Look into moving some of these members out of the class. Chances are every view uses the same QCursor, so use a function:

```
namespace {
  auto get_hand_cursor() {
    static auto cursor = ...;
    return cursor;
  }
}
```

Applications/Spire/Include/Spire/Charting/Charting.hpp
```diff
@@ -1,5 +1,6 @@
 #ifndef SPIRE_CHARTING_HPP
 #define SPIRE_CHARTING_HPP
+#include "Nexus/Definitions/Quantity.hpp"
```
kranar: What is this used for?

Applications/Spire/Include/Spire/Charting/Charting.hpp
```diff
@@ -15,11 +16,106 @@ namespace Spire {
...
+
+  //! Returns the slope of the line reprsented by the given points.
+  /*!
+    \param x1 The x-coordinate of the first point.
+    \param y1 The y-cooridnate of the first point.
+    \param x2 The x-coordinate of the second point.
+    \param y2 The y-coordinate of the second point.
+  */
+  template <typename T>
+  T slope(T x1, T y1, T x2, T y2) {
```
kranar: Use a ```double``` for this instead of a template for all of these functions. Instead of passing in x and y values, look into using ```QPointF```.

Applications/Spire/Source/Charting/ChartView.cpp
```diff
@@ -272,6 +432,65 @@ void ChartView::update_auto_scale() {
...
+  } else {
+    m_current_trend_line_point = ChartPoint();
+    m_current_stationary_point = ChartPoint();
+  }
+  auto distance = std::numeric_limits<double>::infinity();
+  if(std::isinf<double>(line_slope)) {
+    if(is_within_interval(mouse_y, point1_y, point2_y)) {
+      distance = distance_squared(mouse_x, mouse_y, point1_x, mouse_y);
+    }
+  } else if(line_slope == Quantity(0)) {
```
kranar: Why is this being compared to a ```Quantity```?

Applications/Spire/Include/Spire/Charting/Charting.hpp
```diff
@@ -15,11 +16,106 @@ namespace Spire {
...
+  template <typename T>
+  bool is_within_interval(T value, T start, T end) {
+    return is_within_interval(value, start, end, static_cast<T>(0));
+  }
+
+  //! Calculates the y-intercept of a line, given a point and the slope.
+  /*!
+    \param x The x-coordinate of the given point.
+    \param y The y-coordinate of the given point.
+    \param slope The slope ofthe given line.
```
kranar: Fix typo.

Applications/Spire/Include/Spire/Charting/Charting.hpp
```diff
@@ -15,11 +16,106 @@ namespace Spire {
...
+    \param m The slope of the line.
+    \param x The x value.
+    \param b The y-intercept of the line.
+  */
+  template <typename T>
+  T calculate_y(T m, T x, T b) {
+    return m * x + b;
+  }
+
+  //! Returns the shortest distance between a point and two given points.
```
kranar: Shortest squared distance.

Applications/Spire/Include/Spire/Charting/TrendLineStyleDropDownMenu.hpp
```diff
@@ -10,6 +10,9 @@ namespace Spire {
   class TrendLineStyleDropDownMenu : public QWidget {
     public:
 
+      //! Signals that a trend line style was selected.
+      using StyleSignal = Signal<void (TrendLineStyle)>;
```
kranar: Document parameters.

Applications/Spire/Include/Spire/Charting/Charting.hpp
spireken: I can make these functions take doubles, but I'll have to convert ChartPoints to QPointFs in TrendLineModel if I use QPointFs as parameters.

kranar: Okay, use ```QPointF```s then.

Applications/Spire/Include/Spire/Charting/ChartView.hpp
spireken: Where would these go if they were in an anonymous namespace? Shouldn't it be the Ui header and source files?

kranar: ChartView.cpp

Applications/Spire/Include/Spire/Charting/Charting.hpp
spireken: Should I create an alias for: static_cast<double>(static_cast<ChartValue>(value))?

kranar: What's an alias?

spireken: I don't remember the exact syntax, but isn't it possible to do something like:

using point_cast = static_cast<double>(static_cast<ChartValue>());

kranar: No, use a function.

Applications/Spire/Include/Spire/Charting/Charting.hpp
```diff
@@ -1,6 +1,7 @@
 #ifndef SPIRE_CHARTING_HPP
 #define SPIRE_CHARTING_HPP
 #include "Nexus/TechnicalAnalysis/Candlestick.hpp"
+#include <QPointF>
```
kranar: System includes go before local includes.

PR #353, opened 2019-06-13
Chart rendering

spireken:


Applications/Spire/Source/Charting/ChartView.cpp
```diff
@@ -13,6 +13,8 @@ using namespace Nexus;
 using namespace Spire;
 
 namespace {
+  const auto GAP_DIVISOR = 2.95;
```
kranar: Make this local to the function using it.

spireken: Is there no advantage to declaring it in an anonymous namespace?

kranar: Use the most local scope as possible. If something is declared in an anonymous namespace, from my point of view it's a variable I have to be aware of and keep track of for the entire file. If it's local to a function, then I only need to be aware of it for that specific function, so on so forth.

Applications/Spire/Source/Charting/ChartView.cpp
```diff
@@ -278,16 +276,17 @@ void ChartView::paintEvent(QPaintEvent* event) {
       candlestick.GetOpen()});
     auto close = convert_chart_to_pixels({candlestick.GetEnd(),
       candlestick.GetClose()});
+    auto gap_divisor = 2.95;
```
kranar: If this is a constant then declare it as such and use CAMEL_CASE.

```
  const auto GAP_DIVISOR = 2.95;
```

PR #358, opened 2019-07-22
Gap reloading

spireken:


Applications/Spire/Include/Spire/Charting/ChartView.hpp
```diff
@@ -14,6 +14,52 @@ namespace Spire {
...
+        /** The start value of the gap. */
+        ChartValue m_start;
+
+        /** The end value of the gap. */
+        ChartValue m_end;
+      };
+
+      //! Represents an accumulation of loaded data and gaps, and stores values
+      //! that determine what to load next.
+      struct LoadedData {
```
kranar: Why are these ```struct```s public? No changes to the public API should be made without first proposing them.

Applications/Spire/Include/Spire/Charting/ChartView.hpp
```diff
@@ -116,13 +164,10 @@ namespace Spire {
         OFF,
         POINT
       };
-      struct Gap {
-        ChartValue m_start;
-        ChartValue m_end;
-      };
       ChartModel* m_model;
       ChartPoint m_top_left;
```
kranar: Are these two fields a region?

Applications/Spire/Source/Charting/ChartView.cpp
```diff
@@ -73,6 +78,56 @@ namespace {
...
+  }
+
+  QtPromise<ChartView::LoadedData> load_data(
+      QtPromise<std::vector<Candlestick>>& promise,
+      ChartView::LoadedData data, ChartModel* model) {
+    return promise.then([=] (auto result) mutable {
+      auto c = std::move(result.Get());
+      while(!c.empty() && !data.m_candlesticks.empty() &&
+          c.front().GetStart() <= data.m_candlesticks.back().GetStart()) {
+        c.erase(c.begin());
```
kranar: Too slow. Find another way.

Applications/Spire/Source/Charting/ChartView.cpp
```diff
@@ -73,6 +78,56 @@ namespace {
...
+      start = candlestick.GetEnd();
+    }
+    return gap_info;
+  }
+
+  QtPromise<ChartView::LoadedData> load_data(
+      QtPromise<std::vector<Candlestick>>& promise,
+      ChartView::LoadedData data, ChartModel* model) {
+    return promise.then([=] (auto result) mutable {
+      auto c = std::move(result.Get());
```
kranar: Use a better name.

Applications/Spire/Source/Charting/ChartView.cpp
```diff
@@ -73,6 +78,56 @@ namespace {
...
+        data.m_end - data.m_start - gap_info.total_gaps_value) /
+        data.m_values_per_pixel + gap_info.gap_count * GAP_SIZE())),
+        data.m_end_x);
+      data.m_start = data.m_end;
+      data.m_end += (data.m_end_x - data.m_current_x) * data.m_values_per_pixel;
+      if(data.m_current_x < data.m_end_x) {
+        auto new_promise = model->load(data.m_start, data.m_end);
+        return load_data(new_promise, data, model);
+      }
+      return QtPromise<ChartView::LoadedData>([=] {
```
kranar: Don't copy ```data``` into the promise, move it in.

Applications/Spire/Source/Charting/ChartView.cpp
```diff
@@ -227,24 +282,36 @@ void ChartView::set_region(const ChartPoint& top_left,
...
-  m_x_axis_step = calculate_step(m_model->get_x_axis_type(), m_x_range);
-  m_y_range = m_top_left.m_y - m_bottom_right.m_y;
-  m_y_axis_step = calculate_step(m_model->get_y_axis_type(), m_y_range);
-  update_origins();
-  m_candlestick_promise = m_model->load(m_top_left.m_x, m_bottom_right.m_x);
-  m_candlestick_promise.then([=] (auto result) {
-    m_candlesticks = std::move(result.Get());
-    update_gaps();
+  auto bottom_right_pixel_x = m_bottom_right_pixel.x();
+  auto required_data = LoadedData{};
```
kranar: Use ```LoadedData()```, no curly brackets.

Applications/Spire/Source/Charting/ChartView.cpp
```diff
@@ -227,24 +282,36 @@ void ChartView::set_region(const ChartPoint& top_left,
...
+  required_data.m_gaps = std::vector<Gap>();
+  required_data.m_start = m_top_left.m_x;
+  required_data.m_end = m_bottom_right.m_x;
+  required_data.m_current_x = 0;
+  required_data.m_end_x = m_bottom_right_pixel.x();
+  required_data.m_values_per_pixel = (m_bottom_right.m_x - m_top_left.m_x) /
+    bottom_right_pixel_x;
+  m_loaded_data_promise = load_data(m_model->load(required_data.m_start,
+    required_data.m_end), required_data, m_model);
+  m_loaded_data_promise.then([=] (auto result) {
```
kranar: This is a crash waiting to happen. ```then``` returns a new promise but you're not storing that promise anywhere.

Use:
```
m_loaded_data_promise = load_data(...).then(...);
```

Applications/Spire/Include/Spire/Charting/ChartView.hpp
spireken: Is there a way to make it accessible to the anonymous namespace and usable as a private value type member?

kranar: No, you will need to use a private static method.

spireken: Do I declare static members alphabetically with the other private methods, or does it go before or after them?

kranar: Good question, static methods go just above non-static methods.

Applications/Spire/Source/Charting/ChartView.cpp
spireken: So should then(...) return a value or should I make it like:

m_loaded_data_promise = load_data(...);
m_loaded_data_promise = m_loaded_data_promise.then(...);

kranar:
```
m_loaded_data_promise = load_data(m_model->load(required_data.m_start,
    required_data.m_end), required_data, m_model).then([=] (auto result) {
  ...
});
```

spireken: I get the compiler complaining that there's a mismatch of types:

 binary '=': no operator found which takes a right-hand operand of type 'void'

It appears to be evaluating the return value of then(...) as type: std::enable_if_t<true, void>

kranar: Okay keep it as is for now, we'll revisit it later.

Applications/Spire/Include/Spire/Charting/ChartView.hpp
```diff
@@ -197,6 +183,73 @@ namespace Spire {
       bool m_is_multi_select_enabled;
       std::vector<Gap> m_gaps;
 
+      static GapInfo update_gaps(std::vector<ChartView::Gap>& gaps,
+          std::vector<Candlestick>& candlesticks, ChartValue start) {
+        auto gap_info = GapInfo{0, ChartValue()};
```
kranar: Why is this in the header file?

kranar: Implement functions in the source file.

Applications/Spire/Include/Spire/Charting/ChartView.hpp
```diff
@@ -54,9 +64,8 @@ namespace Spire {
       //! Removes the crosshair from the chart.
       void reset_crosshair();
 
-      //! Returns a pair (top left, bottom right) of ChartPoints
-      //! representing the region of the ChartView.
-      std::tuple<ChartPoint, ChartPoint> get_region() const;
+      //! Returns the region of the ChartView.
+      ChartView::Region get_region() const;
```
kranar: Return by ```const&```.

Applications/Spire/Include/Spire/Charting/ChartView.hpp
```diff
@@ -66,6 +75,9 @@ namespace Spire {
       void set_region(const ChartPoint& top_left,
```
kranar: Update this method to work with a ```Region```.

Applications/Spire/Source/Charting/ChartView.cpp
```diff
@@ -52,6 +53,11 @@ namespace {
     return ChartValue();
   }
 
+  static const int GAP_SIZE() {
```
kranar: Don't simply copy and paste the change, undo the change exactly as it was.

Applications/Spire/Source/Charting/ChartView.cpp
```diff
@@ -216,35 +217,49 @@ void ChartView::reset_crosshair() {
...
-std::tuple<ChartPoint, ChartPoint> ChartView::get_region() const {
-  return {m_top_left, m_bottom_right};
+ChartView::Region ChartView::get_region() const {
+  return m_region;
 }
 
 void ChartView::set_region(const ChartPoint& top_left,
     const ChartPoint& bottom_right) {
-  if(top_left == m_top_left && bottom_right == m_bottom_right) {
+  if(top_left == m_region.m_top_left &&
```
kranar: Define ```operator ==``` and ```operator !=``` on ```Region```.

Applications/Spire/Source/Charting/ChartView.cpp
```diff
@@ -546,19 +562,71 @@ void ChartView::update_auto_scale() {
...
+      data.m_values_per_pixel + gap_info.gap_count * GAP_SIZE())),
+      data.m_end_x);
+    data.m_start = data.m_end;
+    data.m_end += (data.m_end_x - data.m_current_x) *
+      data.m_values_per_pixel;
+    if(data.m_current_x < data.m_end_x) {
+      auto new_promise = model->load(data.m_start, data.m_end);
+      return load_data(new_promise, std::move(data), model);
+    }
+    return QtPromise<LoadedData>([=] {
```
kranar: This is making an unneccesary copy of ```data```.

spireken: What's the difference between:

Promise([data_copy = std::move(data); ] { return std::move(data_copy); });

and:

Pormise([&] { return std::move(data); });

The second one was my initial idea, but I came across the first one on SO.

Applications/Spire/Source/Charting/ChartView.cpp
spireken: Should these be defined publicly or in the anonymous namespace?

spireken: Nevermind, it's a struct.

spireken: Question still stands; I forgot about the problem with global vs. member overloaded operators.

Applications/Spire/Source/Charting/ChartView.cpp
kranar: The first one is semantically correct because no copy of ```data``` is ever made. The second one is a crash waiting to happen because while there is no copy, you are capturing ```data``` by reference and that reference will get invalidated as soon as the function goes out of scope.

Applications/Spire/Source/Charting/ChartView.cpp
kranar: The operators should be methods added to ```Region```.

Applications/Spire/Include/Spire/Charting/ChartView.hpp
```diff
@@ -22,6 +22,16 @@ namespace Spire {
 
         /** The bottom right point of the region. */
         ChartPoint m_bottom_right;
+
+        bool operator ==(const Region& rhs) {
+          return m_top_left == rhs.m_top_left &&
```
kranar: Dont implement in a header file.

Applications/Spire/Source/Charting/ChartView.cpp
```diff
@@ -76,6 +76,16 @@ namespace {
...
 
+bool ChartView::Region::operator ==(const Region& rhs) const {
+  return m_top_left == rhs.m_top_left &&
+    m_bottom_right == rhs.m_bottom_right;
+}
+
+bool ChartView::Region::operator !=(const Region& rhs) const {
+  return m_top_left != rhs.m_top_left ||
+    m_bottom_right != rhs.m_bottom_right;
+}
```
kranar: Implement != in terms of ==

Applications/Spire/Source/Charting/ChartView.cpp
```diff
@@ -460,20 +482,80 @@ void ChartView::resizeEvent(QResizeEvent* event) {
...
+      new_candlesticks.begin(), new_candlesticks.end());
+    data.m_current_x += std::min(static_cast<int>(std::ceil((
+      data.m_end - data.m_start - gap_info.total_gaps_value) /
+      data.m_values_per_pixel + gap_info.gap_count * GAP_SIZE())),
+      data.m_end_x);
+    data.m_start = data.m_end;
+    data.m_end += (data.m_end_x - data.m_current_x) *
+      data.m_values_per_pixel;
+    if(data.m_current_x < data.m_end_x) {
+      auto new_promise = model->load(data.m_start, data.m_end);
```
kranar: Inline ```new_promise``` directly. No need to declare a variable that's only used once and doesn't add any semantic value.

Applications/Spire/Source/Charting/ChartView.cpp
```diff
@@ -460,20 +482,80 @@ void ChartView::resizeEvent(QResizeEvent* event) {
...
+      return load_data(new_promise, std::move(data), model);
+    }
+    return QtPromise<LoadedData>([data_copy = std::move(data)] {
```
kranar: ```[data = std::move(data)]``` is the preferred idiom for capturing by move.

You will also need to make this lambda ```mutable``` if you want line 554 to avoid a copy.

spireken:
Do you have any more changes for this PR?

Applications/Spire/Source/Charting/ChartView.cpp
```diff
@@ -546,19 +628,9 @@ void ChartView::update_auto_scale() {
     auto_scale_top = std::max(auto_scale_top, candle.GetHigh());
```
kranar: This is also undefined behavior. Create a task to sanitize all uses of ```std::min``` and ```std::max``` and fix them all in that task instead of this one. There are likely many instances of this.

Applications/Spire/Source/SpireTester/ChartViewTester.cpp
```diff
@@ -0,0 +1,154 @@
...
+  };
+
+  const auto GAP_SIZE = 35;
+
+  const auto TL() {
+    static auto tl = ChartPoint(ChartValue(900), ChartValue(1));
+    return tl;
+  }
+
+  const auto BR() {
```
kranar: Do not use abbreviations.

Applications/Spire/Source/Charting/ChartView.cpp
```diff
@@ -460,20 +482,80 @@ void ChartView::resizeEvent(QResizeEvent* event) {
...
+  }
+  return gap_info;
+}
+
+QtPromise<ChartView::LoadedData> ChartView::load_data(
+    QtPromise<std::vector<Candlestick>>& promise, LoadedData data,
+    ChartModel* model) {
+  return promise.then([=] (auto result) mutable {
+    auto new_candlesticks = std::move(result.Get());
+    if(!new_candlesticks.empty() && !data.m_candlesticks.empty() &&
```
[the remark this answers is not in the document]
spireken: I can remove everything else, but I'll at least need to check if data.m_candlesticks is empty, otherwise it may crash if back() is called on it in the find_if() lambda.

Applications/Spire/Source/Charting/ChartView.cpp
```diff
@@ -460,20 +482,80 @@ void ChartView::resizeEvent(QResizeEvent* event) {
...
   }
   QWidget::showEvent(event);
 }
 
+ChartView::GapInfo ChartView::update_gaps(std::vector<ChartView::Gap>& gaps,
+    std::vector<Candlestick>& candlesticks, ChartValue start) {
+  auto gap_info = GapInfo{0, ChartValue()};
+  for(auto& candlestick : candlesticks) {
+    auto end = candlestick.GetStart();
+    if(start != end) {
```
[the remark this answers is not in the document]
spireken: The spec says that a gap is defined by a gap greater than 10 candlesticks, but if the width's of them vary how can I determine this?

Applications/Spire/Source/Charting/ChartView.cpp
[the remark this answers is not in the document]
kranar: Okay simplify as much as possible.

Applications/Spire/Source/Charting/ChartView.cpp
```diff
@@ -460,20 +482,80 @@ void ChartView::resizeEvent(QResizeEvent* event) {
...
+        return ChartValue();
+      } else if(data.m_candlesticks.empty()) {
+        return new_candlesticks.front().GetStart();
+      }
+      return data.m_candlesticks.back().GetEnd();
+    }();
+    auto gap_info = update_gaps(data.m_gaps, new_candlesticks, last);
+    data.m_candlesticks.insert(data.m_candlesticks.end(),
+      new_candlesticks.begin(), new_candlesticks.end());
+    data.m_current_x += std::min(static_cast<int>(std::ceil((
```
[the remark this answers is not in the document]
spireken: If I understand this right I have to keep the first argument (the call to ceil()) as a variable to prevent it from being destroyed before min() can return it?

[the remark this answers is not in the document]
kranar: Yeah the whole thing is retarded, we'll fix all of these in one shot by writing our own min/max functions.

Applications/Spire/Source/Charting/ChartView.cpp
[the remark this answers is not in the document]
spireken: Do you want me to make a case for this until the chart supports changing the period?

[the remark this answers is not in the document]
kranar: Sure.

Applications/Spire/Source/Charting/ChartView.cpp
```diff
@@ -514,19 +510,16 @@ QtPromise<ChartView::LoadedData> ChartView::load_data(
...
-        data.m_candlesticks.back().GetStart()) {
+    if(!data.m_candlesticks.empty()) {
       auto last = std::find_if(new_candlesticks.begin(),
         new_candlesticks.end(), [&] (const auto& candlestick) {
           return candlestick.GetStart() >=
             data.m_candlesticks.back().GetStart();
         });
-      auto b = new_candlesticks.begin();
       if(last != new_candlesticks.end()) {
         last++;
```
kranar: Use ```++last```

PR #359, opened 2019-07-24
Implemented min() and max() functions, removed use of standard min an…

spireken:
…d max functions

### DIFF
--- a/Applications/Spire/Include/Spire/Spire/Utility.hpp
+++ b/Applications/Spire/Include/Spire/Spire/Utility.hpp
@@ -25,7 +25,7 @@ namespace Spire {
   */
   template<typename T>
   T max(const T& first, const T& second) {
-    if(second > first) {
+    if(first < second) {
       return second;
     }
     return first;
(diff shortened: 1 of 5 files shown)


Applications/Spire/Include/Spire/Spire/Utility.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+  //! Returns the maximum of the given parameters. Returns the first given
+  //! parameter if they are equal.
+  /*!
+    \param first The first value to compare.
+    \param second The second value to compare.
+  */
+  template<typename T>
+  T max(const T& first, const T& second) {
+    if(second > first) {
```
kranar: Use ```<``` since not all types define a ```>```.

spireken: Is that just convention or is there a technical reason for it?

kranar: Convention.

PR #360, opened 2019-07-25
Line dragging

spireken:
There's a problem when the chart is zoomed far out and a line is drawn past the last gap, but I made a separate case for that issue: https://app.asana.com/0/1125736618929093/1132917119979904/f

### DIFF
--- a/Applications/Spire/Source/Charting/ChartView.cpp
+++ b/Applications/Spire/Source/Charting/ChartView.cpp
@@ -223,7 +223,7 @@ void ChartView::set_crosshair(const QPoint& position,
 }
 
 void ChartView::reset_crosshair() {
-  m_crosshair_pos.reset();
+  m_crosshair_pos = std::nullopt;
 }
 
 const ChartView::Region& ChartView::get_region() const {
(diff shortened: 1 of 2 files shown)


Applications/Spire/Source/Charting/ChartView.cpp
```diff
@@ -771,6 +769,7 @@ void ChartView::on_left_mouse_button_press(const QPoint& pos) {
 
 void ChartView::on_left_mouse_button_release() {
   if(m_draw_state == DrawState::LINE || m_draw_state == DrawState::POINT) {
+    m_line_mouse_offset.reset();
```
kranar: Prefer ```m_line_mouse_offset = std::nullopt;```

spireken: What's the difference? I also just read that I can assign it an empty braced initializer list.

kranar: Only difference is coding style. For whatever reason the C++ community has converged to this convention.

PR #364, opened 2019-08-20
Snapshot limit

spireken:


Applications/Spire/Source/Charting/CachedChartModel.cpp
```diff
@@ -17,159 +20,127 @@ ChartValue::Type CachedChartModel::get_y_axis_type() const {
...
+          static_cast<int>(result.Get().size()) == info.m_limit.GetSize()) {
+        return QtPromise<std::vector<Candlestick>>(
+          [data = std::move(result.Get())] () mutable {
+            return std::move(data);
+          });
+      }
+      auto load_first = info.m_requested_first;
+      auto load_last = info.m_requested_last;
+      if(info.m_limit.GetType() == SnapshotLimit::Type::HEAD) {
+        for(auto range : m_ranges) {
```
kranar: ```auto& range : m_ranges```

Applications/Spire/Source/Charting/CachedChartModel.cpp
```diff
@@ -17,159 +20,127 @@ ChartValue::Type CachedChartModel::get_y_axis_type() const {
...
-      remove_range(ranges, range);
     }
   }
-  auto index = std::lower_bound(ranges.begin(), ranges.end(), first,
-    [] (const auto& index, const auto& value) {
-      return index.m_start < value;
-    });
-  ranges.insert(index, ChartRange{new_first, new_last});
-  m_ranges = ranges;
+  m_cache.store(std::move(
```
kranar: I claim you don't need a ```move``` here, why am I right or wrong?

spireken: You're correct because the vector is already an rvalue.

kranar: Good good!

PR #365, opened 2019-10-21
Min max size window

spireken:
I made a temporary change in the TimeAndSalesUiTester to demo this. I kept it consistent with the Qt functions, so the usage looks like:

setMinimumSize(scale(180, 200));
setMaximumSize(scale(250, 600));
resize_body(scale(180, 450)); // set initial window size
set_maximizeable(false);

### DIFF
--- a/Applications/Spire/Include/Spire/Ui/Window.hpp
+++ b/Applications/Spire/Include/Spire/Ui/Window.hpp
@@ -62,6 +62,8 @@ namespace Spire {
       */
       void set_fixed_body_size(const QSize& size);
 
+      void set_maximizeable(bool is_maximizeable);
+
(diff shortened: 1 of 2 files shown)


Applications/Spire/Include/Spire/Ui/Window.hpp
```diff
@@ -62,6 +62,13 @@ namespace Spire {
       */
       void set_fixed_body_size(const QSize& size);
 
+      //! Sets whether the window can be maximized.
+      /*
+        \param is_maximizeable True if the window can be maximized, false
+                               otherwise.
+      */
+      void set_maximizeable(bool is_maximizeable);
```
kranar: Is there a way to do this without a new method?

PR #375, opened 2019-11-15
Updated local imbalance model to remove disconnected signals when imb…

spireken:
…alance is inserted and prevent reentrancyq, added tests for new functionality

### DIFF
--- a/Applications/Spire/Source/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorModel.cpp
+++ b/Applications/Spire/Source/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorModel.cpp
@@ -25,11 +25,8 @@ OrderImbalanceIndicatorModel::SubscriptionResult
 void LocalOrderImbalanceIndicatorModel::insert(
     const OrderImbalance& imbalance) {
   m_imbalances.push_back(imbalance);
-  for(auto i = std::size_t(0); i < m_signals.size(); ++i) {
-    if(m_signals[i].m_start_time <= imbalance.m_timestamp &&
-        imbalance.m_timestamp <= m_signals[i].m_end_time) {
-      m_signals[i].m_imbalance_signal(imbalance);
-    }
+  if(!m_signals.empty()) {
+    update_signals(0, m_signals.size(), m_signals.size() - 1, imbalance);
   }
 }
 
@@ -37,3 +34,24 @@ LocalOrderImbalanceIndicatorModel::Subscription::Subscription(
   const boost::posix_time::ptime& start,
   const boost::posix_time::ptime& end)
   : m_start_time(start), m_end_time(end) {}
+
+void LocalOrderImbalanceIndicatorModel::update_signals(
+    std::size_t current_index, std::size_t swap_index, std::size_t end_index,
+    const OrderImbalance& imbalance) {
+  if(m_signals[current_index].m_start_time <= imbalance.m_timestamp &&
+      imbalance.m_timestamp <= m_signals[current_index].m_end_time) {
+    m_signals[current_index].m_imbalance_signal(imbalance);
+  }
+  if(current_index == swap_index || current_index == end_index) {
+    if(swap_index <= end_index ||
+        m_signals[end_index].m_imbalance_signal.num_slots() == 0) {
+      m_signals.erase(m_signals.begin() + current_index,
+        m_signals.begin() + end_index + 1);
+    }
+    return;
+  } else if(m_signals[current_index].m_imbalance_signal.num_slots() == 0) {
+    std::swap(m_signals[current_index], m_signals[--swap_index]);
+    --current_index;
+  }
+  update_signals(++current_index, swap_index, end_index, imbalance);
+}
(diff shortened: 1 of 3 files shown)


Applications/Spire/Source/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorModel.cpp
```diff
@@ -25,15 +25,33 @@ OrderImbalanceIndicatorModel::SubscriptionResult
...
 
 LocalOrderImbalanceIndicatorModel::Subscription::Subscription(
   const boost::posix_time::ptime& start,
   const boost::posix_time::ptime& end)
   : m_start_time(start), m_end_time(end) {}
+
+void LocalOrderImbalanceIndicatorModel::update_signals(
+    std::size_t current_index, std::size_t swap_index, std::size_t end_index,
+    const OrderImbalance& imbalance) {
+  if(m_signals[current_index].m_start_time <= imbalance.m_timestamp &&
```
kranar: Use a for loop instead of a recursive function.
Test for ```num_slots``` before calling the signal, no point in calling a signal that isn't listening to anything.
Find a way to sort ```m_signals``` to avoid having to always loop over ever element of ```m_signals```.

spireken: The issue with sorting is that a signal may be inserted into the range that is begin evaluated in the insert(OrderImbalance&) function.

I can put new signals in a temporary list (waiting) and only insert them when calling the signals is done, but I haven't found a nice way of doing it. My quick implementation would look something like:

```
  // in the subscribe(...) function
  if(is_signaling) {
    m_waiting.push_back(Subscription(...));
  } else {
    m_signals.push_back(Subscription(start, end));
  }
```
```
  // in the insert(...) function
  is_signaling = true;
  m_imbalances.push_back(imbalance);
  auto end = m_signals.size();
  for(auto i = std::size_t(0); i < end; ++i) {
    if(m_signals[i].m_imbalance_signal.num_slots() != 0 &&
        m_signals[i].m_start_time <= imbalance.m_timestamp &&
        imbalance.m_timestamp <= m_signals[i].m_end_time) {
      m_signals[i].m_imbalance_signal(imbalance);
    } else {
      //swap signal to end
    }
  }
  //remove disconnected
  is_signaling = false;
  if(!m_waiting.empty()) {
    m_signals.sorted_insert(waiting);
    m_waiting.clear();
  }
```

kranar:
Good point, do not sort.

Applications/Spire/Source/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorModel.cpp
```diff
@@ -25,33 +25,24 @@ OrderImbalanceIndicatorModel::SubscriptionResult
 void LocalOrderImbalanceIndicatorModel::insert(
     const OrderImbalance& imbalance) {
   m_imbalances.push_back(imbalance);
-  if(!m_signals.empty()) {
-    update_signals(0, m_signals.size(), m_signals.size() - 1, imbalance);
+  auto end_index = m_signals.size();
+  auto swap_index = end_index;
+  for(auto i = std::size_t(0); i < end_index && i != swap_index; ++i) {
```
kranar: When is it possible for ```i >= end_index``` and ```i != swap_index```?

Applications/Spire/Source/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorModel.cpp
```diff
@@ -25,33 +25,24 @@ OrderImbalanceIndicatorModel::SubscriptionResult
...
+    if(m_signals[i].m_imbalance_signal.num_slots() != 0) {
```
kranar: Factor out ```m_signals[i]```.

Applications/Spire/Source/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorModel.cpp
```diff
@@ -25,33 +25,24 @@ OrderImbalanceIndicatorModel::SubscriptionResult
...
+      if(m_signals[i].m_start_time <= imbalance.m_timestamp &&
+          imbalance.m_timestamp <= m_signals[i].m_end_time) {
+        m_signals[i].m_imbalance_signal(imbalance);
+      }
+    } else {
+      std::swap(m_signals[i], m_signals[--swap_index]);
+      --i;
```
kranar: Use a ```while``` loop then. For loops should never modify the iterator.

Applications/Spire/Source/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorModel.cpp
```diff
@@ -25,33 +25,24 @@ OrderImbalanceIndicatorModel::SubscriptionResult
...
+    }
   }
+  m_signals.erase(m_signals.begin() + swap_index,
+    m_signals.begin() + end_index);
```
kranar: When will ```end_index``` ever be different from ```m_signals.size()```?

spireken: When a called signal results in a call to subscribe and a new signal gets inserted.

kranar: Good.

PR #377, opened 2019-11-19
Cached imbalance model

spireken:


Applications/Spire/Include/Spire/OrderImbalanceIndicator/CachedOrderImbalanceIndicatorModel.hpp
```diff
@@ -0,0 +1,61 @@
...
+    public:
+
+      //! Constructs a CachedOrderImbalanceModel with a given source model.
+      /*
+        \param source_model The model supplying data to the cached model.
+      */
+      CachedOrderImbalanceIndicatorModel(
+        std::shared_ptr<OrderImbalanceIndicatorModel> source_model);
+
+      OrderImbalanceIndicatorModel::SubscriptionResult subscribe(
```
kranar: You can remove the ```OrderImbalanceIndicatorModel::```, anything in the ```OrderImbalanceIndicatorModel``` namespace/scope will implicitly be available to the ```CachedOrderImbalanceIndicatorModel```.

Applications/Spire/Source/OrderImbalanceIndicator/CachedOrderImbalanceIndicatorModel.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+OrderImbalanceIndicatorModel::SubscriptionResult
+    CachedOrderImbalanceIndicatorModel::subscribe(
+    const boost::posix_time::ptime& start,
+    const boost::posix_time::ptime& end,
+    const OrderImbalanceSignal::slot_type& slot) {
+  if(contains(m_ranges, continuous_interval<ptime>::closed(start, end))) {
+    return get_subscription(start, end, slot);
+  }
+  return load_imbalances(start, end, slot);
```
kranar: Look into only loading the portion that isn't cached.

Applications/Spire/Source/OrderImbalanceIndicator/CachedOrderImbalanceIndicatorModel.cpp
```diff
@@ -0,0 +1,108 @@
...
+    const OrderImbalanceSignal::slot_type& slot) {
+  m_subscriptions.push_back(Subscription(start, end));
+  auto imbalances = std::vector<OrderImbalance>();
+  std::copy_if(m_imbalances.begin(), m_imbalances.end(),
+    std::back_inserter(imbalances), [=] (auto& imbalance) {
+      return start <= imbalance.m_timestamp && imbalance.m_timestamp <= end;
+    });
+  return {m_subscriptions.back().m_imbalance_signal.connect(slot),
+    QtPromise([requested_imbalances = std::move(imbalances)] {
+      return std::move(requested_imbalances);
```
kranar: Mark the lambda as ```mutable``` so that the ```std::move``` actually moves instead of copies.

Applications/Spire/Source/OrderImbalanceIndicator/CachedOrderImbalanceIndicatorModel.cpp
spireken: The loaded portions are calculated on line 47-48, where I take the difference of the requested range and the previously loaded ranges.

Applications/Spire/Source/OrderImbalanceIndicator/CachedOrderImbalanceIndicatorModel.cpp
```diff
@@ -0,0 +1,108 @@
...
+        on_order_imbalance(imbalance);
+      });
+    m_connections.push_back(std::move(connection));
+    promises.push_back(std::move(promise));
+  }
+  m_subscriptions.push_back(Subscription(start, end));
+  return {m_subscriptions.back().m_imbalance_signal.connect(slot),
+    all(std::move(promises)).then([=] (auto& imbalances_lists) {
+      m_ranges.add({start, end});
+      auto loaded_imbalances_lists = imbalances_lists.Get();
```
kranar: Line 61 makes a copy of a ```vector```, instead of making a copy iterate over the expression directly:

    for(auto& list : imbalances_lists.Get())

Applications/Spire/Include/Spire/OrderImbalanceIndicator/CachedOrderImbalanceIndicatorModel.hpp
```diff
@@ -0,0 +1,61 @@
...
+        OrderImbalanceSignal m_imbalance_signal;
+        boost::posix_time::ptime m_start_time;
+        boost::posix_time::ptime m_end_time;
+
+        Subscription(const boost::posix_time::ptime& start,
+          const boost::posix_time::ptime& end);
+      };
+
+      std::shared_ptr<OrderImbalanceIndicatorModel> m_source_model;
+      std::unordered_set<Nexus::OrderImbalance, OrderImbalanceHash>
```
kranar: Can you use a ```vector``` sorted by timestamp instead of an ```unordered_set```? Looks like every subscription involves iterating over this entire set when it could be possible to calculate the index of the first imbalance, last imbalance, and return everything inbetween.

Applications/Spire/Include/Spire/OrderImbalanceIndicator/CachedOrderImbalanceIndicatorModel.hpp
spireken: I get an error about a missing type specifier when I leave it out.

Applications/Spire/Include/Spire/OrderImbalanceIndicator/CachedOrderImbalanceIndicatorModel.hpp
spireken: I was trying to avoid having to write the code to merge the results because it involves finding the parts of the data that overlap and removing them.

Would a set work? I can't remember why I settled on an unordered_set, but if a set works it would provide the best combination of both; simpler merging and logarthmic searching.

Applications/Spire/Include/Spire/OrderImbalanceIndicator/CachedOrderImbalanceIndicatorModel.hpp
kranar: Determine the cause of the compiler error and resolve it. Just like you don't write out ```OrderImbalanceIndicatorModel::OrderImbalanceSignal::slot_type``` you also don't write out ```OrderImbalanceIndicatorModel::SubscriptionResult```.

Applications/Spire/Include/Spire/OrderImbalanceIndicator/CachedOrderImbalanceIndicatorModel.hpp
kranar: Yes you will need to remove duplicates. An ```std::set``` won't work unfortunately.

spireken: What will be the secondary sort criteria if two imbalances share the same timestamps?

kranar: You don't need a secondary sort criteria luckily.

Applications/Spire/Include/Spire/OrderImbalanceIndicator/CachedOrderImbalanceIndicatorModel.hpp
spireken: Do you mean this for just the header file?

Applications/Spire/Include/Spire/OrderImbalanceIndicator/CachedOrderImbalanceIndicatorModel.hpp
```diff
@@ -0,0 +1,56 @@
...
+          const boost::posix_time::ptime& end);
+      };
+
+      std::shared_ptr<OrderImbalanceIndicatorModel> m_source_model;
+      std::vector<Nexus::OrderImbalance> m_imbalances;
+      std::vector<Subscription> m_subscriptions;
+      boost::icl::interval_set<boost::posix_time::ptime> m_ranges;
+      std::vector<boost::signals2::scoped_connection> m_connections;
+
+      OrderImbalanceIndicatorModel::SubscriptionResult
```
kranar: Remove ```OrderImbalanceIndicatorModel::```.

Applications/Spire/Include/Spire/OrderImbalanceIndicator/CachedOrderImbalanceIndicatorModel.hpp
```diff
@@ -0,0 +1,56 @@
...
+        get_subscription(const boost::posix_time::ptime& start,
```
kranar: If this is a "getter" then it needs to be ```const``` since as the name suggests, a getter does not mutate anything. If this does need to modify/write data, then it needs to be renamed, for example from ```get_subscription``` to ```make_subscription```.

Applications/Spire/Include/Spire/SpireTester/TestOrderImbalanceIndicatorModel.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+  //! Represents an OrderImbalanceIndicatorModel for testing, allowing for the
+  //! observation of subscription requests sent to the model and simulation
+  //! of asynchronous loads.
+  class TestOrderImbalanceIndicatorModel :
+      public OrderImbalanceIndicatorModel {
+    public:
+
+      //! Stores a request to subscribe to an OrderImbalanceModel.
+      class SubscribeEntry {
```
kranar: Generally classes and "data structures" are named after nouns since they represent objects. Functions are named as verbs since they perform an action. If you wish to use "subscribe" in its noun form to describe the "entry" then you need to rename this to ```SubscriptionEntry``` since subscription is a noun whereas subscribe is a verb.

You named the function ```subscribe``` which is correct, but you incorrectly named some other functions like ```pop_subscribe``` which should be ```pop_subscription``` since "pop" is the verb and "subscription" is the noun. Go over this class and revise the names to use the noun or verb depending on the situation.

Applications/Spire/Source/OrderImbalanceIndicator/CachedOrderImbalanceIndicatorModel.cpp
```diff
@@ -0,0 +1,121 @@
...
+}
+
+OrderImbalanceIndicatorModel::SubscriptionResult
+    CachedOrderImbalanceIndicatorModel::load_imbalances(
+    const boost::posix_time::ptime& start,
+    const boost::posix_time::ptime& end,
+    const OrderImbalanceSignal::slot_type& slot) {
+  auto promises = std::vector<QtPromise<std::vector<OrderImbalance>>>();
+  auto ranges = interval_set<ptime>(
+    {continuous_interval<ptime>::closed(start, end)}) - m_ranges;
```
kranar: What do the curly brackets do in ```{continuous_interval...}```? What happens if you remove the curly brackets?

spireken: I didn't realize there was a constructor for single values.

PR #381, opened 2019-12-02
Local imbalance model

spireken:


Applications/Spire/Include/Spire/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorModel.hpp
```diff
@@ -29,6 +29,11 @@ namespace Spire {
 
       std::shared_ptr<OrderImbalanceChartModel> get_chart_model(
         const Nexus::Security& security) override;
+
+    private:
+      std::vector<Nexus::OrderImbalance> m_imbalances;
+      OrderImbalanceSignal m_imbalance_published_signal;
+      boost::optional<Nexus::OrderImbalance> m_last_published_imbalance;
```
kranar: What is the purpose of this member variable?

Applications/Spire/Source/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorModel.cpp
```diff
@@ -1 +1,43 @@
...
+
+void LocalOrderImbalanceIndicatorModel::insert(
+    const OrderImbalance& imbalance) {
+  m_imbalances.push_back(std::move(imbalance));
+}
+
+QtPromise<std::vector<Nexus::OrderImbalance>>
+    LocalOrderImbalanceIndicatorModel::load(const TimeInterval& interval) {
+  auto imbalances = std::vector<OrderImbalance>();
+  for(auto& imbalance : m_imbalances) {
```
kranar: Optimize this.

Applications/Spire/Include/Spire/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorModel.hpp
spireken: The last published imbalance won't necessarily be the last imbalance in the vector of stored imbalances.

Applications/Spire/Source/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorModel.cpp
spireken: Should I sort the stored imbalances?

Applications/Spire/Source/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorModel.cpp
```diff
@@ -1,30 +1,37 @@
...
 }
 
 QtPromise<std::vector<Nexus::OrderImbalance>>
     LocalOrderImbalanceIndicatorModel::load(const TimeInterval& interval) {
-  auto imbalances = std::vector<OrderImbalance>();
-  for(auto& imbalance : m_imbalances) {
-    if(boost::icl::contains(interval, imbalance.m_timestamp)) {
-      imbalances.push_back(imbalance);
-    }
+  auto first = std::find_if(m_imbalances.begin(), m_imbalances.end(),
```
kranar: Too slow.

Applications/Spire/Source/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorModel.cpp
```diff
@@ -41,3 +48,12 @@ std::shared_ptr<OrderImbalanceChartModel>
     const Security& security) {
   throw std::runtime_error("method not implemented");
 }
+
+void LocalOrderImbalanceIndicatorModel::insert_sorted(
+    const OrderImbalance& imbalance) {
+  auto index = std::find_if(m_imbalances.begin(), m_imbalances.end(),
```
kranar: Too slow.

Applications/Spire/Include/Spire/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorModel.hpp
```diff
@@ -29,6 +29,13 @@ namespace Spire {
...
```
kranar: Remove ```m_last_published_imbalance```.

Applications/Spire/Source/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorModel.cpp
```diff
@@ -1 +1,57 @@
...
+    LocalOrderImbalanceIndicatorModel::load(const TimeInterval& interval) {
+  auto first = std::lower_bound(m_imbalances.begin(), m_imbalances.end(),
+    interval.lower(), [] (const auto& imbalance, const auto& timestamp) {
+      return imbalance.m_timestamp < timestamp;
+    });
+  auto last = std::upper_bound(m_imbalances.begin(), m_imbalances.end(),
+    interval.upper(), [] (const auto& timestamp, const auto& imbalance) {
+      return imbalance.m_timestamp > timestamp;
+    });
+  return QtPromise(
```
kranar: What if ```interval``` is an open interval?

Applications/Spire/Source/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorModel.cpp
```diff
@@ -1 +1,57 @@
...
+    [imbalances = std::vector<OrderImbalance>(first, last)] () mutable {
+      return std::move(imbalances);
+  });
+}
+
+SubscriptionResult<boost::optional<Nexus::OrderImbalance>>
+    LocalOrderImbalanceIndicatorModel::subscribe(
+    const OrderImbalanceSignal::slot_type& slot) {
+  return {m_imbalance_published_signal.connect(slot),
+    QtPromise([imbalance = m_last_published_imbalance] () mutable {
```
kranar: Use m_imbalances.back() as the last published imbalance. If it doesn't exist then use ```boost::none```.

spireken: How can we make that assumption, though? If I insert A and C, then publish B, the subscribe method will return C as the last published imbalance. Or is the last published imbalance strictly the newest imbalance?

kranar: Good point, the intention is that the ```subscribe``` method returns the most recent imbalance. ```insert``` is used to add historical imbalances and ```publish``` is used to add the newest imbalance.

Applications/Spire/Source/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorModel.cpp
spireken: Then the implementation probably gets pretty ugly. Is it possible to create a ClosedInterval type?

kranar: ```interval``` can be open or closed or half open/closed.

Applications/Spire/Source/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorModel.cpp
```diff
@@ -1 +1,78 @@
...
+}
+
+void LocalOrderImbalanceIndicatorModel::insert(
+    const OrderImbalance& imbalance) {
+  insert_sorted(imbalance);
+}
+
+QtPromise<std::vector<Nexus::OrderImbalance>>
+    LocalOrderImbalanceIndicatorModel::load(const TimeInterval& interval) {
+  auto first = std::vector<OrderImbalance>::iterator();
```
kranar: Initialize your variables at the point of declaration.

    auto first = [&] {
      if(boost::icl::is_left_closed(interval.bounds())) {
        return std::lower_bound(m_imbalances.begin(), m_imbalances.end(),
          interval.lower(), [] (const auto& imbalance, const auto& timestamp) {
            return imbalance.m_timestamp < timestamp;
          });
      } else {
        return std::upper_bound(m_imbalances.begin(), m_imbalances.end(),
          interval.lower(), [] (const auto& timestamp, const auto& imbalance) {
            return imbalance.m_timestamp > timestamp;
        });
      }
    }();

Applications/Spire/Source/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorModel.cpp
```diff
@@ -1 +1,78 @@
...
+  auto last = std::vector<OrderImbalance>::iterator();
+  if(boost::icl::is_left_closed(interval.bounds())) {
```
kranar: Add a ```using namespace boost;``` and ```using namespace boost::icl;```. You can/should do that with every namespace other than ```std```.

Applications/Spire/Source/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorModel.cpp
```diff
@@ -1 +1,78 @@
...
+        return imbalance.m_timestamp < timestamp;
+      });
+  } else {
+    first = std::upper_bound(m_imbalances.begin(), m_imbalances.end(),
+      interval.lower(), [] (const auto& timestamp, const auto& imbalance) {
+        return imbalance.m_timestamp > timestamp;
+      });
+  }
+  if(boost::icl::is_right_closed(interval.bounds())) {
+    last = std::upper_bound(m_imbalances.begin(), m_imbalances.end(),
```
kranar: Optimize this further by making use of the property that ```last >= first```.

Applications/Spire/Source/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorModel.cpp
```diff
@@ -1 +1,78 @@
...
+
+std::shared_ptr<OrderImbalanceChartModel>
+    LocalOrderImbalanceIndicatorModel::get_chart_model(
+    const Security& security) {
+  throw std::runtime_error("method not implemented");
+}
+
+void LocalOrderImbalanceIndicatorModel::insert_sorted(
+    const OrderImbalance& imbalance) {
+  auto index = std::lower_bound(m_imbalances.begin(), m_imbalances.end(),
```
kranar: Make use of the heuristic that almost all inserts will be at the end of the imbalance list.

Applications/Spire/Source/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorModel.cpp
spireken: Should I leave it on the boost::optional because we sometimes mixed std and boost optionals?

kranar: No because we never do ```using namespace std;``` so there is never a conflict.

PR #383, opened 2019-12-04
Filtered imbalance model

spireken:


Applications/Spire/Source/OrderImbalanceIndicator/FilteredOrderImbalanceIndicatorModel.cpp
```diff
@@ -0,0 +1,114 @@
...
+FilteredOrderImbalanceIndicatorModel::FilteredOrderImbalanceIndicatorModel(
+  std::shared_ptr<OrderImbalanceIndicatorModel> source_model,
+  std::vector<Filter> filters)
+  : m_source_model(std::move(source_model)),
+    m_filters(std::move(filters)) {}
+
+QtPromise<std::vector<Nexus::OrderImbalance>>
+    FilteredOrderImbalanceIndicatorModel::load(
+    const TimeInterval& interval) {
+  return m_source_model->load(interval).then([=] (const auto& imbalances) {
```
kranar: ```FilteredOrderImbalanceIndicatorModel``` needs to have ownership of the promise that the source model returns, otherwise you risk a dangling reference.

Applications/Spire/Source/OrderImbalanceIndicator/FilteredOrderImbalanceIndicatorModel.cpp
```diff
@@ -0,0 +1,114 @@
...
+    return filter_imbalances(imbalances.Get());
+  });
+}
+
+SubscriptionResult<boost::optional<Nexus::OrderImbalance>>
+    FilteredOrderImbalanceIndicatorModel::subscribe(
+    const OrderImbalanceSignal::slot_type& slot) {
+  return m_source_model->subscribe([=] (const auto& imbalance) {
```
kranar: Same issue here, risk of a dangling reference.

Applications/Spire/Source/OrderImbalanceIndicator/FilteredOrderImbalanceIndicatorModel.cpp
spireken: Does the caller not take ownership?

kranar: Add a test case as follows:

* Construct a test model T (the reified model used for testing) (as a shared_ptr).
* Construct a filtered model F (as a unique_ptr) and pass as an argument T.
* Call F->load(P) for some arbitrary value P.
* Call F.reset(), destroying the filtered model.
* Call T.pop_load() and assign it to O.
* Call O.set_result(R) for some arbitrary value R.
* Expected result, no crash.

spireken: Do you mean to reset T (the test model), or F?

I have:

```
    auto T = std::make_shared<TestOrderImbalanceIndicatorModel>();
    auto F = std::make_unique<FilteredOrderImbalanceIndicatorModel>(T,
      std::vector<FilteredOrderImbalanceIndicatorModel::Filter>());
    F->load(TimeInterval::closed(from_time_t(0), from_time_t(1000)));
    F.reset();
    auto O = wait(T->pop_load());
    O->set_result({});
```

And it's working as expected, in that even while the filter model, F, is destroyed, the promise still completes.

spireken: Okay, I see the problem.

kranar: Also use more descriptive names than the ones I give for your actual unit test.

spireken: Is it possible to add a method to see if a QtPromise has called its continuation function?

If I understand it, I need to keep a handle to the QtPromise returned by the source model, but if load(...) is called 10 times and takes a year to resolve each call, I'll need to store those 10 promises and clean them up when they've resolved. I could store each promise as a struct with the promise and an is_completed bool, but I know you prefer a 'single source of truth'.

kranar: No, you will have a map of ```QtPromise``` objects, you add them into that map and then in the callback you remove it from the map.

    auto id = ++m_next_promise_id;
    auto some_promise = some_load(
      [=] {
        m_pending_promises.erase(id);
      });
    m_pending_promises.insert(std::make_pair(id, std::move(some_promise)));

Applications/Spire/Source/OrderImbalanceIndicator/FilteredOrderImbalanceIndicatorModel.cpp
spireken: In the first example, the call to load() creates a promise, and the call to then() creates another. The issue being the original load() promise will go out of scope if it's not stored.

But how does the dangling reference issue apply here if the sole promise is being returned?

kranar: Add a test case as follows:

* Construct a local model L (as a shared_ptr).
* Construct a filtered model F (as a unique_ptr) and pass as an argument L.
* Call F->subscribe(P) for some arbitrary value P.
* Call F.reset(), destroying the filtered model.
* Call L.publish(M).
* Expected result, no crash.

spireken: So is the expectation that it won't crash in its current form? Or that the correct implementation won't crash?

It's passing in its current form but is it possible that reset isn't actually deleting anything so the filtered model still exists in memory?

kranar: Push the test.

spireken: It's the last test in the filtered imbalance model tester file. I kept the placeholder names to make reference the PR notes easier.

I temporary copied the test model into the tester source because it's in the other cache branch, but if I need it I'll move it into it's own files.

kranar: Are you running these tests in debug mode or release mode?

spireken: Debug.

kranar: I can't test this right now but if you go over the test it should be clear that the application should fail.

spireken: I think my original implementation may be partially correct. Following the call stack for the source (test) model's load method, it calls then() (link below), which constructs another promise from the original promise; so storing a handle to the original promise may not be required.

[QtPromise Link](https://github.com/eidolonsystems/nexus/blob/d3c5a8374c344c9a598cc575d7875b0000d40405/Applications/Spire/Include/Spire/Spire/QtPromise.hpp#L174)

This doesn't fix the issue with the filter model being destroyed, though.

spireken: I had to add a wait() call for the filtered model's load() promise. I suspect the test was ending, then the test model would be destroyed, so the test model's load method didn't return a value to the filtered model for the crash to happen.

Applications/Spire/Source/OrderImbalanceIndicator/FilteredOrderImbalanceIndicatorModel.cpp
```diff
@@ -0,0 +1,116 @@
+#include "Spire/OrderImbalanceIndicator/FilteredOrderImbalanceIndicatorModel.hpp"
+
+using Filter = Spire::FilteredOrderImbalanceIndicatorModel::Filter;
+using namespace Nexus;
+using namespace Spire;
+
+FilteredOrderImbalanceIndicatorModel::FilteredOrderImbalanceIndicatorModel(
+  std::shared_ptr<OrderImbalanceIndicatorModel> source_model,
+  std::shared_ptr<std::vector<Filter>> filters)
```
kranar: ```filters``` can not be a ```shared_ptr```.

Applications/Spire/Source/OrderImbalanceIndicator/FilteredOrderImbalanceIndicatorModel.cpp
```diff
@@ -0,0 +1,116 @@
...
+  : m_source_model(std::move(source_model)),
+    m_filters(std::move(filters)) {}
+
+QtPromise<std::vector<Nexus::OrderImbalance>>
+    FilteredOrderImbalanceIndicatorModel::load(
+    const TimeInterval& interval) {
+  return m_source_model->load(interval).then(
+    [&filters = *m_filters] (const auto& imbalances) {
```
kranar: You need to make a copy of ```m_filters```.

spireken: I thought we were trying to avoid copying them? Or if they're supposed to be a shared_ptr, does this not contradict the first change request?

kranar: You want to avoid copying the ```vector```, but ```m_filter``` is not a ```vector```, it's a ```shared_ptr``` so copying it is cheap.

Applications/Spire/Source/OrderImbalanceIndicator/FilteredOrderImbalanceIndicatorModel.cpp
spireken: So how does the model store them?

kranar: Use a ```shared_ptr``` to store them, but not as a parameter.

spireken: How would that work? Wouldn't that mean creating a shared_ptr from either a stack-allocated variable or a raw pointer?

kranar: Yep that's right.

spireken: Doesn't this now create the opposite problem, where the lifetime of the whatever is holding the filters would have to be longer than the filter model? Is the assumption that the filter models would be used in a limited scope?

spireken: I think the filter model should just take and store the shared_ptr. If the filter model outlives whatever has a handle to the model, then the filters will still be valid in this case.

kranar: FilteredOrderImbalanceIndicatorModel::FilteredOrderImbalanceIndicatorModel(
      std::shared_ptr<OrderImbalanceIndicatorModel> source_model,
      std::vector<Filter> filters)
      : m_filters(std::make_shared<std::vector<Filter>>(std::move(filters)))

PR #385, opened 2019-12-11
Imbalance chart model

spireken:
I'd like to merge the local model changes in this branch first, because I'm not sure how much work the cache version is going to take.


Applications/Spire/Source/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorModel.cpp
```diff
@@ -19,36 +20,28 @@ void LocalOrderImbalanceIndicatorModel::insert(
...
       return std::move(imbalances);
   });
 }
 
+QtPromise<std::vector<Nexus::OrderImbalance>>
+    LocalOrderImbalanceIndicatorModel::load(const Security& security,
+    const TimeInterval& interval) {
+  auto imbalances = std::vector<OrderImbalance>();
+  auto [first, last] = get_iterators_from_interval(interval);
+  for(auto i = first; i < last; ++i) {
```
kranar: Use ```i != last``` instead of ```i < last```.

PR #392, opened 2020-01-21
Range input model

spireken:


Applications/Spire/Include/Spire/Spire/LocalRangeInputModel.hpp
```diff
@@ -0,0 +1,31 @@
...
+  //! Implements a RangeInputModel in memory.
+  class LocalRangeInputModel : public RangeInputModel {
+    public:
+
+      //! Constructs a LocalRangeInputModel from provided data, with the
+      //! minimum and maximum values inferred from the data.
+      /*
+        \param data The data for the model.
+      */
+      LocalRangeInputModel(const std::vector<ChartValue> data);
```
kranar: Pass by value, remove the ```const```.

Applications/Spire/Include/Spire/Spire/RangeInputModel.hpp
```diff
@@ -0,0 +1,36 @@
...
+      
+        //! A histogram where each element is a bin, and the value
+        //! of the element is the bin's frequency.
+        std::vector<int> m_histogram;
+
+        //! The highest frequency count amoung all the histogram's bins.
+        int m_highest_frequency;
+      };
+
+      //! Returns the smallest value among the data.
```
kranar: Needs a virtual destructor.

Applications/Spire/Source/Common/LocalRangeInputModel.cpp
```diff
@@ -0,0 +1,46 @@
+#include "Spire/Spire/LocalRangeInputModel.hpp"
+#include "Spire/Spire/Utility.hpp"
+
+using namespace Nexus;
+using namespace Spire;
+
+LocalRangeInputModel::LocalRangeInputModel(
+    const std::vector<ChartValue> data)
+    : m_data(data),
```
kranar: Use a move.

Applications/Spire/Source/Common/LocalRangeInputModel.cpp
```diff
@@ -0,0 +1,46 @@
...
+      m_minimum_value(0),
+      m_maximum_value(std::numeric_limits<int>::max()) {
```
kranar: These initializations are incorrect.

Applications/Spire/Include/Spire/Spire/RangeInputModel.hpp
spireken: Even if the class doesn't have dynamically allocated members?

kranar: Good question, but yes, even without any dynamically allocated members. Take some time to read about why it's important.

spireken: From Effective C++: "The C++ language standard is unusually clear on this topic. When you try to delete a derived class object through a base class pointer and the base class has a non-virtual destructor (as EnemyTarget does), the results are undefined."

I found a lot of explanations, but _if correct_, this short and sweet explanation clarifies it.

Applications/Spire/Source/Common/LocalRangeInputModel.cpp
spireken: What should the defaults be if there is no data?

kranar: If there is no data then the min should equal the max. Also avoid situations where you initialize data and then overwrite it, for example:

    auto x = 0;
    if(some_condition) {
      x = 10;
    }

Instead prefer:

    if(some_condition) {
      x = 10;
    } else {
      x = 0;
    }

In other words, remove the initialization from 10 and 11 and initialize those in the body of the constructor.

PR #394, opened 2020-01-30
Single range widget

spireken:
I put a Scratch demo up in drafts. Nothing fancy, just when one button is clicked it sets the value of the other through a ScalarWidget.

### DIFF
--- a/Applications/Spire/Include/Spire/Ui/ScalarWidget.hpp
+++ b/Applications/Spire/Include/Spire/Ui/ScalarWidget.hpp
@@ -22,12 +22,12 @@ namespace Spire {
         \param setter Method to call to set the value that the widget displays.
       */
       template<typename W, typename T>
-      ScalarWidget(W* widget,
-        boost::signals2::connection (W::* connector)(Signal<T>),
-        std::function<void (const T&)> setter);
+      ScalarWidget(W* widget, boost::signals2::connection (W::* connector)(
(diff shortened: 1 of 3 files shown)


Applications/Spire/Include/Spire/Ui/ScalarWidget.hpp
```diff
@@ -0,0 +1,60 @@
+#ifndef SPIRE_SCALAR_WIDGET_HPP
+#define SPIRE_SCALAR_WIDGET_HPP
+#include <QWidget>
+#include "Spire/Spire/Scalar.hpp"
+
+namespace Spire {
+
+  //! Provides a common widget with an interface for getting and setting a
```
kranar: //! Wraps a QWidget used to edit a scalar type T into a QWidget that edits a Scalar.

Applications/Spire/Source/Ui/ScalarWidget.cpp
```diff
@@ -0,0 +1,26 @@
...
+#include <QHBoxLayout>
+
+using namespace boost::signals2;
+using namespace Spire;
+
+void ScalarWidget::set_value(Scalar value) {
+  m_setter(value);
+}
+
+void ScalarWidget::set_layout(QWidget* widget) {
```
kranar: Implement these methods in the order of their declaration.

PR #398, opened 2020-02-10
Time input widget

spireken:
The demo is in drafts. The label is changed when the time change signal is triggered, and the buttons will set the current time of the widget.


Applications/Spire/Include/Spire/Ui/TimeInputWidget.hpp
```diff
@@ -0,0 +1,83 @@
...
+#include "Spire/Ui/Ui.hpp"
+
+namespace Spire {
+
+  //! Displays a time input widget.
+  class TimeInputWidget : public QWidget {
+    public:
+
+      //! Signals that the time has been modified.
+      using TimeChangeSignal = Signal<void (Scalar)>;
```
kranar: We do not have a time input, we have either a time duration which represents an elapsed period of time, a datetime which represents a single point in time, or a timeofday which represents a time of day.

Furthermore none of these widgets should make use of a Scalar, it defeats the entire point of having the widget that transforms to/from Scalars.

spireken: Will the DateInputWidget be separate from the DateTimeInputWidget?

kranar:
Yes it will.

PR #399, opened 2020-02-10
Smooth chart scrolling

andrei-sa:
Includes all the features implemented previously, such as axis grid, drawing mode, etc.
Axis grid behavior is not final as the specification does not clearly define it.
I will also go through the specifications once again to check for any discrepancies. This PR though fixes the scrolling assuming the original implementation otherwise matched the spec.


kranar:
Merge into spire2018, not into master.

PR #401, opened 2020-02-12
Smooth chart scrolling

andrei-sa:
spec.


kranar:
I am closing this pull request. This task is quite challenging and unfortunately I am not satisfied with this implementation. It's not that it's bad but the implementation and API doesn't quite fit with how this widget will be used in the future. Nevertheless there are good ideas in this implementation that should be used in a future implementation so we'll keep this branch on Github.

PR #403, opened 2020-02-19
Duration widget

spireken:


spireken:
The demo doesn't format the label, it's just to show that the value being signaled is correct.

Applications/Spire/Include/Spire/Ui/DurationInputWidget.hpp
```diff
@@ -0,0 +1,61 @@
...
+      //! Signals that the time duration was modified.
+      using DurationSignal = Signal<void (boost::posix_time::time_duration)>;
+
+      //! Constructs a DurationInputWidget.
+      /*
+        \param parent The parent widget.
+      */
+      explicit DurationInputWidget(QWidget* parent = nullptr);
+
+      //! Sets the displayed time duration without trigger the duration
```
kranar: Fix the typo/documentation. Also why doesn't this trigger the signal?

spireken: The thought was that if something was setting the duration on the widget then it's dictating the value and hence already knows what the value is.

kranar: Reasonable idea but widgets can be shared so the signal is necessary. In general all setters for a widget should trigger its corresponding signal (if one exists).

PR #407, opened 2020-03-17
Imbalance context menu

spireken:
The imbalance drafts folder has a ContextMenu tester build.

### DIFF
--- a/Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
+++ b/Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
@@ -22,7 +22,7 @@ namespace Spire {
       /*
         \param parent The parent widget.
       */
-      OrderImbalanceIndicatorContextMenu(QWidget* parent = nullptr);
+      explicit OrderImbalanceIndicatorContextMenu(QWidget* parent = nullptr);
 
       //! Connects a slot to the export table signal.
       boost::signals2::connection connect_export_table_signal(
@@ -80,7 +80,8 @@ namespace Spire {
       mutable ToggledSignal m_notional_value_signal;
       QMenu* m_table_columns_menu;
 
-      void add_check_box(const QString& text, QWidgetAction* action);
+      void add_check_box(const QString& text, ToggledSignal& signal);
+      void add_menu_item(const QString& text, SelectedSignal& signal);
   };
 }
 
--- a/Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
+++ b/Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
@@ -69,7 +69,7 @@ namespace Spire {
       mutable SelectedSignal m_reset_signal;
       mutable ToggledSignal m_side_signal;
       mutable ToggledSignal m_size_signal;
-      mutable ToggledSignal m_ref_px_signal;
+      mutable ToggledSignal m_reference_price_signal;
       mutable ToggledSignal m_date_signal;
       mutable ToggledSignal m_time_signal;
       mutable ToggledSignal m_notional_value_signal;
--- a/Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
+++ b/Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
@@ -36,10 +36,6 @@ namespace Spire {
       boost::signals2::connection connect_reset_signal(
         const SelectedSignal::slot_type& slot) const;
 
-      //! Connects a slot to the security toggle signal.
-      boost::signals2::connection connect_security_toggled_signal(
-        const ToggledSignal::slot_type& slot) const;
-
       //! Connects a slot to the side toggle signal.
       boost::signals2::connection connect_side_toggled_signal(
         const ToggledSignal::slot_type& slot) const;
@@ -71,7 +67,6 @@ namespace Spire {
       mutable SelectedSignal m_export_table_signal;
       mutable SelectedSignal m_export_chart_signal;
       mutable SelectedSignal m_reset_signal;
-      mutable ToggledSignal m_security_signal;
       mutable ToggledSignal m_side_signal;
       mutable ToggledSignal m_size_signal;
       mutable ToggledSignal m_ref_px_signal;
--- a/Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
+++ b/Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
@@ -1,13 +1,42 @@
 #ifndef SPIRE_ORDER_IMBALANCE_INDICATOR_CONTEXT_MENU_HPP
 #define SPIRE_ORDER_IMBALANCE_INDICATOR_CONTEXT_MENU_HPP
 #include <QMenu>
+#include <QWidgetAction>
+#include "Spire/Spire/Spire.hpp"
+#include "Spire/Ui/CheckBox.hpp"
 
 namespace Spire {
 
   class OrderImbalanceIndicatorContextMenu : public QMenu {
     public:
 
+      using SelectedSignal = Signal<void ()>;
+
+      using ToggledSignal = Signal<void (bool)>;
+
       OrderImbalanceIndicatorContextMenu(QWidget* parent = nullptr);
+
+    private:
+      mutable SelectedSignal m_export_table_signal;
+      mutable SelectedSignal m_export_chart_signal;
+      mutable SelectedSignal m_reset_signal;
+      CheckBox* m_security_check_box;
+      CheckBox* m_side_check_box;
+      CheckBox* m_size_check_box;
+      CheckBox* m_ref_px_check_box;
+      CheckBox* m_date_check_box;
+      CheckBox* m_time_check_box;
+      CheckBox* m_notional_value_check_box;
+      mutable ToggledSignal m_security_signal;
+      mutable ToggledSignal m_side_signal;
+      mutable ToggledSignal m_size_signal;
+      mutable ToggledSignal m_ref_px_signal;
+      mutable ToggledSignal m_date_signal;
+      mutable ToggledSignal m_time_signal;
+      mutable ToggledSignal m_notional_value_signal;
+
+      void add_check_box(CheckBox* check_box, const QString& text,
+        QWidgetAction* action);
   };
 }
 
--- a/Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
+++ b/Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
@@ -16,10 +16,14 @@ namespace Spire {
 
       OrderImbalanceIndicatorContextMenu(QWidget* parent = nullptr);
 
+    protected:
+      bool eventFilter(QObject* watched, QEvent* event) override;
+
     private:
       mutable SelectedSignal m_export_table_signal;
       mutable SelectedSignal m_export_chart_signal;
       mutable SelectedSignal m_reset_signal;
+      QMenu* m_table_columns_menu;
       CheckBox* m_security_check_box;
       CheckBox* m_side_check_box;
       CheckBox* m_size_check_box;
--- a/Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
+++ b/Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
@@ -1,0 +1,14 @@
+#ifndef SPIRE_ORDER_IMBALANCE_INDICATOR_CONTEXT_MENU_HPP
+#define SPIRE_ORDER_IMBALANCE_INDICATOR_CONTEXT_MENU_HPP
+#include <QMenu>
+
+namespace Spire {
+
+  class OrderImbalanceIndicatorContextMenu : public QMenu {
+    public:
+
+      OrderImbalanceIndicatorContextMenu(QWidget* parent = nullptr);
+  };
+}
+
+#endif
--- a/Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
+++ b/Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
@@ -10,11 +10,56 @@ namespace Spire {
   class OrderImbalanceIndicatorContextMenu : public QMenu {
     public:
 
+      //! Signals that a menu option was selected.
       using SelectedSignal = Signal<void ()>;
 
-      using ToggledSignal = Signal<void (bool)>;
+      //! Signals that a column check box was toggled.
+      /*
+        \param is_checked True if the check box is in the checked state.
+      */
+      using ToggledSignal = Signal<void (bool is_checked)>;
 
       OrderImbalanceIndicatorContextMenu(QWidget* parent = nullptr);
+
+      //! Connects a slot to the export table signal.
+      boost::signals2::connection connect_export_table_signal(
+        const SelectedSignal::slot_type& slot) const;
+
+      //! Connects a slot to the export chart signal.
+      boost::signals2::connection connect_export_chart_signal(
+        const SelectedSignal::slot_type& slot) const;
+
+      //! Connects a slot to the reset filters signals.
+      boost::signals2::connection connect_reset_signal(
+        const SelectedSignal::slot_type& slot) const;
+
+      //! Connects a slot to the security toggle signal.
+      boost::signals2::connection connect_security_toggled_signal(
+        const ToggledSignal::slot_type& slot) const;
+
+      //! Connects a slot to the side toggle signal.
+      boost::signals2::connection connect_side_toggled_signal(
+        const ToggledSignal::slot_type& slot) const;
+
+      //! Connects a slot to the size toggle signal.
+      boost::signals2::connection connect_size_toggled_signal(
+        const ToggledSignal::slot_type& slot) const;
+
+      //! Connects a slot to the side reference price toggle signal.
+      boost::signals2::connection connect_reference_price_toggled_signal(
+        const ToggledSignal::slot_type& slot) const;
+
+      //! Connects a slot to the date toggle signal.
+      boost::signals2::connection connect_date_toggled_signal(
+        const ToggledSignal::slot_type& slot) const;
+
+      //! Connects a slot to the time toggle signal.
+      boost::signals2::connection connect_time_toggled_signal(
+        const ToggledSignal::slot_type& slot) const;
+
+      //! Connects a slot to the notional value toggle signal.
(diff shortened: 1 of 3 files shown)


Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
```diff
@@ -0,0 +1,87 @@
...
+      /*
+        \param is_checked True if the check box is in the checked state.
+      */
+      using ToggledSignal = Signal<void (bool is_checked)>;
+
+      //! Constructs an OrderImbalanceINdicatorContextMenu.
+      /*
+        \param parent The parent widget.
+      */
+      OrderImbalanceIndicatorContextMenu(QWidget* parent = nullptr);
```
andrei-sa: What will happen in `parent` is `nullptr`? Is it a valid use case?
Regardless, unless you actually need implicit conversion from `QWidget` pointers to `OrderImbalanceIndicatorContextMenu`, make this constructor `explicit`.

Applications/Spire/Source/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.cpp
```diff
@@ -0,0 +1,226 @@
...
+      m_table_columns_menu->move(pos().x() + scale_width(136),
+        pos().y() + scale_height(6));
+    }
+  }
+  return QMenu::eventFilter(watched, event);
+}
+
+void OrderImbalanceIndicatorContextMenu::add_check_box(const QString& text,
+    QWidgetAction* action) {
+  auto check_box = new CheckBox(text, this);
```
andrei-sa: Use `std::unique_ptr`, then call check_box.release() where transferring ownership to `action`.
Also, `check_box` is only used at the end of the function, consider moving its initialization there.

Applications/Spire/Source/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.cpp
```diff
@@ -0,0 +1,226 @@
...
+    .arg(scale_height(12)).arg(scale_height(5))
+    .arg(scale_height(3)).arg(scale_width(8))
+    .arg(scale_height(6)).arg(scale_width(4)));
+  m_table_columns_menu = new QMenu(tr("Table Columns"), this);
+  m_table_columns_menu->installEventFilter(this);
+  m_table_columns_menu->setFixedSize(scale(140, 150));
+  m_table_columns_menu->setContentsMargins(0, scale_height(0), 0,
+    scale_height(4));
+  addMenu(m_table_columns_menu);
+  auto security_action = new QWidgetAction(this);
```
andrei-sa: For all the actions, use `std::unique_ptr` and call `action.get()` where you need a raw pointer handle without ownership transfer and `action.release()` to transfer ownership.
`std::unique_ptr` provides exception safety and incurs no runtime overhead. As a side effect, in the modern C++ world, in makes the intent clearer regarding whether the object is borrowed or acquired.

Applications/Spire/Source/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+using namespace boost::signals2;
+using namespace Spire;
+
+OrderImbalanceIndicatorContextMenu::OrderImbalanceIndicatorContextMenu(
+    QWidget* parent)
+    : QMenu(parent) {
+  setWindowFlag(Qt::NoDropShadowWindowHint);
+  DropShadow context_menu_shadow(true, true, this);
+  setFixedSize(scale(140, 90));
```
andrei-sa: Unless you intend to parameterize these in the nearest future, move constants out to the beginning of the file so that they are easier to locate and modify if necessary.
There is no need to obsessive about that, but most important constants (like window size) probably deserve explicit declaration.

Applications/Spire/Source/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.cpp
```diff
@@ -0,0 +1,226 @@
...
+      });
+  m_table_columns_menu->addAction(date_action);
+  auto time_action = new QWidgetAction(this);
+  add_check_box(tr("Time"), time_action);
+  connect(static_cast<QCheckBox*>(time_action->defaultWidget()),
+    &QCheckBox::stateChanged, [=] (int state) {
+        m_time_signal(state == Qt::Checked);
+      });
+  m_table_columns_menu->addAction(time_action);
+  auto notional_value_action = new QWidgetAction(this);
```
andrei-sa: This action connection routine is very similar for all actions. Check if you can factor it into a function taking the name and signal as arguments.

Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
```diff
@@ -0,0 +1,87 @@
...
+      mutable ToggledSignal m_date_signal;
+      mutable ToggledSignal m_time_signal;
+      mutable ToggledSignal m_notional_value_signal;
+      QMenu* m_table_columns_menu;
+
+      void add_check_box(const QString& text, QWidgetAction* action);
+  };
+}
+
+#endif
```
andrei-sa: Are there newline characters at the end of these files? I am just unsure how Github displays those.

Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
```diff
@@ -0,0 +1,87 @@
...
+      using ToggledSignal = Signal<void (bool is_checked)>;
+
+      //! Constructs an OrderImbalanceINdicatorContextMenu.
+      /*
+        \param parent The parent widget.
+      */
+      OrderImbalanceIndicatorContextMenu(QWidget* parent = nullptr);
+
+      //! Connects a slot to the export table signal.
+      boost::signals2::connection connect_export_table_signal(
```
andrei-sa: Why do these methods have to be `const`?

Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
spireken: There is, Github just removes them. I think Github has some kind of error style for when they're not there.

Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
spireken: I've marked it explicit. The default parameter here is a Qt convention that we're copied for Spire.

Applications/Spire/Source/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.cpp
```diff
@@ -0,0 +1,226 @@
...
+#include <QWidgetAction>
+#include "Spire/Spire/Dimensions.hpp"
+#include "Spire/Ui/CheckBox.hpp"
+#include "Spire/Ui/DropShadow.hpp"
+
+using namespace boost::signals2;
+using namespace Spire;
+
+OrderImbalanceIndicatorContextMenu::OrderImbalanceIndicatorContextMenu(
+    QWidget* parent)
```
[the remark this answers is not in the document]
spireken: For Spire we've been putting the initializer list on a new line after the parameters.

Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
spireken: It's a convention I've been following, I'd ask Kamal if you wanted a full answer.

Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
```diff
@@ -0,0 +1,87 @@
...
+      mutable SelectedSignal m_export_chart_signal;
+      mutable SelectedSignal m_reset_signal;
+      mutable ToggledSignal m_security_signal;
+      mutable ToggledSignal m_side_signal;
+      mutable ToggledSignal m_size_signal;
+      mutable ToggledSignal m_ref_px_signal;
+      mutable ToggledSignal m_date_signal;
+      mutable ToggledSignal m_time_signal;
+      mutable ToggledSignal m_notional_value_signal;
+      QMenu* m_table_columns_menu;
```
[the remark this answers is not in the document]
kranar: Good point in general, unfortunately Qt uses a very old style of C++ programmer and so we have to balance modern C++ coding style with old pre-C++11 coding styles.

Qt has its own object ownership model that it uses for objects that inherit from ```QObject``` so for ```QObject```s we use Qt's ownership model. That model does not make use of ```unique_ptr``` but rather every ```QObject``` has a ```parent``` it will attach itself to, which is why many ```QObject``` constructors take in a ```parent``` parameter that is defaulted to ```nullptr```.

Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
kranar: As a coding convention, connecting to signals is a ```const``` operation. It allows us to pass ```const``` models that can be listened to for changes without being mutated.

Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
```diff
@@ -0,0 +1,87 @@
...
+    public:
+
+      //! Signals that a menu option was selected.
+      using SelectedSignal = Signal<void ()>;
+
+      //! Signals that a column check box was toggled.
+      /*
+        \param is_checked True if the check box is in the checked state.
+      */
+      using ToggledSignal = Signal<void (bool is_checked)>;
```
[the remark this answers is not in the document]
kranar: Yes, naming function parameters helps for documentation purposes.

Applications/Spire/Source/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.cpp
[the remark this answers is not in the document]
kranar: Thanks for pointing this out, as we do code reviews we will likely identify discrepancies between our styles. I will make a note of such discrepancies and add a task to update our coding style so we are all on the same page. For this instance the colon always goes on a new line.

Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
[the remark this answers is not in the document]
andrei-sa: OK, but it does not seem like we cannot use `std::unique_ptr` along with `get()` and `release()` methods to adapt to Qt style where composition is concerned. So it just a convention for Spire?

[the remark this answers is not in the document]
kranar: It ends up being awkward using two orthogonal ownership models for ```QObjects``` that we have to keep track of, and so for ```QObjects``` we use Qt's ownership model and for everything else we use more conventional RAII/smart pointers etc... Unfortunately you'll find many other instances of this, where for Qt there is one set of rules, and for everything else there is another set of rules. It's just an artifact of the fact that Qt is a very old framework that hasn't kept up that well with modern C++.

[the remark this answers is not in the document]
andrei-sa: Also, regardless of the convention, `m_table_columns_menu` is only passed to `addMenu`, which does not take ownership. Where will `m_table_columns_menu` be destroyed?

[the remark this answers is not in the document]
kranar: ```m_table_columns_menu``` is allocated on line 60. As per Qt's object ownership model, it is owned by ```this``` so it will get destroyed when ```this``` gets destroyed. That's why all ```QObject```s have that ```parent``` parameter that is defaulted to ```nullptr```, because ```QObject```'s form an ownership hierarchy. I agree in principle that it's frustrating that Qt works this way given how modern C++ codebases establish object ownership, but this is one of those areas where it's better to use Qt's way of doing things when dealing with ```QObject```s rather than mixing two different ways that end clashing with one another.

Applications/Spire/Source/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.cpp
kranar: On the subject of ownership, the ```this``` being passed to the ```QWidgetAction``` constructor establishes ownership, namely that ```this``` owns ```security_action``` so using a ```unique_ptr``` to also establish ownership as well ends up being redundant and error prone.

Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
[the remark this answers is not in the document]
andrei-sa: Got it. The same applies to all those local `QWidgetAction`'s, right?

Applications/Spire/Source/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.cpp
kranar: This is a good idea.

Applications/Spire/Source/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.cpp
andrei-sa: Seems like it also conflicts with Qt ownership model.

kranar: Moving ```check_box``` to the end is a good idea, the scope of a variable should be minimized as much as possible. As for ownership, the ```CheckBox``` is owned by ```this```, which is fine.

Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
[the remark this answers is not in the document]
kranar: Yes exactly, the ```QWidgetAction```s are all owned by ```this```. All the points you made are good valid points under normal circumstances, as we do more code reviews we will just have to converge onto a uniform coding style for UI code and code involving Qt and then we'll update our coding style on the wiki. Admittedly this will be frustrating for you but I don't want to discourage you from making recommendations for good coding style, on the other hand UI code is intrinsically a lot messier and free form. So the first few code reviews involving UI code will be a work in progress so to speak as we settle on a coding convention for UI applications.

Applications/Spire/Source/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.cpp
spireken: Per Spire convention, I'll add a method in an anonymous namespace to return the size but because it relies on the DPI which is only known at runtime I can't mark it explicitly as constant.

Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.hpp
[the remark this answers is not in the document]
andrei-sa: I just overlooked that `this` magic and made wrong assumptions about what Qt will be capable of doing in terms of memory management. Sorry for that.

Applications/Spire/Source/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.cpp
andrei-sa: Well, the return value of `scale` cannot be constant, but those integer literals most certainly can. You can move those height, width constants out, or you can do what you said if you find it better.

spireken:
Had to remove the security column from the context menu.

andrei-sa:
Looks fine to me.

Applications/Spire/Source/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.cpp
```diff
@@ -0,0 +1,197 @@
...
+  check_box->setFixedSize(CHECK_BOX_SIZE());
+  check_box->setChecked(true);
+  check_box->set_stylesheet(text_style, indicator_style, checked_style,
+    hover_style, focused_style);
+  check_box->setLayoutDirection(Qt::RightToLeft);
+  auto action = new QWidgetAction(this);
+  action->setDefaultWidget(check_box);
+  m_table_columns_menu->addAction(action);
+  connect(static_cast<QCheckBox*>(action->defaultWidget()),
+    &QCheckBox::stateChanged, [&] (int state) {
```
kranar: Never capture by reference if the closure will outlive the scope of its declaration!

Applications/Spire/Source/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.cpp
```diff
@@ -0,0 +1,197 @@
...
+        signal(state == Qt::Checked);
+      });
+}
+
+void OrderImbalanceIndicatorContextMenu::add_menu_item(const QString& text,
+    SelectedSignal& signal) {
+  auto action = new QAction(text, this);
+  connect(action, &QAction::triggered, [&] { signal(); });
```
kranar: Never capture by reference if the closure will outlive its scope.

Applications/Spire/Source/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.cpp
kranar: You will have to do something annoying like:

    [signal = &signal] (int state) {
      (*signal)(...);
    }

What this does is capture a pointer to the existing ```signal``` by value.

Applications/Spire/Source/OrderImbalanceIndicator/OrderImbalanceIndicatorContextMenu.cpp
```diff
@@ -0,0 +1,197 @@
...
+  auto check_box = new CheckBox(text, this);
+  check_box->setFixedSize(CHECK_BOX_SIZE());
+  check_box->setChecked(true);
+  check_box->set_stylesheet(text_style, indicator_style, checked_style,
+    hover_style, focused_style);
+  check_box->setLayoutDirection(Qt::RightToLeft);
+  auto action = new QWidgetAction(this);
+  action->setDefaultWidget(check_box);
+  m_table_columns_menu->addAction(action);
+  connect(static_cast<QCheckBox*>(action->defaultWidget()),
```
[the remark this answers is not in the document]
spireken: I get a compilation error. It looks like Qt tries to verify that the type of the object and the type that has the function pointer are the same.

PR #411, opened 2020-03-20
Key bindings API

andrei-sa:


Applications/Spire/Include/Spire/KeyBindings/KeyBindings.hpp
```diff
@@ -1,8 +1,180 @@
...
+        boost::optional<Nexus::Quantity> m_quantity;
+
+        //! The list of custom tags.
+        std::vector<CustomTag> m_tags;
+      };
+
+      /*! \enum CancelAction
+          \brief Represents an order cancellation action.
+      */
+      enum class CancelAction : int {
```
kranar: Remove the ```int``` unless there is an explicit reason for this to use an ```int```.

Applications/Spire/Include/Spire/KeyBindings/KeyBindings.hpp
```diff
@@ -1,8 +1,180 @@
...
+                boost::none otherwise.
+      */
+      boost::optional<const Action&> find_binding(const Nexus::Region& region,
+        const QKeySequence& sequence) const;
+
+      //! Lists all order action bindings.
+      /*!
+        \return The list of all order action bindings.
+      */
+      OrderBindingsList list_order_bindings() const;
```
kranar: Rename to ```build_order_bindings_list()``` etc...

PR #412, opened 2020-03-20
Drop shadow fix

spireken:
The build is either in the drafts folder or in the drafts folder in a builds folder Darryl created, which I don't have access to. You can right click to show the context menu.

This case includes some other miscellaneous fixes for other style issues.


Applications/Spire/Include/Spire/Ui/CalendarWidget.hpp
```diff
@@ -36,6 +36,9 @@ namespace Spire {
       boost::signals2::connection connect_date_signal(
         const DateSignal::slot_type& slot) const;
 
+    protected:
+      void paintEvent(QPaintEvent* event);
```
kranar: ```override```

Applications/Spire/Include/Spire/Charting/DropDownColorPicker.hpp
```diff
@@ -43,6 +44,7 @@ namespace Spire {
       QColor m_current_color;
       QColor m_stored_color;
       bool m_has_mouse;
+      std::unique_ptr<DropShadow> m_color_picker_drop_shadow;
```
kranar: Remove use of ```unique_ptr```s for ```DropShadow``` in situations where a parent is provided.

Applications/Spire/Source/Ui/DateInputWidget.cpp
```diff
@@ -67,11 +67,7 @@ void DateInputWidget::moveEvent(QMoveEvent* event) {
 }
 
 void DateInputWidget::move_calendar() {
-  auto x_pos = window()->mapToGlobal(
-    geometry().bottomLeft()).x();
-  auto y_pos = window()->mapToGlobal(
-    frameGeometry().bottomLeft()).y();
-  m_calendar_widget->move(x_pos, y_pos + 1);
+  m_calendar_widget->move( window()->mapToGlobal(geometry().bottomLeft()));
```
kranar: Remove space.

Applications/Spire/Include/Spire/Charting/DropDownColorPicker.hpp
spireken: Should I keep passing a parent, or is it redundant?

kranar: ```DropShadow``` makes use of a parent so it's necessary.

PR #413, opened 2020-03-24
Scroll area

spireken:
There's a ScrollArea demo in the drafts folder.

### DIFF
--- a/Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
+++ b/Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
@@ -22,8 +22,8 @@ namespace Spire {
 
       void fade_out_horizontal_scroll_bar();
       void fade_out_vertical_scroll_bar();
-      bool is_within_horizontal_scroll_bar(const QPoint& pos);
-      bool is_within_vertical_scroll_bar(const QPoint& pos);
+      bool is_within_horizontal_scroll_bar(int pos_y);
+      bool is_within_vertical_scroll_bar(int pos_x);
       void set_scroll_bar_style(int handle_size);
   };
 }
--- a/Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
+++ b/Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
@@ -1,0 +1,29 @@
+#ifndef SPIRE_SCROLL_AREA_HPP
+#define SPIRE_SCROLL_AREA_HPP
+#include <QScrollArea>
+#include <QTimer>
+
+namespace Spire {
+
+  class ScrollArea : public QScrollArea {
+    public:
+
+      ScrollArea(QWidget* parent = nullptr);
+
+    protected:
+      void leaveEvent(QEvent* event) override;
+      void mouseMoveEvent(QMouseEvent* event) override;
+
+    private:
+      QTimer m_h_scroll_bar_timer;
+      QTimer m_v_scroll_bar_timer;
+
+      void fade_out_horizontal_scroll_bar();
+      void fade_out_vertical_scroll_bar();
+      bool is_within_horizontal_scroll_bar(const QPoint& pos);
+      bool is_within_vertical_scroll_bar(const QPoint& pos);
+      void set_scroll_bar_style(int handle_size);
+  };
+}
+
+#endif
--- a/Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
+++ b/Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
@@ -32,7 +32,7 @@ namespace Spire {
 
       void hide_horizontal_scroll_bar();
       void hide_vertical_scroll_bar();
-      bool is_within_scroll_bar(QScrollBar* scroll_bar, int pos,
+      bool is_within_opposite_scroll_bar(QScrollBar* scroll_bar, int pos,
         int scroll_size, int widget_size);
       void set_scroll_bar_style(int handle_size);
   };
--- a/Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
+++ b/Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
@@ -14,14 +14,15 @@ namespace Spire {
 
     protected:
       bool eventFilter(QObject* watched, QEvent* event);
+      void leaveEvent(QEvent* event) override;
       void wheelEvent(QWheelEvent* event) override;
 
     private:
       QTimer m_h_scroll_bar_timer;
       QTimer m_v_scroll_bar_timer;
 
-      void fade_out_horizontal_scroll_bar();
-      void fade_out_vertical_scroll_bar();
+      void hide_horizontal_scroll_bar();
+      void hide_vertical_scroll_bar();
       bool is_within_horizontal_scroll_bar(int pos_y);
       bool is_within_vertical_scroll_bar(int pos_x);
       void set_scroll_bar_style(int handle_size);
--- a/Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
+++ b/Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
@@ -5,11 +5,20 @@
 
 namespace Spire {
 
+  //! Displays a scroll area with styled scroll bars.
   class ScrollArea : public QScrollArea {
     public:
 
+      //! Constructs an empty ScrollArea.
+      /*
+        \param parent The parent widget.
+      */
       ScrollArea(QWidget* parent = nullptr);
 
+      //! Sets the widget displayed in the scroll area.
+      /*
+        \param widget The widget to display.
+      */
       void setWidget(QWidget* widget);
 
     protected:
--- a/Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
+++ b/Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
@@ -29,11 +29,15 @@ namespace Spire {
     private:
       QTimer m_horizontal_scroll_bar_timer;
       QTimer m_vertical_scroll_bar_timer;
+      double m_horizontal_scrolling_error;
+      double m_vertical_scrolling_error;
 
       void hide_horizontal_scroll_bar();
       void hide_vertical_scroll_bar();
       bool is_within_opposite_scroll_bar(QScrollBar* scroll_bar, int pos,
-        int scroll_size, int widget_size);
+        int scroll_size, int widget_size) const;
+      void update_scrollbar_position(QScrollBar* scroll_bar, int delta,
+        double& scrolling_error);
       void set_scroll_bar_style(int handle_size);
   };
 }
--- a/Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
+++ b/Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
@@ -22,7 +22,7 @@ namespace Spire {
       void setWidget(QWidget* widget);
 
     protected:
-      bool eventFilter(QObject* watched, QEvent* event);
+      bool eventFilter(QObject* watched, QEvent* event) override;
       void leaveEvent(QEvent* event) override;
       void wheelEvent(QWheelEvent* event) override;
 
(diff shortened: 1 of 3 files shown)


Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+  //! Displays a scroll area with styled scroll bars.
+  class ScrollArea : public QScrollArea {
+    public:
+
+      //! Constructs an empty ScrollArea.
+      /*
+        \param parent The parent widget.
+      */
+      ScrollArea(QWidget* parent = nullptr);
```
andrei-sa: Shouldn't it be `explicit`?

Applications/Spire/Source/Ui/ScrollArea.cpp
```diff
@@ -0,0 +1,165 @@
...
+      setHorizontalScrollBarPolicy(Qt::ScrollBarAsNeeded);
+      setVerticalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
+    } else if(is_within_vertical_scroll_bar(e->pos().x()) &&
+        !horizontalScrollBar()->isSliderDown()) {
+      set_scroll_bar_style(SCROLL_BAR_MAX_SIZE);
+      setVerticalScrollBarPolicy(Qt::ScrollBarAlwaysOn);
+      setHorizontalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
+    } else {
+      set_scroll_bar_style(SCROLL_BAR_MIN_SIZE);
+      if(!m_v_scroll_bar_timer.isActive() &&
```
andrei-sa: You can move `!m_v_scroll_bar_timer.isActive()` to an outer `if` check for both vertical and horizontal routines so that you do not have to check twice.

Applications/Spire/Source/Ui/ScrollArea.cpp
```diff
@@ -0,0 +1,165 @@
...
+        hide_vertical_scroll_bar();
+      }
+      if(!m_h_scroll_bar_timer.isActive() &&
+          horizontalScrollBarPolicy() != Qt::ScrollBarAlwaysOff &&
+          !horizontalScrollBar()->isSliderDown()) {
+        hide_horizontal_scroll_bar();
+      }
+    }
+  } else if(event->type() == QEvent::HoverLeave) {
+    if(!m_h_scroll_bar_timer.isActive() &&
```
andrei-sa: Same as above, you can check for `!m_h_scroll_bar_timer.isActive()` only once.
Furthermore, the behavior here is similar to the `else` branch on `HoverMove`, check if you can restructure the condition tree to avoid code duplication.

Applications/Spire/Source/Ui/ScrollArea.cpp
```diff
@@ -0,0 +1,165 @@
...
+}
+
+void ScrollArea::wheelEvent(QWheelEvent* event) {
+  if(event->modifiers() & Qt::ShiftModifier) {
+    setVerticalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
+    setHorizontalScrollBarPolicy(Qt::ScrollBarAsNeeded);
+    horizontalScrollBar()->setValue(horizontalScrollBar()->value() -
+      (event->delta() / 2));
+    m_h_scroll_bar_timer.start();
+  } else if(!event->modifiers().testFlag(Qt::ShiftModifier)) {
```
andrei-sa: Is `event->modifiers() & Qt::ShiftModifier` any different from `event->modifiers().testFlag(Qt::ShiftModifier)`? API-wise, they seem to do the same thing. In case they do, you don't need the `else` condition and can move `m_h_scroll_bar_timer.start();` out.

Applications/Spire/Source/Ui/ScrollArea.cpp
```diff
@@ -0,0 +1,165 @@
...
+  m_v_scroll_bar_timer.stop();
+  setVerticalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
+}
+
+bool ScrollArea::is_within_horizontal_scroll_bar(int pos_y) {
+  if(widget() == nullptr) {
+    return false;
+  }
+  auto bar = verticalScrollBar();
+  auto scroll_adj = map_to(static_cast<double>(bar->value()),
```
andrei-sa: Avoid abbreviations.

Applications/Spire/Source/Ui/ScrollArea.cpp
```diff
@@ -0,0 +1,165 @@
...
+    static_cast<double>(bar->minimum()), static_cast<double>(bar->maximum()),
+    0, widget()->height() - height());
+  return pos_y - scroll_adj > height() - scale_height(SCROLL_BAR_MAX_SIZE);
```
andrei-sa: Would help if you gave explicit variable names to the sides.

Applications/Spire/Source/Ui/ScrollArea.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+bool ScrollArea::is_within_vertical_scroll_bar(int pos_x) {
+  if(widget() == nullptr) {
+    return false;
+  }
+  auto bar = horizontalScrollBar();
+  auto scroll_adj = map_to(static_cast<double>(bar->value()),
+    static_cast<double>(bar->minimum()), static_cast<double>(bar->maximum()),
+    0, widget()->width() - width());
+  return pos_x - scroll_adj > width() - scale_width(SCROLL_BAR_MAX_SIZE);
```
andrei-sa: Same as with `is_within_horizontal_scroll_bar`.

Applications/Spire/Source/Ui/ScrollArea.cpp
```diff
@@ -0,0 +1,165 @@
...
+    0, widget()->height() - height());
+  return pos_y - scroll_adj > height() - scale_height(SCROLL_BAR_MAX_SIZE);
+}
+
+bool ScrollArea::is_within_vertical_scroll_bar(int pos_x) {
+  if(widget() == nullptr) {
+    return false;
+  }
+  auto bar = horizontalScrollBar();
+  auto scroll_adj = map_to(static_cast<double>(bar->value()),
```
andrei-sa: Consider factoring this mapping routine out as the base is identical to `is_within_horizontal_scroll_bar`.

Applications/Spire/Source/Ui/ScrollArea.cpp
spireken: I can change it like this:

```
  if(event->modifiers().testFlag(Qt::ShiftModifier)) {
    ...scroll horizontally...
  } else {
    ...scroll vertically...
  }
```

But each condition does something different so I can't remove anything else.

andrei-sa: Oh, `h_` looks like "handle" from my point, so I didn't notice those are 2 different timers.
This comment mostly referred to the unnecessary check in the `else` condition itself.
Can you, however, get rid of those abbreviations?

Applications/Spire/Source/Ui/ScrollArea.cpp
spireken: I'm not sure I can restructure this and make it clearer. There's subtle differences in the conditions for each branch.

andrei-sa: Ok, let us keep it as is.

Applications/Spire/Source/Ui/ScrollArea.cpp
```diff
@@ -0,0 +1,156 @@
...
+  hide_horizontal_scroll_bar();
+  hide_vertical_scroll_bar();
+}
+
+void ScrollArea::wheelEvent(QWheelEvent* event) {
+  if(event->modifiers().testFlag(Qt::ShiftModifier)) {
+    setVerticalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
+    setHorizontalScrollBarPolicy(Qt::ScrollBarAsNeeded);
+    horizontalScrollBar()->setValue(horizontalScrollBar()->value() -
+      (event->delta() / 2));
```
kranar: I am worried about the accumulation of errors here. If the delta is an odd number then this is off by 1. Imagine if the delta is always 1, then no change ever occurs.

spireken: It looks like delta() was actually deprecated. I can use the [angle delta](https://doc.qt.io/qt-5/qwheelevent.html#angleDelta) instead.

What about taking the larger of (angleDelta / 2) or 1?

kranar: You can use angle delta, that's fine, but it doesn't solve the accumulation of errors. To solve that you will need to store the accumulated error and add it back in. Otherwise you will need to find some objective value to use to update the scrollbar's value.

spireken: Or -1, if the angleDelta is negative.

kranar: Keep track of the rounding error and add it back in when it's equal to 1.

spireken: Would it look something like this:

`
auto delta = angleDelta() / 2 + accumulated_error;
accumulated_error = modf(delta, &delta);
scrollBar()->setValue(delta);
`

spireken: Sorry, I keep missing your replies, the page isn't refreshing to show them.

kranar: Yeah something like that is about right. accumulated error should just be either 0 or 0.5.

spireken:
I'm not sure how to verify this because my mouse works in 30 * 1/8deg 'ticks'.

Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
```diff
@@ -0,0 +1,45 @@
...
+      explicit ScrollArea(QWidget* parent = nullptr);
+
+      //! Sets the widget displayed in the scroll area.
+      /*
+        \param widget The widget to display.
+      */
+      void setWidget(QWidget* widget);
+
+    protected:
+      bool eventFilter(QObject* watched, QEvent* event);
```
kranar: ```override```

Applications/Spire/Source/Ui/ScrollArea.cpp
```diff
@@ -0,0 +1,167 @@
...
+void ScrollArea::leaveEvent(QEvent* event) {
+  hide_horizontal_scroll_bar();
+  hide_vertical_scroll_bar();
+}
+
+void ScrollArea::wheelEvent(QWheelEvent* event) {
+  if(event->modifiers().testFlag(Qt::ShiftModifier)) {
+    setVerticalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
+    setHorizontalScrollBarPolicy(Qt::ScrollBarAsNeeded);
+    auto delta = event->angleDelta().y() / 2 + m_horizontal_scrolling_error;
```
kranar: Remove line 89.

Applications/Spire/Source/Ui/ScrollArea.cpp
```diff
@@ -0,0 +1,167 @@
...
+    static_cast<double>(scroll_bar->minimum()),
+    static_cast<double>(scroll_bar->maximum()), 0, scroll_size - widget_size);
+  return pos - scroll_adjustment > widget_size -
+    scale_width(SCROLL_BAR_MAX_SIZE);
+}
+
+void ScrollArea::update_scrollbar_position(QScrollBar* scroll_bar, int delta,
+    double& scrolling_error) {
+  auto adjusted_delta = delta / 2 + scrolling_error;
+  scrolling_error = modf(adjusted_delta, &adjusted_delta);
```
kranar: ```std::modf```

kranar:
Is this ready to be merged?

spireken:
You're asking Andre, right?

kranar:
Either of you, looks like changes are still requested for this pull request so wondering what is left to do.

andrei-sa:
I have no further concerns. Will explicitly approve changes next time.

PR #414, opened 2020-03-24
Implement KeyBindings

andrei-sa:


Applications/Spire/Include/Spire/KeyBindings/KeyBindings.hpp
```diff
@@ -4,12 +4,15 @@
...
 #include <vector>
 #include <boost/optional/optional.hpp>
-#include <Nexus/Definitions/Definitions.hpp>
-#include <Nexus/Definitions/OrderType.hpp>
-#include <Nexus/Definitions/Region.hpp>
-#include <Nexus/Definitions/Side.hpp>
-#include <Nexus/Definitions/Tag.hpp>
 #include <QKeySequence>
+#include <QHash>
+#include "Nexus/Definitions/Definitions.hpp"
```
andrei-sa: Formatted Nexus includes as internal as everywhere else in Spire.

Applications/Spire/Include/Spire/KeyBindings/KeyBindings.hpp
```diff
@@ -136,7 +139,7 @@ namespace Details {
         \param sequence The key sequence.
         \param action The action.
       */
-      void set_binding(QKeySequence sequence, Action action);
+      void set_binding(QKeySequence sequence, const Action& action);
```
andrei-sa: Changed to const reference here because the internal data structure does not consume rvalues.

Applications/Spire/Source/KeyBindings/KeyBindings.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+boost::optional<const KeyBindings::Action&>
+    KeyBindings::find_binding(const Region& region,
+    const QKeySequence& sequence) const {
+  auto i = m_bindings.find(sequence);
+  if(i == m_bindings.end()) {
+    return boost::none;
+  }
+  auto& action = i->Get(region);
+  if(action) {
```
andrei-sa: It might be a good idea to return `boost::optional<Action>` by copy instead. Optional references are not really safer than pointers, nor are they easier to comprehend and deal with. Furthermore, `std::optional` does not support references and there is a chance `boost::optional` will not as well later down the road.

Applications/Spire/Source/KeyBindings/KeyBindings.cpp
```diff
@@ -0,0 +1,85 @@
+#include "Spire/KeyBindings/KeyBindings.hpp"
+#include <algorithm>
+#include <type_traits>
+
+using namespace Nexus;
+using namespace Spire;
```
kranar: Good good, I prefer ```using namespace ...``` with the only exception being no ```using namespace std;```.

Applications/Spire/Source/KeyBindings/KeyBindings.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+KeyBindings::ActionBindingsList
+    KeyBindings::build_action_bindings_list() const {
+  auto list = ActionBindingsList();
+  std::for_each(m_bindings.constKeyValueBegin(), m_bindings.constKeyValueEnd(),
+    [&] (auto& e) {
+      auto& elements = e.second;
+      auto region_actions = std::vector<std::tuple<const Region,
+        Actions::Element>>();
+      std::copy_if(elements.Begin(), elements.End(),
```
andrei-sa: I did not optimize these copies out because we return a copy anyway.

Applications/Spire/Source/KeyBindings/KeyBindings.cpp
kranar: ```return i->Get(region);```

Applications/Spire/Source/KeyBindings/KeyBindings.cpp
```diff
@@ -0,0 +1,85 @@
...
+  if(action) {
+    return *action;
+  } else {
+    return boost::none;
+  }
+}
+
+KeyBindings::OrderBindingsList
+    KeyBindings::build_order_bindings_list() const {
+  auto list = OrderBindingsList();
```
kranar: Incomplete.

Applications/Spire/Source/KeyBindings/KeyBindings.cpp
```diff
@@ -0,0 +1,85 @@
...
+  return list;
+}
+
+KeyBindings::CancelBindingsList
+    KeyBindings::build_cancel_bindings_list() const {
+  auto list = CancelBindingsList();
```
kranar: Incomplete.

andrei-sa:
I will add test cases for `build_xxx_bindings_list` methods shortly.

Applications/Spire/Source/KeyBindings/KeyBindings.cpp
kranar: A single for loop can be used to consolidate the ```copy_if``` and the ```transform```.

Applications/Spire/Include/Spire/KeyBindings/KeyBindings.hpp
```diff
@@ -4,12 +4,15 @@
 #include <variant>
 #include <vector>
 #include <boost/optional/optional.hpp>
-#include <Nexus/Definitions/Definitions.hpp>
-#include <Nexus/Definitions/OrderType.hpp>
-#include <Nexus/Definitions/Region.hpp>
-#include <Nexus/Definitions/Side.hpp>
-#include <Nexus/Definitions/Tag.hpp>
 #include <QKeySequence>
+#include <QHash>
```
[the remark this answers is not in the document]
andrei-sa: Wow, I really need to learn that alphabet.

Applications/Spire/Source/KeyBindings/KeyBindings.cpp
andrei-sa: For that to work, I need to return `boost::optional<Action>` rather than `boost::optional<const Action&>`. Let me know if I should change the API in this way.

kranar: Change the API then.

Applications/Spire/Source/KeyBindings/KeyBindings.cpp
```diff
@@ -37,30 +30,50 @@ void KeyBindings::reset_binding(const Region& region,
...
 }
 
 KeyBindings::OrderBindingsList
     KeyBindings::build_order_bindings_list() const {
   auto list = OrderBindingsList();
+  std::for_each(m_bindings.constKeyValueBegin(), m_bindings.constKeyValueEnd(),
+    [&] (auto& e) {
+      auto& elements = e.second;
+      for(auto i = elements.Begin(); i != elements.End(); ++i) {
+        auto& action = std::get<1>(*i);
```
kranar: Prefer to use:

    if(auto& action = std::get<1>(*i)) {
      if(auto order_action = std::get_if(...)) {
        ...
      }
    }

PR #415, opened 2020-03-25
Switched DropShadow instantiation from managed to raw pointers

spireken:
I put a build in the drafts/demos folder in case it's needed.

I tested the callback using the context menu; when the context menu is hidden/destroyed there's two calls to deleteLater(), one for each menu's drop shadow.

### DIFF
--- a/Applications/Spire/Include/Spire/Ui/DropShadow.hpp
+++ b/Applications/Spire/Include/Spire/Ui/DropShadow.hpp
@@ -6,7 +6,8 @@
 
 namespace Spire {
 
-  //! Paints a drop shadow around its parent widget.
+  //! Paints a drop shadow around its parent widget. The drop shadow instance
+  //! is deleted when its parent is deleted.
   class DropShadow : public QWidget, public QAbstractNativeEventFilter {
     public:
 
--- a/Applications/Spire/Include/Spire/Ui/DropShadow.hpp
+++ b/Applications/Spire/Include/Spire/Ui/DropShadow.hpp
@@ -6,8 +6,7 @@
 
 namespace Spire {
 
-  //! Paints a drop shadow around its parent widget. The drop shadow instance
-  //! is deleted when its parent is deleted.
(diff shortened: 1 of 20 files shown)


Applications/Spire/Include/Spire/Ui/DropShadow.hpp
```diff
@@ -6,7 +6,8 @@
 
 namespace Spire {
 
-  //! Paints a drop shadow around its parent widget.
+  //! Paints a drop shadow around its parent widget. The drop shadow instance
+  //! is deleted when its parent is deleted.
```
kranar: Don't need to specify the deletion policy, this is how all ```QObject```s behave.

Applications/Spire/Source/Ui/DropShadow.cpp
```diff
@@ -43,6 +43,9 @@ DropShadow::DropShadow(bool is_menu_shadow, bool has_top, QWidget* parent)
   setAttribute(Qt::WA_TranslucentBackground);
   setAttribute(Qt::WA_ShowWithoutActivating);
   m_parent->window()->installEventFilter(this);
+  connect(m_parent,
+    static_cast<void (QObject::*)(QObject*)>(&QObject::destroyed),
```
kranar: Just to confirm, the following didn't work?

    connect(m_parent, &QObject::destroyed, ...);

spireken: No, the compiler complained something to the effect of 'this doesn't evaluate to an expression taking one parameter'.

kranar: Good, use ```qOverload``` from now on:

    connect(m_parent, qOverload<>(QObject::destroyed), ...);

spireken:
Just verified it. The line has to be broken up (deleteLater() on its own line) in VS otherwise it only triggers the breakpoint on construction.

PR #416, opened 2020-03-26
Scroll area usage

spireken:
I put the builds in the drafts/demos folder. The toolbar window is just there to show its stability after removing the unique_ptr use.

The book view window had the two lists of its properties widgets updated with the new scroll area.

The charting window is to demo that the drop down box is still functioning correctly. There's no functionality so far to limit how many items are displayed, so the scroll bars can't be viewed right now.

The time and sales window was updated to use the dynamic scroll bars.

The security input was updated and can be viewed in one of the tester builds.

### DIFF
--- a/Applications/Spire/Include/Spire/Spire/Utility.hpp
+++ b/Applications/Spire/Include/Spire/Spire/Utility.hpp
@@ -1,7 +1,20 @@
 #ifndef SPIRE_UTILITY_HPP
 #define SPIRE_UTILITY_HPP
+#include <QObject>
 
 namespace Spire {
+
+  //! Deletes an object when control returns to the thread's event loop.
+  /*
+    \param object The object to schedule for deletion.
+  */
+  template<typename T>
+  std::enable_if_t<std::is_base_of_v<QObject, T>> deleteLater(T*& object) {
+    if(object != nullptr) {
+      object->deleteLater();
+      object = nullptr;
+    }
+  }
 
   //! Returns the minimum of the given parameters. Returns the first given
   //! parameter if they are equal.
--- a/Applications/Spire/Include/Spire/Spire/Utility.hpp
+++ b/Applications/Spire/Include/Spire/Spire/Utility.hpp
@@ -9,7 +9,7 @@ namespace Spire {
     \param object The object to schedule for deletion.
   */
   template<typename T>
-  std::enable_if_t<std::is_base_of_v<QObject, T>> deleteLater(T*& object) {
+  std::enable_if_t<std::is_base_of_v<QObject, T>> delete_later(T*& object) {
     if(object != nullptr) {
       object->deleteLater();
       object = nullptr;
--- a/Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
+++ b/Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
@@ -27,7 +27,7 @@ namespace Spire {
       //! Sets the scroll area's border style.
       /*
         \param width The border width.
-        \param color the Border color.
+        \param color The border color.
       */
       void set_border_style(int width, const QColor& color);
 
--- a/Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
+++ b/Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
@@ -24,6 +24,13 @@ namespace Spire {
       */
       ScrollArea(bool is_dynamic, QWidget* parent = nullptr);
 
+      //! Sets the scroll area's border style.
+      /*
+        \param width The border width.
+        \param color the Border color.
+      */
+      void set_border_style(int width, const QColor& color);
+
       //! Sets the widget displayed in the scroll area.
       /*
         \param widget The widget to display.
@@ -41,6 +48,8 @@ namespace Spire {
       QTimer m_vertical_scroll_bar_timer;
       double m_horizontal_scrolling_error;
       double m_vertical_scrolling_error;
+      QColor m_border_color;
+      int m_border_width;
 
       void hide_horizontal_scroll_bar();
       void hide_vertical_scroll_bar();
--- a/Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
+++ b/Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
@@ -22,7 +22,7 @@ namespace Spire {
                           bars.
         \param parent The parent widget.
       */
-      ScrollArea(bool is_dynamic, QWidget* parent = nullptr);
+      explicit ScrollArea(bool is_dynamic, QWidget* parent = nullptr);
 
       //! Sets the scroll area's border style.
       /*
--- a/Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
+++ b/Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
@@ -9,11 +9,20 @@ namespace Spire {
   class ScrollArea : public QScrollArea {
     public:
 
-      //! Constructs an empty ScrollArea.
+      //! Constructs an empty ScrollArea with static scroll bars.
       /*
         \param parent The parent widget.
       */
       explicit ScrollArea(QWidget* parent = nullptr);
+      
+      //! Constructs an empty ScrollArea.
+      /*
+        \param is_dynamic True if the scroll area should support dynamic-sized,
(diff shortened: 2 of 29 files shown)


Applications/Spire/Source/BookView/BookViewController.cpp
```diff
@@ -36,7 +37,8 @@ void BookViewController::close() {
   if(m_window == nullptr) {
     return;
   }
-  m_window.reset();
+  m_window->deleteLater();
+  m_window = nullptr;
```
andrei-sa: It does not seem like these resets (the `nullptr` part) are always necessary, are they? It only makes a difference if there are explicit checks against `nullptr` somewhere afterwards.
Where necessary, you may consider wrapping these two statements into an inline function.

Applications/Spire/Source/Ui/DropShadow.cpp
```diff
@@ -44,7 +44,7 @@ DropShadow::DropShadow(bool is_menu_shadow, bool has_top, QWidget* parent)
   setAttribute(Qt::WA_ShowWithoutActivating);
   m_parent->window()->installEventFilter(this);
   connect(m_parent, qOverload<QObject*>(&QObject::destroyed),
-    [=] (auto object) { deleteLater(); });
+    this, &DropShadow::on_parent_destroyed);
```
andrei-sa: Why do you need a method now?

Applications/Spire/Source/Ui/SecurityWidget.cpp
```diff
@@ -98,23 +102,29 @@ void SecurityWidget::show_overlay_widget() {
 }
 
 void SecurityWidget::hide_overlay_widget() {
-  m_overlay_widget.reset();
+  if(m_overlay_widget != nullptr) {
```
andrei-sa: This entire test delete routine may be factored out as well.

Applications/Spire/Source/BookView/BookViewController.cpp
kranar: Feel free to add a free function in Utility.hpp called ```deleteLater(QObject*& object)``` to do this.

Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
```diff
@@ -9,11 +9,27 @@ namespace Spire {
...
       explicit ScrollArea(QWidget* parent = nullptr);
+      
+      //! Constructs an empty ScrollArea.
+      /*
+        \param is_dynamic True if the scroll area should support dynamic-sized,
+                          auto-hiding scroll bars. False for static scroll
+                          bars.
+        \param parent The parent widget.
+      */
+      ScrollArea(bool is_dynamic, QWidget* parent = nullptr);
```
andrei-sa: Mark `explicit`.

Applications/Spire/Source/BookView/BookViewController.cpp
spireken: I think I understand the purpose of taking a reference to a pointer here (can reassign the original pointer), but how do I pass the object to this function? Passing in the window in BookViewController (type: BookViewWindow*) to this function gives me the error:

a reference of type "QObject*&" (not const-qualified) cannot be initialized with a value of type "Spire::BookViewWindow*"

The BookViewWindow is obviously derived from a QObject, so shouldn't the function parameter required it to be passed as a reference?

kranar: Yes that's right, what I said is not legal or possible (bonus points if it makes sense to you why it's not possible). You will need to make it a template function.

    template<typename T>
    std::enable_if_t<std::is_base_of_v<QObject, T>> deleteLater(T*& object) {
      ...;
    }

Applications/Spire/Source/Ui/DropShadow.cpp
spireken: I was having issues with what looked like the lambda getting prematurely destroyed.

Applications/Spire/Source/BookView/BookViewController.cpp
spireken: Is the issue potentially reassigning the reference?

kranar: Yep, you got it! Very nice Sir.

spireken: So how does the template version work, then? The template limits the types to derivatives of QObject, but they're still references to pointers.

kranar: Okay sure, let's be more detailed about it then.  Imagine I had the following:

    void f(QObject*& object) {
      object = new QTimer(...);
    }

    void g() {
      auto q = new QScrollBar();
      f(q);
    }

Well it's clear why ```f(q)``` won't work, because ```f``` assigns to ```object``` a ```QTimer```, but we called ```f``` from within ```g``` with a ```QScrollBar```, so this is clearly not something that should be possible. Even though it's legal to assign a ```QTimer``` to a ```QObject```, it is not possible to assign a ```QTimer``` to a ```QScrollBar```.

However, if instead we had declared ```f``` as follows:

    void f(QObject*& const object);

Then this would be fine, because ```object``` can not be reassigned since it's ```const```, so there is no type safety violation.

spireken: That makes sense.

andrei-sa: Why not just `void deleteLater(QObject** object)` ?

Applications/Spire/Source/Ui/DropShadow.cpp
andrei-sa: A lambda, as a syntactic construct, cannot be destroyed per se. The closure it is assigned to can, but I think `connect` takes functors by copy so it should not be the case (and otherwise you would get some warnings). I think the only difference is that you pass the context `this` in the newer version, but you can do so with a lambda as well. Have you tried
```
  connect(m_parent, qOverload<QObject*>(&QObject::destroyed),
    [=] (auto object) { deleteLater(); });
    this, [=] (auto object) { deleteLater(); });
```
?

spireken: This was actually related to an issue I fixed in the key bindings branch. I'll revert it back to using the lambda.

spireken: I missed your comment before I replied, but the issue of semantics is something I should keep in mind.

spireken:
I'm getting denied access when uploading the TimeAndSalesUiTester, but added the book view and charting testers no problem.

Applications/Spire/Include/Spire/Spire/Utility.hpp
```diff
@@ -1,8 +1,21 @@
...
+#include <QObject>
 
 namespace Spire {
 
+  //! Deletes an object when control returns to the thread's event loop.
+  /*
+    \param object The object to schedule for deletion.
+  */
+  template<typename T>
+  std::enable_if_t<std::is_base_of_v<QObject, T>> deleteLater(T*& object) {
```
andrei-sa: It seems like snake_case should be used for free functions, and there won't be conflicts with `deleteLater` method so you can remove the `Spire::` prefix.

Applications/Spire/Source/BookView/BookViewController.cpp
andrei-sa: Even with the reference version though, why do we need to explicitly disable the template for types that are not subclasses of `QObject`? Isn't it an example of type pessimization we normally avoid?

kranar: I place a very high value on type safety. Once compiler support for C++20 and concepts becomes mainstream then there will be a major refactoring of generic code to be a lot more type safe than it currently is.

Applications/Spire/Source/BookView/BookQuoteTableView.cpp
```diff
@@ -8,9 +8,10 @@
 using namespace Spire;
 
 BookQuoteTableView::BookQuoteTableView(
-    std::unique_ptr<BookQuoteTableModel> model, QWidget* parent)
+    BookQuoteTableModel* model, QWidget* parent)
     : QTableView(parent),
       m_model(std::move(model)) {
```
kranar: Don't use ```move``` on a pointer.

PR #417, opened 2020-03-27
Key bindings window

spireken:
I put the book view and time and sales demos in the drafts/demos folder called 'XYUpdatedSignals' to prove they function correctly.

Returning '{}' from get_key_bindings() gives me a warning about returning a local object by reference, so I added the m_key_bindings member to get it to compile. This member probably won't be in the actual implementation.

### DIFF
--- a/Applications/Spire/Include/Spire/KeyBindings/KeyBindingsWindow.hpp
+++ b/Applications/Spire/Include/Spire/KeyBindings/KeyBindingsWindow.hpp
@@ -5,6 +5,7 @@
 
 namespace Spire {
 
+  //! Displays the user's key bindings.
   class KeyBindingsWindow : public Window {
     public:
 
@@ -14,11 +15,18 @@ namespace Spire {
       */
       using ApplySignal = Signal<void (const KeyBindings& binding)>;
 
-      explicit KeyBindingsWindow(const KeyBindings& key_binding,
+      //! Constructs a key bindings window.
+      /*
+        \param key_bindings The initial key bindings.
+        \param parent The parent widget.
+      */
+      explicit KeyBindingsWindow(const KeyBindings& key_bindings,
         QWidget* parent = nullptr);
 
+      //! Returns the current key bindings.
       const KeyBindings& get_key_bindings() const;
 
+      //! Connects a slot to the applied signal.
       boost::signals2::connection connect_apply_signal(
         const ApplySignal::slot_type& slot) const;
 
--- a/Applications/Spire/Include/Spire/KeyBindings/KeyBindingsWindow.hpp
+++ b/Applications/Spire/Include/Spire/KeyBindings/KeyBindingsWindow.hpp
@@ -1,0 +1,31 @@
+#ifndef SPIRE_KEY_BINDINGS_WINDOW_HPP
+#define SPIRE_KEY_BINDINGS_WINDOW_HPP
+#include "Spire/KeyBindings/KeyBindings.hpp"
+#include "Spire/Ui/Window.hpp"
+
+namespace Spire {
+
+  class KeyBindingsWindow : public Window {
+    public:
+
+      //! Signals that the user applied the current key bindings.
+      /*
+        /param key_bindings The applied key bindings.
+      */
+      using ApplySignal = Signal<void (const KeyBindings& binding)>;
+
+      explicit KeyBindingsWindow(const KeyBindings& key_binding,
+        QWidget* parent = nullptr);
+
+      const KeyBindings& get_key_bindings() const;
+
+      boost::signals2::connection connect_apply_signal(
+        const ApplySignal::slot_type& slot) const;
+
+    private:
+      KeyBindings m_key_bindings;
+      mutable ApplySignal m_apply_signal;
+  };
+}
+
+#endif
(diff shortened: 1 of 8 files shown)


Applications/Spire/Source/KeyBindings/KeyBindingsWindow.cpp
```diff
@@ -0,0 +1,18 @@
+#include "Spire/KeyBindings/KeyBindingsWindow.hpp"
+
+using namespace boost::signals2;
+using namespace Spire;
+
+KeyBindingsWindow::KeyBindingsWindow(const KeyBindings& key_bindings,
+  QWidget* parent)
```
andrei-sa: I'm not sure which indentation is correct, but most often in Spire I see 4 whitespaces on both the continuation line and before the colon.

Applications/Spire/Source/KeyBindings/KeyBindingsWindow.cpp
```diff
@@ -0,0 +1,18 @@
...
+  : Window(parent),
+    m_key_bindings(key_bindings) {}
+
+const KeyBindings& KeyBindingsWindow::get_key_bindings() const {
+  return m_key_bindings;
```
andrei-sa: Not that important really, but if you want a pure stub without touching the header file, you can always create a static instance and return a reference to it.
Feel free to keep at as is.

Applications/Spire/Source/KeyBindings/KeyBindingsWindow.cpp
kranar: Use two levels of indentation only when there is a need to disambiguate the body of an expression with its "introduction".

In this case the introduction is the initializer, but the body is empty, so there is no need to disambiguate anything. However if the body was not empty then the initializer would need to be double indented to make clear the distinction between the initializer and the body.

Similar thing goes for other introducers such as ```if```, ```for```, etc...

Applications/Spire/Include/Spire/KeyBindings/KeyBindingsWindow.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+namespace Spire {
+
+  //! Displays the user's key bindings.
+  class KeyBindingsWindow : public Window {
+    public:
+
+      //! Signals that the user applied the current key bindings.
+      /*
+        /param key_bindings The applied key bindings.
```
kranar: Use ```\``` over ```/```

Applications/Spire/Include/Spire/KeyBindings/KeyBindingsWindow.hpp
```diff
@@ -0,0 +1,39 @@
...
+      */
+      using ApplySignal = Signal<void (const KeyBindings& binding)>;
```
kranar: Remove the parameter.

Applications/Spire/Source/BookView/BookViewWindow.cpp
```diff
@@ -131,7 +131,7 @@ void BookViewWindow::show_context_menu(const QPoint& pos) {
 void BookViewWindow::show_properties_dialog() {
   BookViewPropertiesDialog dialog(get_properties(), Security(), this);
   m_dialog_apply_connection = dialog.connect_apply_signal(
-    [=] (auto p) { set_properties(p); });
+    [=, &dialog] { set_properties(dialog.get_properties()); });
```
kranar: Capture by ```&```, ie. replace ```[=, &dialog]``` with ```[&]```. Try to understand why that's the case.

Applications/Spire/Source/TimeAndSales/TimeAndSalesWindow.cpp
```diff
@@ -164,7 +164,9 @@ void TimeAndSalesWindow::export_table() {
 
 void TimeAndSalesWindow::show_properties_dialog() {
   TimeAndSalesPropertiesDialog dialog(m_properties, this);
-  dialog.connect_apply_signal([=] (auto p) { set_properties(p); });
+  dialog.connect_apply_signal([=, &dialog] {
```
kranar: Similar principle applies, also fix indentation.

spireken: Do we not double indent when breaking up lambdas into multiple lines?

kranar: Nope.

Applications/Spire/Source/BookView/BookViewWindow.cpp
spireken: As far as I know in this case the lambda is capturing 'this' and the dialog by reference. What's the difference in capturing 'this' be reference?

kranar: The lambda is capturing ```this``` by value and ```dialog``` by reference. My comment is that everything should be captured by reference as so to simply use ```[&]``` as the capture.

spireken: Sorry, my comment was poorly worded. I meant 'this' and the dialog were being captured, but only the dialog was being captured by reference.

kranar: Yep, they should both be captured by reference since the closure does not escape the lambda expression's lexical scope. By default (as-in unless you have a good reason to) if the closure does not escape the lambda's lexical scope then the capture should be by reference. If the closure does escape the lambda's lexical scope, then the capture should be by value.

Feel free to give some thought as to why this is and do your best to remember this principle.

PR #422, opened 2020-04-02
Added loading pages to controllers.

spiresheep:


Applications/WebPortal/WebApp/library/source/pages/application_controller.tsx
```diff
@@ -36,7 +37,7 @@ export class ApplicationController extends React.Component<Properties, State> {
 
   public render(): JSX.Element {
     if(this.state.isLoading) {
-      return <div></div>;
+      return <LoadingPage/>;
```
kranar, line 40: Keep an empty ```<div/>``` here. This is the landing page and there is no suitable loading page for it (there may be one designed in the future).

PR #427, opened 2020-04-14
Account dir controller

spiresheep:


Applications/WebPortal/WebApp/library/source/pages/account_directory_page/account_directory_controller.tsx
```diff
@@ -18,13 +32,55 @@ export class AccountDirectoryController extends
...
+      onFilterChange={this.onFilterChange} 
+      onCardClick={this.onCardClick}/>
+  }
+
+  private async onCardClick(group: Beam.DirectoryEntry) {
+    if(this.state.openedGroups.get(group)) {
+      this.state.openedGroups.remove(group);
+    } else {
+      this.timerId = setTimeout(
+        async () => {
```
kranar, line 62: Remove the timeout.  That might have been there for testing purposes but won't be needed any more.

spiresheep:
It does not work properly. I'm going to close this so we can work this out in asana.

PR #434, opened 2020-04-30
Repeatable autofocus

spiresheep:
There was a issue where the focus would only occur the first time. Autofocus on the input didn't provide the right behavior. So, I'm using a workaround of granting focus on componentDidMount. The slight delay is needed otherwise when the modal is opened the second time focus isn't there.

### DIFF
--- a/Applications/WebPortal/WebApp/library/source/components/text_field.tsx
+++ b/Applications/WebPortal/WebApp/library/source/components/text_field.tsx
@@ -79,7 +79,7 @@ export class TextField extends React.Component<Properties> {
 
   public componentDidMount() {
     if(this.props.autoFocus) {
-      setTimeout(()=>{this.inputRef.current.focus();}, 1);
+      setTimeout(()=>{this.inputRef.current.focus();}, 0);
     }
   }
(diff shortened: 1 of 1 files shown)


Applications/WebPortal/WebApp/library/source/components/text_field.tsx
```diff
@@ -62,15 +68,21 @@ export class TextField extends React.Component<Properties> {
...
+        ref={this.inputRef}
         onChange={(event: React.ChangeEvent<HTMLInputElement>) => {
           this.props.onInput(event.target.value);}}
         className={css(TextField.EXTRA_STYLE.effects, errorStyle) + ' ' +
           this.props.className}/>);
   }
 
+  public componentDidMount() {
+    if(this.props.autoFocus) {
+      setTimeout(()=>{this.inputRef.current.focus();}, 1);
```
kranar, line 82: Can the value be changed from 1 to 0?

spiresheep: Yup.

PR #437, opened 2020-05-05
Replace typedef with using

andrei-sa:


Nexus/Include/Nexus/Definitions/Tick.hpp
```diff
@@ -1,30 +1,24 @@
...
+    //! Up tick.
+    UP,
 
-      //! Down tick.
-      DOWN
-    };
-  };
-}
-
-  typedef Beam::Enumerator<Details::TickDirectionDefinition> TickDirection;
```
andrei-sa, line 27: Could not find the definition of `Beam::Enumerator` anywhere in the repository.

Nexus/Include/Nexus/MarketDataService/MarketOrderImbalanceExpression.hpp
```diff
@@ -30,20 +30,20 @@ namespace MarketDataService {
   class MarketOrderImbalanceExpression :
       public Beam::Expressions::IterationExpression<OrderImbalance> {
```
andrei-sa, line 31: Could not find `Beam::Expressions` either, so this file does not really compile.

Nexus/Include/Nexus/Definitions/Tick.hpp
kranar: Delete this file, it's dead code.

Nexus/Include/Nexus/MarketDataService/MarketOrderImbalanceExpression.hpp
kranar: This is also dead code, delete it.

PR #441, opened 2020-05-14
Dest column fix

spireken:
I'm sure this will be overwritten with the Destination column case, but here's a fix for the Destination column requiring a Security before accepting input.

### DIFF
--- a/Applications/Spire/Source/KeyBindings/TaskKeyBindingsTableModel.cpp
+++ b/Applications/Spire/Source/KeyBindings/TaskKeyBindingsTableModel.cpp
@@ -230,6 +230,9 @@ bool TaskKeyBindingsTableModel::setData(const QModelIndex& index,
               binding.m_region = *(binding.m_region.GetSecurities().begin());
             }
           } else {
+            if(binding.m_region.GetSecurities().empty()) {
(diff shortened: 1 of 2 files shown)


Applications/Spire/Source/KeyBindings/TaskKeyBindingsTableModel.cpp
```diff
@@ -222,15 +222,18 @@ bool TaskKeyBindingsTableModel::setData(const QModelIndex& index,
...
       case Columns::DESTINATION:
         {
-          auto name = value.toString();
-          if(name.isEmpty()) {
+          auto name = value.value<Region>().GetName();
+          if(name.empty()) {
             if(binding.m_region.GetSecurities().empty()) {
               binding.m_region = Region();
             } else {
               binding.m_region = *(binding.m_region.GetSecurities().begin());
```
kranar, line 230: Use ```front()``` over ```begin()```

spireken: GetSecurities() returns an unordered_set.

spireken:
Can this be merged?

PR #449, opened 2020-05-26
Date Time Fix

spiresheep:
The code isn't compiling because of some changes. This fixes it so it compiles.


kranar:
Most likely you need to run the configure.sh script. Anytime you do a git pull, you need to run the configure.sh script to ensure your dependencies are up to date.

PR #450, opened 2020-05-26
Cached chart model

andrei-sa:
For `LocalChartModel`, I did a straightforward implementation requiring full linear search for TAIL queries. In general, it is tough to do well for arbitrary queries (uniformly distributed start and end times or something like that), but it is possible to approximate logarithmic complexity as long as the average squared width of the candlesticks is not much greater than the average squared distance between two consecutive candlesticks. This pattern should match most UI needs, especially when we are talking about non-overlapping and fixed width candlesticks. I will implement that in a separate PR.

As for the cached model, there are some details I did not cover previously. Probably the most important one is the openness of intervals registered in the cached interval tree, which depends on whether the limit has been reached with a model query. The new test case covers the scenario where a candlestick is omitted by two surrounding queries and then included by an unlimited query spanning the time point, which only became possible due to registering cached intervals as semi-open.
I did not use `boost::icl` for `queried_intervals` because the former only works with aggregators, i.e. there is a possibility of merging intervals while combining their satellite data. By design, future arrays stored as the query interval map satellite data cannot be aggregated, nor is it possible for queried intervals to overlap, so I used a start-keyed balanced tree to record queried intervals.


Applications/Spire/Source/Charting/CachedChartModel.cpp
```diff
@@ -1,141 +1,277 @@
...
 Scalar::Type CachedChartModel::get_y_axis_type() const {
-  return m_chart_model->get_y_axis_type();
+  return m_cache.get_y_axis_type();
 }
 
 QtPromise<std::vector<Candlestick>> CachedChartModel::load(Scalar first,
     Scalar last, const SnapshotLimit& limit) {
-  return load_from_cache({first, last, first, last, limit});
+  if(limit.GetSize() == 0) {
+    return QtPromise([] {
```
kranar, line 54: return QtPromise(std::vector<Candlestick>());

Applications/Spire/Source/Charting/CachedChartModel.cpp
```diff
@@ -1,141 +1,277 @@
...
+  auto interval = [&] {
+    if(first == last) {
+      return Interval::closed(first, last);
+    } else {
+      return Interval::open(first, last);
+    }
+  }();
+  auto [bound, cached_interval, queried_interval] = [&] {
+    if(limit.GetType() == SnapshotLimit::Type::HEAD) {
+      return std::make_tuple(first, find_leftmost_cached_overlap(interval),
```
kranar, line 67: Prefer ```std::tuple(...)``` over ```std::make_tuple```.

Applications/Spire/Include/Spire/Charting/CachedChartModel.hpp
```diff
@@ -31,26 +39,37 @@ namespace Spire {
...
-      QtPromise<std::vector<Candlestick>> load_from_model(
-        const LoadInfo& info);
-      void on_data_loaded(const std::vector<Candlestick>& data,
-        const boost::icl::continuous_interval<Scalar>& loaded_range);
-      void on_data_loaded(const std::vector<Candlestick>& data,
-        const LoadInfo& info);
+      boost::icl::interval_set<Scalar> m_cached_intervals;
+      std::map<Scalar, QueryInfo> m_queried_intervals;
+
+      QtPromise<std::vector<Candlestick>> on_cached_interval(
```
kranar, line 54: ```on_xxx``` is reserved for event handlers, like ```on_mouse_event```. Rename these to ```load_xxx```.

Applications/Spire/Source/Charting/CachedChartModel.cpp
```diff
@@ -1,141 +1,277 @@
...
+  } else {
+    return on_new_interval(interval, cached_interval, queried_interval,
+      limit);
+  }
 }
 
 connection CachedChartModel::connect_candlestick_slot(
     const CandlestickSignal::slot_type& slot) const {
-  return m_chart_model->connect_candlestick_slot(slot);
+  return connection();
```
kranar, line 86: return m_model.connect_candlestick_slot(slot);

Applications/Spire/Source/Charting/CachedChartModel.cpp
```diff
@@ -1,141 +1,277 @@
...
-            break;
-          }
-        }
+QtPromise<std::vector<Candlestick>> CachedChartModel::on_cached_interval(
+    const Interval& interval, const Interval& cached_interval,
+    const SnapshotLimit& limit) {
+  auto is_head = limit.GetType() == SnapshotLimit::Type::HEAD;
+  auto [start, end] = [&] {
+    if(is_head) {
+      return std::make_pair(interval.lower(),
```
kranar, line 95: Use ```tuple``` unless you need something specific about ```pair```.

Applications/Spire/Source/Charting/CachedChartModel.cpp
```diff
@@ -1,141 +1,277 @@
...
+        std::min(interval.upper(), cached_interval.upper()));
+    } else {
+      return std::make_pair(std::max(interval.lower(),
+        cached_interval.lower()), interval.upper());
+    }
+  }();
+  return m_cache.load(start, end, limit).then(
+    [=, start = start, end = end] (auto result) {
```
kranar, line 103: You can remove the ```start = start```, ```end = end```.

Applications/Spire/Source/Charting/CachedChartModel.cpp
```diff
@@ -1,141 +1,277 @@
...
+      auto candlesticks = result.Get();
```
kranar, line 104: Use ```move```.

Applications/Spire/Source/Charting/CachedChartModel.cpp
```diff
@@ -1,141 +1,277 @@
...
+      if((is_head && end == interval.upper()) || (
+          !is_head && start == interval.lower()) ||
+          candlesticks.size() == limit.GetSize()) {
+        return QtPromise(candlesticks);
```
kranar, line 108: Use ```move```.

Applications/Spire/Source/Charting/CachedChartModel.cpp
```diff
@@ -1,141 +1,277 @@
...
+            return limit;
+          } else {
+            return SnapshotLimit(limit.GetType(), limit.GetSize() +
+              boundary_candlesticks_count);
           }
-        }
+        }();
+        return load(new_start, new_end, new_limit).then(
+          [=, candlesticks = std::move(candlesticks)] (auto result) mutable {
+            auto new_candlesticks = result.Get();
```
kranar, line 134: Use ```move```.

Applications/Spire/Source/Charting/CachedChartModel.cpp
```diff
@@ -1,141 +1,277 @@
...
-        last_iterator = std::lower_bound(data.begin(), data.end(),
-          range.lower(),
-          [] (const auto& index, const auto& value) {
-            return index.GetEnd() <= value;
-          });
-        if(data.size() > 1 && first_iterator < last_iterator) {
-          --last_iterator;
-        }
+    }
+    auto futures = std::move(m_queried_intervals[start].m_futures);
```
kranar, line 205: Use ```m_queried_intervals.at(start)```.

Applications/Spire/Source/Charting/CachedChartModel.cpp
```diff
@@ -1,141 +1,277 @@
...
+    m_queried_intervals.erase(start);
+    for(auto& future : futures) {
+      future.resolve();
+    }
+    if(candlesticks.size() == limit_size) {
+      return QtPromise([candlesticks = std::move(candlesticks)] {
```
kranar, line 211: return QtPromise(std::move(candlesticks));

Applications/Spire/Source/Charting/CachedChartModel.cpp
```diff
@@ -1,141 +1,277 @@
...
+    CachedChartModel::find_leftmost_cached_overlap(
+    const Interval& interval) const {
+  auto it = m_cached_intervals.lower_bound(interval);
+  if(it == m_cached_intervals.end()) {
+    return boost::none;
+  } else {
+    return *it;
+  }
+}
+boost::optional<CachedChartModel::Interval>
```
kranar, line 230: Add vertical whitespace before 230.

Applications/Spire/Source/Charting/CachedChartModel.cpp
```diff
@@ -1,141 +1,277 @@
...
-  } else {
-    auto range = continuous_interval<Scalar>::left_open(
-      std::max(data.front().GetStart(), info.m_first), info.m_last);
-    on_data_loaded(data, range);
+boost::optional<CachedChartModel::Interval>
+    CachedChartModel::find_leftmost_queried_overlap(
+    const Interval& interval) const {
+  auto it = m_queried_intervals.lower_bound(interval.lower());
+  if(it != m_queried_intervals.begin()) {
+    auto prev = it;
```
kranar, line 250: ```auto prev = std::prev(it)```.

Applications/Spire/Source/Charting/CachedChartModel.cpp
andrei-sa: I did it to work around the intellisense bug for capturing structured bindings. However, I've just confirmed it and MSBuild seems to run fine without it, just the ugly intellisense errors when running VS.

kranar: Intellisense sucks, doesn't even use the same compiler front end as MSVC.

kranar:
Resubmit the PR when testing of the formal properties of the cache have been added. As a general principle I'd even advise you to write the tests before writing the implementation.

PR #452, opened 2020-05-28
Cached chart model

andrei-sa:
I tried to formalize the "laws" of the cache a bit more rigorously, especially in terms of interval openness, and ensure that the tests provide enough coverage for those. Please find the rules below and how they are addressed:

_1. Any load C.load(A, B, L) cannot result in any M.load(A', B', L') such as A' < A || B' > B || L' > L._
This law is tested implicitly by comparing outputs of the reference model and the cached model filled with the same data.

2. _Let us define a loaded interval for R = M.load(A, B, L). If len(R) < L, the loaded interval is [A, B]. Otherwise_:
- _L.type == HEAD: if A == R.back().start, the loaded interval is an empty set. Otherwise, it is [A, R.back().start())_
- _L.type == TAIL: if B == R.front().start, the loaded interval is an empty set. Otherwise, it is (R.front().start, B]_

_There cannot be any M.load(A', B', L) such as [A', B'] overlaps with any loaded interval._
This law is tested using the new `TestLocalChartModel` that exposes the ranges of performed loads inside `CachedChartModel`. `TestLocalChartModel` is different from `TestChartModel` in that it uses a real `LocalChartModel` for storage, streamlining the process for serial scenarios.

_3. If an ongoing query Q1 = M.load(A, B, L) is followed with a concurrent query Q2 = C.load(A', B', L'), Q2 will only have to wait until Q1 finishes if [A, B] and [A', B'] overlap (i.e. A' <= B && B' >= A) and, if A' < A, only after loading from range [A', A]._
This law is tested using `TestChartModel`, which allows us to simulate concurrent queries. I have added a few more scenarios for concurrent loads, with helper functions added to prove the absence of waiting by testing every permutation.

Let me know if I am missing something.


Applications/Spire/Source/Charting/CachedChartModel.cpp
```diff
@@ -1,141 +1,273 @@
...
+                    new_candlesticks.size()));
+              } else {
+                return std::tuple(&new_candlesticks, &candlesticks,
+                  std::min(boundary_candlesticks_count,
+                    candlesticks.size()));
+              }
+            }();
+            std::move(tail->begin() + offset, tail->end(),
+              std::back_inserter(*head));
+            return *head;
```
kranar, line 146: ```return std::move(*head);```

PR #455, opened 2020-06-03
Page not found page

spiresheep:
Created the page not found page and a demo for it. Added it to the main app.


Applications/WebPortal/WebApp/library/source/pages/page_not_found_page.tsx
```diff
@@ -0,0 +1,89 @@
...
+  displaySize: DisplaySize;
+}
+
+/* Displays a page that indicates the desired page cannot be found. */
+export class PageNotFoundPage extends React.Component<Properties> {
+
+  public render(): JSX.Element {
+    const headerTextStyle = (() => {
+      if(this.props.displaySize === DisplaySize.LARGE) {
+        return PageNotFoundPage.STYLE.headerTextLarge;
```
kranar, line 19: Use the new "array" like style you used in a previous pull request. ```PageNotFoundPage.Style[this.props.displaySize]```. You can then inline the style into the component instead of initializing it up here.

PR #457, opened 2020-06-05
Fixed where the test was populated.

spiresheep:
This is for the profile page tester error task.


Applications/WebPortal/WebApp/tests/profile_page_tester/source/index.tsx
```diff
@@ -103,6 +102,10 @@ class TestApp extends React.Component<Properties, State> {
       </Dali.VBoxLayout>);
   }
 
+  public componentDidMount() {
```
kranar, line 105: Mark as void, ```componentDidMount(): void```.

PR #464, opened 2020-06-24
Color dropdown

spireken:
There's 3 demos called ColorSelectorButton-XYZ, where XYZ is the window name.


Applications/Spire/Include/Spire/Ui/RecentColors.hpp
```diff
@@ -0,0 +1,44 @@
...
+      //! Returns true if two RecentColors share the same recently selected
+      //! colors in the same order.
+      bool operator ==(const RecentColors& rhs) const;
+
+      //! Returns false if two RecentColors do not share the same recently
+      //! selected colors in the same order.
+      bool operator !=(const RecentColors& rhs) const;
+
+    private:
+      std::deque<QColor> m_recent_colors;
```
andrei-sa, line 40: Use a combination of a hash set (probably `QSet`) and `std::list` instead. Your `add_color` should then be constant time.
It is more of an exercise than an actual performance consideration with the container size of 9, but it communicates the intention more clearly once you are used to such algorithms.

Applications/Spire/Include/Spire/Ui/RecentColors.hpp
```diff
@@ -0,0 +1,44 @@
...
+#define SPIRE_RECENT_COLORS_HPP
+#include <array>
+#include <deque>
+#include <QColor>
+
+namespace Spire {
+
+  //! Represents a data structure for storing recently selected colors in a
+  //  FIFO manner.
+  class RecentColors {
```
andrei-sa, line 11: Add default move constructor and assignment. Then, you can use it for copy-and-move everywhere an object of `RecentColors` is acquired.

Applications/Spire/Source/BookView/BookViewHighlightPropertiesWidget.cpp
```diff
@@ -133,12 +134,13 @@ BookViewHighlightPropertiesWidget::BookViewHighlightPropertiesWidget(
...
+    recent_colors, this);
   m_market_highlight_color_button->setFixedWidth(scale_width(100));
-  update_color_button_stylesheet(m_market_highlight_color_button,
-    Qt::yellow);
-  m_market_highlight_color_button->connect_clicked_signal(
-    [=] { on_market_highlight_color_button_clicked(); });
+  m_market_highlight_color_button->connect_color_signal(
+    [=] (auto color) { on_market_highlight_color_selected(color); });
+  m_market_highlight_color_button->connect_recent_colors_signal(
+    [=] (auto recent_colors) { on_recent_colors_changed(recent_colors); });
```
andrei-sa, line 143: Either take `recent_colors` by reference, or use copy and move where possible.

Applications/Spire/Source/BookView/BookViewHighlightPropertiesWidget.cpp
```diff
@@ -182,12 +184,11 @@ BookViewHighlightPropertiesWidget::BookViewHighlightPropertiesWidget(
...
-  m_order_highlight_color_button = new FlatButton(this);
+  m_order_highlight_color_button = new ColorSelectorButton(
+    properties.get_order_highlight_color(), recent_colors, this);
   m_order_highlight_color_button->setFixedWidth(scale_width(100));
-  update_color_button_stylesheet(m_order_highlight_color_button,
-    properties.get_order_highlight_color());
-  m_order_highlight_color_button->connect_clicked_signal(
-    [=] { on_order_highlight_color_button_clicked(); });
+  m_order_highlight_color_button->connect_recent_colors_signal(
+    [=] (auto recent_colors) { on_recent_colors_changed(recent_colors); });
```
andrei-sa, line 191: Either take `recent_colors` by reference, or use copy and move where possible.

Applications/Spire/Source/BookView/BookViewLevelPropertiesWidget.cpp
```diff
@@ -127,31 +129,32 @@ BookViewLevelPropertiesWidget::BookViewLevelPropertiesWidget(
   band_color_label->setStyleSheet(generic_label_style);
   band_properties_layout->addWidget(band_color_label);
   band_properties_layout->addStretch(4);
-  m_band_color_button = new FlatButton(this);
+  m_band_color_button = new ColorSelectorButton(bg_colors[0], recent_colors,
+    this);
+  m_band_color_button->connect_color_signal(
+    [=] (auto color) { on_band_color_selected(color); });
+  m_band_color_button->connect_recent_colors_signal(
+    [=] (auto recent_colors) { on_recent_colors_changed(recent_colors); });
```
andrei-sa, line 137: Either take `recent_colors` by reference, or use copy and move where possible.

Applications/Spire/Source/BookView/BookViewLevelPropertiesWidget.cpp
```diff
@@ -127,31 +129,32 @@ BookViewLevelPropertiesWidget::BookViewLevelPropertiesWidget(
...
   band_properties_layout->addStretch(4);
-  m_gradient_start_button = new FlatButton(this);
+  m_gradient_start_button = new ColorSelectorButton(bg_colors[0],
+    recent_colors, this);
   m_gradient_start_button->setFixedHeight(scale_height(26));
-  set_color_button_stylesheet(m_gradient_start_button, bg_colors[0]);
-  m_gradient_start_button->connect_clicked_signal(
-    [=] { on_gradient_start_button_clicked(); });
+  m_gradient_start_button->connect_recent_colors_signal(
+    [=] (auto recent_colors) { on_recent_colors_changed(recent_colors); });
```
andrei-sa, line 150: Either take `recent_colors` by reference, or use copy and move where possible.

Applications/Spire/Source/BookView/BookViewLevelPropertiesWidget.cpp
```diff
@@ -127,31 +129,32 @@ BookViewLevelPropertiesWidget::BookViewLevelPropertiesWidget(
...
-  m_gradient_end_button = new FlatButton(this);
+  m_gradient_end_button = new ColorSelectorButton(
+    bg_colors[bg_colors.size() - 1], recent_colors, this);
   m_gradient_end_button->setFixedHeight(scale_height(26));
-  set_color_button_stylesheet(m_gradient_end_button,
-    bg_colors[bg_colors.size() - 1]);
-  m_gradient_end_button->connect_clicked_signal(
-    [=] { on_gradient_end_button_clicked(); });
+  m_gradient_end_button->connect_recent_colors_signal(
+    [=] (auto recent_colors) { on_recent_colors_changed(recent_colors); });
```
andrei-sa, line 157: Either take `recent_colors` by reference, or use copy and move where possible.

Applications/Spire/Source/BookView/BookViewPropertiesDialog.cpp
```diff
@@ -67,10 +67,14 @@ BookViewPropertiesDialog::BookViewPropertiesDialog(
     })").arg(scale_height(12)).arg(scale_height(20)).arg(scale_height(10))
         .arg(scale_width(2)).arg(scale_width(80)).arg(scale_width(1)));
   m_levels_tab_widget = new BookViewLevelPropertiesWidget(properties,
-    m_tab_widget);
+    recent_colors, m_tab_widget);
+  m_levels_tab_widget->connect_recent_colors_signal(
+    [=] (auto recent_colors) { on_recent_colors_changed(recent_colors); });
```
andrei-sa, line 72: Either take `recent_colors` by reference, or use copy and move where possible.

Applications/Spire/Source/BookView/BookViewPropertiesDialog.cpp
```diff
@@ -67,10 +67,14 @@ BookViewPropertiesDialog::BookViewPropertiesDialog(
...
   m_tab_widget->addTab(m_levels_tab_widget, tr("Price Levels"));
   m_highlights_tab_widget = new BookViewHighlightPropertiesWidget(
-    properties, m_tab_widget);
+    properties, recent_colors, m_tab_widget);
+  m_highlights_tab_widget->connect_recent_colors_signal(
+    [=] (auto recent_colors) { on_recent_colors_changed(recent_colors); });
```
andrei-sa, line 77: Either take `recent_colors` by reference, or use copy and move where possible.

Applications/Spire/Source/Charting/ChartingWindow.cpp
```diff
@@ -168,10 +168,13 @@ void ChartingWindow::set_models(std::shared_ptr<ChartModel> chart_model,
...
   m_draw_line_button->setEnabled(true);
   m_draw_line_button->set_toggled(false);
-  m_trend_line_editor_widget = new TrendLineEditor(m_technicals_panel);
+  m_trend_line_editor_widget = new TrendLineEditor(m_recent_colors,
+    m_technicals_panel);
   m_trend_line_editor_widget->hide();
   m_trend_line_editor_widget->connect_color_signal(
     [=] { on_trend_line_color_selected(); });
+  m_trend_line_editor_widget->connect_recent_colors_signal(
+    [=] (auto recent_colors) { on_recent_colors_changed(recent_colors); });
```
andrei-sa, line 177: Either take `recent_colors` by reference, or use copy and move where possible.

Applications/Spire/Source/BookView/BookViewWindow.cpp
```diff
@@ -132,9 +139,12 @@ void BookViewWindow::show_context_menu(const QPoint& pos) {
 }
 
 void BookViewWindow::show_properties_dialog() {
-  BookViewPropertiesDialog dialog(get_properties(), Security(), this);
+  BookViewPropertiesDialog dialog(get_properties(), Security(),
+    m_recent_colors, this);
   m_dialog_apply_connection = dialog.connect_apply_signal(
     [&] { set_properties(dialog.get_properties()); });
+  m_dialog_recent_colors_connection = dialog.connect_recent_colors_signal(
+    [&] (auto recent_colors) { m_recent_colors_signal(recent_colors); });
```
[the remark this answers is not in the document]
andrei-sa: Either take `recent_colors` by reference, or use copy and move where possible.

Applications/Spire/Source/Charting/TrendLineEditor.cpp
```diff
@@ -27,11 +28,14 @@ TrendLineEditor::TrendLineEditor(QWidget* parent)
...
-  color_picker->connect_color_signal(
-    [=] (auto& color) { on_color_change(color); });
-  color_picker->setFixedSize(scale(70, 18));
-  layout->addWidget(color_picker);
+  auto color_button = new ColorSelectorButton(QColor("#FFCA19"),
+    recent_colors, this);
+  color_button->connect_color_signal(
+    [=] (auto color) { on_color_change(color); });
+  color_button->connect_recent_colors_signal(
+    [=] (auto recent_colors) { on_recent_colors_changed(recent_colors); });
```
andrei-sa, line 36: Either take `recent_colors` by reference, or use copy and move where possible.

Applications/Spire/Source/TimeAndSales/TimeAndSalesPropertiesDialog.cpp
```diff
@@ -95,12 +97,12 @@ TimeAndSalesPropertiesDialog::TimeAndSalesPropertiesDialog(
...
-    [=] { set_text_color(); });
-  set_color_button_stylesheet(m_text_color_button,
-    properties.get_text_color(
-    PriceRange::UNKNOWN));
+  m_text_color_button = new ColorSelectorButton(properties.get_text_color(
+    PriceRange::UNKNOWN), m_recent_colors, this);
+  m_text_color_button->connect_color_signal(
+    [=] (auto color) { set_text_color(color); });
+  m_text_color_button->connect_recent_colors_signal(
+    [=] (auto recent_colors) { on_recent_colors_changed(recent_colors); });
```
andrei-sa, line 105: Either take `recent_colors` by reference, or use copy and move where possible.

Applications/Spire/Source/TimeAndSales/TimeAndSalesPropertiesDialog.cpp
```diff
@@ -109,11 +111,12 @@ TimeAndSalesPropertiesDialog::TimeAndSalesPropertiesDialog(
...
-  m_band_color_button->connect_clicked_signal(
-    [=] { set_band_color(); });
-  set_color_button_stylesheet(m_band_color_button, properties.get_band_color(
-    PriceRange::UNKNOWN));
+  m_band_color_button = new ColorSelectorButton(properties.get_band_color(
+    PriceRange::UNKNOWN), m_recent_colors, this);
+  m_band_color_button->connect_color_signal(
+    [=] (auto color) { set_band_color(color); });
+  m_band_color_button->connect_recent_colors_signal(
+    [=] (auto recent_colors) { on_recent_colors_changed(recent_colors); });
```
andrei-sa, line 119: Either take `recent_colors` by reference, or use copy and move where possible.

Applications/Spire/Source/TimeAndSales/TimeAndSalesWindow.cpp
```diff
@@ -163,15 +170,24 @@ void TimeAndSalesWindow::export_table() {
 }
 
 void TimeAndSalesWindow::show_properties_dialog() {
-  TimeAndSalesPropertiesDialog dialog(m_properties, this);
+  TimeAndSalesPropertiesDialog dialog(m_properties, m_recent_colors, this);
   dialog.connect_apply_signal([&] {set_properties(dialog.get_properties()); });
+  dialog.connect_recent_colors_signal([=] (auto recent_colors) {
```
andrei-sa, line 175: Either take `recent_colors` by reference, or use copy and move where possible.

Applications/Spire/Source/Ui/ColorSelectorHueSlider.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include "Spire/Spire/Dimensions.hpp"
+#include "Spire/Spire/Utility.hpp"
+#include "Spire/Ui/Ui.hpp"
+
+using namespace boost::signals2;
+using namespace Spire;
+
+namespace {
+  auto HANDLE_Y() {
+    static auto y = scale_height(3);
```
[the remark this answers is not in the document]
spireken: For this slider the handle only moves along the x axis.

Applications/Spire/Include/Spire/Ui/RecentColors.hpp
spireken: Wouldn't a set lose the ordering?

andrei-sa: It certainly will. The idea is that in your algorithm, you perform the following four operations:

1. Check for existence of an element and locate it if it does exist.
2. Remove the element located if any.
3. Pop the first element.
4. Push a new element to the end of the container.

`std::deque` allows you to achieve O(1) for (3) and (4), but both (1) and (2) are linear as it behaves largely as a normal array.
Now, if you simply replace `std::deque` with `std::list`, you are able to achieve O(1) for (2), while keeping the constant complexity of (3) and (4). The complexity of (1) remains linear though, as you have to traverse the entire list to look for the element. As a single `add_color` operations needs to at least perform (1), one would argue that the overall complexity is still linear.

Finally, you are able to add a supplementary structure to make (1) constant time as well. The thinking is: for fast existence/absence tests, as well for the fast location elements, we use hash tables. If you maintain a table from element keys (`QColor`) to list iterators, you are then able to locate an existing element in constant time, and then remove it from the list (in constant time as well). In other words, the ordering is maintained by the list, while the hash serves as an index for that list.

To correct myself, you probably need `QHash` (or `std::unordered_map`, if `std::hash` is implemented for `QColor`) instead of `QSet` to store list iterators as satellite data.

As mentioned before, you are unlikely to benefit from constant complexity in this case, yet this is a repetitive pattern and I have come across similar sub-optimal implementation. The main goal is to understand how you can benefit from using other data structures in situations like this, and to avoid poor implementations in less trivial scenarios.

kranar: Keep the implementation as is. Increasing the complexity by adding a hash and a list in order to keep track of 9 colors is simply overkill.

kranar: If any change is to be made it would be to simplify it even further and just use an ```std::array```.

Applications/Spire/Source/BookView/BookViewWindow.cpp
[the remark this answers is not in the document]
spireken: The exec() function is supposed to block until the dialog is closed.

Applications/Spire/Include/Spire/Ui/RecentColors.hpp
spireken: I don't understand how point (2) is solved with the iterator hashes; wouldn't it have to be 'refreshed' every time the list is modified?

andrei-sa: (2) is solved with the use of a linked list rather than the hash. It pays to understand why pointer removals from linked lists are constant.
List modifications do not invalidate existing iterators (except for removed elements). If you have an iterator to, say, 4th element in a list and then remove the 2nd element, the iterator to the 4th element is unchanged. Similarly to the case of node removal, it is a good idea to understand why the iterators stay the same, as well as how arrays and linked lists are different at all.

Applications/Spire/Include/Spire/Spire/Spire.hpp
```diff
@@ -17,11 +15,6 @@ namespace Spire {
   class SpireController;
   template<typename T> struct SubscriptionResult;
   class TechnicalsModel;
-
-  //! Defines the common type of boost signal used throughout Spire.
-  template<typename F>
```
andrei-sa, line 22: Include Singal.hpp in Spire.hpp for compatibility.

Applications/Spire/Include/Spire/BookView/BookViewHighlightPropertiesWidget.hpp
```diff
@@ -3,6 +3,7 @@
 #include <QListWidget>
 #include <QWidget>
 #include "Spire/BookView/BookView.hpp"
+#include "Spire/Spire/Spire.hpp"
```
andrei-sa, line 6: Only include Signal.hpp unless you otherwise need Spire.hpp. Applies to a few more occurrences below.

Applications/Spire/Include/Spire/BookView/BookViewLevelPropertiesWidget.hpp
```diff
@@ -3,6 +3,7 @@
 #include <QListWidget>
 #include <QWidget>
 #include "Spire/BookView/BookView.hpp"
+#include "spire/Spire/Spire.hpp"
```
andrei-sa, line 6: Fix case.

Applications/Spire/Source/Ui/RecentColors.cpp
```diff
@@ -0,0 +1,50 @@
...
+  m_recent_colors.push_front(color);
+  if(m_recent_colors.size() > RECENT_COLOR_COUNT) {
+    m_recent_colors.pop_back();
+  }
+}
+
+std::array<QColor, RecentColors::RECENT_COLOR_COUNT>
+    RecentColors::get_colors() const {
+  auto colors = std::array<QColor, RECENT_COLOR_COUNT>();
+  for(auto iter = m_recent_colors.begin(); iter != m_recent_colors.end();
```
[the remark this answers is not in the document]
andrei-sa: This is market resolved but I don't see any update yet. Let me know if you intend to push more commits.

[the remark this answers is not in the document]
spireken: I think I made the change but then undid it after the other notes left by Kamal. I'll make this change then push another commit.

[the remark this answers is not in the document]
spireken: What's the purpose of the counter versus doing this:

`  std::transform(m_recent_colors.begin(), m_recent_colors.end(),
    colors.begin(), [] (const auto& color) { return color; });`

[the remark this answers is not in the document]
andrei-sa: You don't need a counter, disregard that. You could use a counter if you wanted to use a range-for loop instead, which is also viable. The thing with `std::distance` is a bit too indirect though.

Applications/Spire/Include/Spire/BookView/BookViewHighlightPropertiesWidget.hpp
```diff
@@ -11,43 +14,61 @@ namespace Spire {
...
       /*!
         \param properties The instance to apply the properties to.
       */
       void apply(BookViewProperties& properties) const;
 
+      //! Sets the current recent colors.
+      /*!
+        \param recent_colors The current recent colors.
+      */
+      void set_recent_colors(const RecentColors& recent_colors);
```
kranar, line 44: I don't see the purpose of this method. The get/set recent colors is a property of the ColorSelectorButton, not its parent. Almost all of these ```set_recent_colors``` methods need to be removed.

As a general principle, never change an API without getting approval first, I am open to hearing a good argument for why these should stay here but it's almost certainly going to need to be removed.

Applications/Spire/Include/Spire/Ui/RecentColors.hpp
```diff
@@ -0,0 +1,44 @@
...
+    public:
+
+      //! The maximum number of recent colors to store.
+      static const int RECENT_COLOR_COUNT = 9;
```
kranar, line 15: Use ```auto``` even for ```static``` fields.

Use ```constexpr``` if you want to access that field in a constant expression, such as in ```std::array```.

The end result is:

    static constexpr auto RECENT_COLOR_COUNT = 9;

Applications/Spire/Source/Ui/ColorSelectorValueSlider.cpp
```diff
@@ -0,0 +1,116 @@
...
+  painter.setPen({QColor("#C8C8C8"), 1});
+  painter.drawRect(0, 0, width() - 1, height() - 1);
+}
+
+void ColorSelectorValueSlider::resizeEvent(QResizeEvent* event) {
+  update_gradient();
+}
+
+QPoint ColorSelectorValueSlider::get_mouse_pos(const QColor& color) const {
+  return { static_cast<int>(map_to(color.saturation(), 0, 255, 1.0,
```
kranar, line 90: Remove space after ```{```.

Applications/Spire/Source/Ui/RecentColors.cpp
```diff
@@ -0,0 +1,49 @@
...
+    colors.add_color(Qt::blue);
+    colors.add_color(Qt::green);
+    colors.add_color(Qt::red);
+    return colors;
+  }();
+  return default_colors;
+}
+
+void RecentColors::add_color(const QColor& color) {
+  auto iter = std::find_if(m_recent_colors.begin(), m_recent_colors.end(),
```
kranar, line 24: Should be able to use ```std::find```.

Applications/Spire/Source/Ui/RecentColors.cpp
```diff
@@ -0,0 +1,49 @@
...
+    [&] (const auto& element) { return element == color; });
+  if(iter != m_recent_colors.end()) {
+    m_recent_colors.erase(iter, std::min(iter + 1, m_recent_colors.end()));
```
kranar, line 27: Justify the use of ```std::min``` here.

Applications/Spire/Source/Ui/RecentColors.cpp
```diff
@@ -0,0 +1,49 @@
...
+  m_recent_colors.push_front(color);
+  if(m_recent_colors.size() > RECENT_COLOR_COUNT) {
+    m_recent_colors.pop_back();
+  }
+}
+
+std::array<QColor, RecentColors::RECENT_COLOR_COUNT>
+    RecentColors::get_colors() const {
+  auto colors = std::array<QColor, RECENT_COLOR_COUNT>();
+  std::transform(m_recent_colors.begin(), m_recent_colors.end(),
```
kranar, line 38: Use ```std::copy``` instead of ```std::transform```.

Applications/Spire/Include/Spire/BookView/BookViewHighlightPropertiesWidget.hpp
spireken: Darryl said the recent colors are shared globally, so windows will need some way to have their recent colors updated if they're modified from another window.

In this case because the Book View properties tabs are distinct widgets, I'll still need some way to update the recent colors in the other tab's widget.

kranar: They will be shared globally but implementing that is going to be much further down the road.

Applications/Spire/Source/Ui/RecentColors.cpp
spireken: I can't. If this line is hit then that means iter + 1 will be at most equal to m_recent_colors.end().

Applications/Spire/Include/Spire/BookView/BookViewHighlightPropertiesWidget.hpp
spireken: So what functionality should this retain? Are we ignoring the recent colors completely for now?

For simplicity, I can just give the color selector drop down the default colors and remove the signals and setters for them.

kranar: RecentColors has a public static method: ```RecentColors& get_instance();```

It should also have signals indicating when it's been updated.

spireken: So in effect, every widget that uses RecentColors uses them independently?

spireken:
The 3 demos have been updated.

Applications/Spire/Include/Spire/Ui/RecentColors.hpp
```diff
@@ -0,0 +1,49 @@
...
+      void add_color(const QColor& color);
+
+      //! Returns the collection of recently selected colors.
+      std::array<QColor, RECENT_COLOR_COUNT> get_colors() const;
+
+      //! Connects a slot to the change signal.
+      boost::signals2::connection connect_change_signal(
+        const ChangeSignal::slot_type& slot) const;
+
+      RecentColors(const RecentColors&) = delete;
```
kranar, line 37: If deleting a constructor, declare them in the private section.

Applications/Spire/Include/Spire/Ui/RecentColors.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+      RecentColors& operator =(const RecentColors&) = delete;
+
+    private:
+      RecentColors();
```
kranar, line 42: Constructors/methods are declared after fields.

Applications/Spire/Source/KeyBindings/NameItemDelegate.cpp
```diff
@@ -3,6 +3,7 @@
 #include <QPainter>
 #include "Spire/KeyBindings/NameInputEditor.hpp"
 #include "Spire/Spire/Dimensions.hpp"
+#include "Spire/Spire/Signal.hpp"
```
kranar, line 6: Where is this include used?

Applications/Spire/Source/KeyBindings/KeySequenceItemDelegate.cpp
```diff
@@ -3,6 +3,7 @@
 #include <QKeyEvent>
 #include <QPainter>
 #include "Spire/Spire/Dimensions.hpp"
+#include "Spire/Spire/Signal.hpp"
```
kranar, line 6: Similarly, where is this include used?

Applications/Spire/Source/Ui/ColorSelectorDropDown.cpp
```diff
@@ -0,0 +1,243 @@
...
+    const ColorSignal::slot_type& slot) const {
+  return m_color_signal.connect(slot);
+}
+
+void ColorSelectorDropDown::childEvent(QChildEvent* event) {
+  if(event->added()) {
+    event->child()->installEventFilter(this);
+  }
+}
+bool ColorSelectorDropDown::eventFilter(QObject* watched, QEvent* event) {
```
kranar, line 183: Vertical whitespace needed.

Applications/Spire/Source/Ui/ColorSelectorDropDown.cpp
```diff
@@ -0,0 +1,243 @@
...
+    m_recent_colors.add_color(m_current_color);
+    m_original_color = m_current_color;
+    update_recent_colors_layout();
+  }
+}
+
+void ColorSelectorDropDown::add_basic_color_button(int x, int y,
+    const QColor& color) {
+  auto button = create_color_button(color, this);
+  button->connect_clicked_signal([=, color = std::move(color)] {
```
kranar, line 204: Don't need to explicitly capture  ```color```. It doesn't gain anything.

Applications/Spire/Source/Ui/ColorSelectorDropDown.cpp
```diff
@@ -0,0 +1,243 @@
...
+    on_color_button_clicked(color);
+  });
+  m_basic_colors_layout->addWidget(button, x, y);
+}
+
+void ColorSelectorDropDown::add_recent_color_button(int index,
+    const QColor& color) {
+  auto button = create_color_button(color, this);
+  button->connect_clicked_signal([=, color = std::move(color)] {
```
kranar, line 213: Same as above.

Applications/Spire/Source/Ui/ColorSelectorDropDown.cpp
```diff
@@ -0,0 +1,243 @@
...
+  }
+}
+
+void ColorSelectorDropDown::on_color_button_clicked(const QColor& color) {
+  set_color(color);
+  on_color_selected(color);
+}
+
+void ColorSelectorDropDown::on_color_selected(const QColor& color) {
+  m_color_signal(color);
```
kranar, line 237: This violates principles of re-entrancy. You must only signal to the outside world that a change has been made after you make the change, not before. Anyone listening to the color signal will see inconsistencies between what color was selected and the current color.

Applications/Spire/Source/Ui/RecentColors.cpp
```diff
@@ -0,0 +1,50 @@
...
+  return current_colors;
+}
+
+void RecentColors::add_color(const QColor& color) {
+  if(m_recent_colors.front() == color) {
+    return;
+  }
+  auto iter = std::find(m_recent_colors.begin(), m_recent_colors.end(), color);
+  if(iter != m_recent_colors.end()) {
+    m_recent_colors.erase(iter, iter + 1);
```
kranar, line 30: No need for ```iter + 1```.

spireken: Based on my testing, calling erase(iter, iter) doesn't erase anything.

The closest mention of this I can find is from [ccpreference](https://en.cppreference.com/w/cpp/container/deque/erase):

"The iterator first does not need to be dereferenceable if first==last: erasing an empty range is a no-op."

kranar: Good looking up the definition, on that page you will see two definitions provided (1) and (2), use definition (1).

kranar: This is not yet resolved.

kranar:
I will approve this and overall it's a good, solid, implementation but this whole change involving Spire.hpp and Signal.hpp has introduced a giant mess and turned what should have been a fairly straight forward change-set into a massive 100+ file change. I will create a separate task to revert that change entirely.

PR #465, opened 2020-06-25
PageNotFound is displayed when an account cannot be loaded.

spiresheep:
This pattern/approach can be used in plenty of other places when data cannot be fetched.


spiresheep:
Can you give me details about cases where it isn't working? It does work in the case of the path being /account/8989089 or some other id that doesn't have an account associated with it.

jiminssyspire:
Sure Ill write it up for tomorrow.

PR #470, opened 2020-07-07
Time sales revisions

spireken:
The demo is called TimeAndSalesUiTester-Revisions.


Applications/Spire/Source/TimeAndSales/TimeAndSalesTableView.cpp
```diff
@@ -85,6 +83,8 @@ TimeAndSalesTableView::TimeAndSalesTableView(QWidget* parent)
     new CustomVariantItemDelegate(), this));
   m_layout->addWidget(m_table);
   setWidget(main_widget);
+  m_minimum_column_widths = {{0, scale_width(54)}, {1, scale_width(46)},
```
andrei-sa, line 86: Any reason to not include this in the initializer list?

Applications/Spire/Source/TimeAndSales/TimeAndSalesWindowModel.cpp
```diff
@@ -106,18 +130,38 @@ QVariant TimeAndSalesWindowModel::headerData(int section,
...
         return QVariant();
     }
+  } else if(role == Qt::FontRole) {
+    return m_header_font;
   }
   return QVariant();
 }
 
+bool TimeAndSalesWindowModel::is_short_text(int column, int breakpoint) const {
+  if(m_column_size_reference.find(column) !=
```
andrei-sa, line 158: In cases like this, use one lookup instead:

```
auto it = m_column_size_reference.find(column);
if(it != m_column_size_reference.end()) {
  return it->second < breakpoint;
}
```

As an aside, the advantage of method `at` over `operator []` is that the former will throw when the element is not present, as opposed to default constructing (in case of maps) or undefined behavior (in case of lists). That said, when you have already checked for existence of the element e.g. using `find`, using `at` is excessive (in single-threaded environments at least). The presence of the exception handling in the compiled code imposes quite some overhead so, when the element is definitely present, you should prefer calling `operator []` instead.

Applications/Spire/Include/Spire/TimeAndSales/TimeAndSalesTableView.hpp
```diff
@@ -46,6 +49,7 @@ namespace Spire {
       boost::signals2::scoped_connection m_model_end_loading_connection;
       SnapshotLoadingWidget* m_loading_widget;
       TransitionWidget* m_transition_widget;
+      std::unordered_map<int, int> m_minimum_column_widths;
```
kranar, line 52: Use an array and define the columns, or use an existing definition if it's available, ie.

```static constexpr auto PRICE_COLUMN = 0;```

File: Applications/Spire/Include/Spire/TimeAndSales/TimeAndSalesWindowModel.hpp [hunk 4]
```diff
@@ -63,11 +72,14 @@ namespace Spire {
       std::shared_ptr<TimeAndSalesModel> m_model;
       boost::signals2::scoped_connection m_time_and_sale_connection;
       TimeAndSalesProperties m_properties;
+      QFont m_header_font;
       std::vector<TimeAndSalesModel::Entry> m_entries;
       bool m_is_loading;
       bool m_is_fully_loaded;
       QtPromise<std::vector<TimeAndSalesModel::Entry>> m_snapshot_promise;
+      std::unordered_map<int, int> m_column_size_reference;
```
kranar, line 80: See above.

Applications/Spire/Source/TimeAndSales/TimeAndSalesTableView.cpp
```diff
@@ -141,6 +138,17 @@ void TimeAndSalesTableView::resizeEvent(QResizeEvent* event) {
   m_header->setFixedWidth(widget()->width());
 }
 
+void TimeAndSalesTableView::showEvent(QShowEvent* event) {
+  if(!m_was_shown && m_model != nullptr) {
+    m_was_shown = true;
+    m_header->resizeSection(0, scale_width(74));
```
kranar, line 144: Same principle here, this would benefit a lot more from names instead of magic values.

Applications/Spire/Source/TimeAndSales/TimeAndSalesWindowModel.cpp
```diff
@@ -12,15 +14,27 @@ using PriceRange = TimeAndSalesProperties::PriceRange;
...
   const auto SNAPSHOT_COUNT = 4 * LOADING_THRESHOLD;
+
+  auto BREAKPOINT_PADDING() {
+    static auto padding = scale_width(8);
+    return padding;
+  }
+
+  auto BREAKPOINT(const QFont& font, const QString& text) {
+    auto width = QFontMetrics(font).horizontalAdvance(text);
+    return width + BREAKPOINT_PADDING() * 2;
```
kranar, line 25: Constants first, common math convention that you write ```2x``` instead of ```x2```

Applications/Spire/Source/TimeAndSales/TimeAndSalesWindowModel.cpp
```diff
@@ -12,15 +14,27 @@ using PriceRange = TimeAndSalesProperties::PriceRange;
 namespace {
   const auto LOADING_THRESHOLD = std::size_t{25};
   const auto SNAPSHOT_COUNT = 4 * LOADING_THRESHOLD;
+
+  auto BREAKPOINT_PADDING() {
+    static auto padding = scale_width(8);
+    return padding;
+  }
+
+  auto BREAKPOINT(const QFont& font, const QString& text) {
```
kranar, line 23: If this is a function then name it ```get_breakpoint(...);```

spireken:
I can't find the code that your std::array comment is referencing: it's not linked anywhere.

kranar:
Ignore it.

Applications/Spire/Source/TimeAndSales/TimeAndSalesTableView.cpp
```diff
@@ -13,6 +13,7 @@
 #include "Spire/Ui/TransitionWidget.hpp"
 
 using namespace Spire;
+using Columns = TimeAndSalesProperties::Columns;
```
kranar, line 16: This should be renamed from ```Columns``` to ```Column``` and ```TimeAndSalesProperties::Columns``` should be renamed to ```TimeAndSalesProperties::Column```.

A type should be named based on a single instance. If an instance represents a single column, then the name should be ```Column```. If an instance represents multiple columns, only then would it make sense to name it ```Columns```.

PR #473, opened 2020-07-09
Implement TagModel

andrei-sa:
For primary tag keys, I temporarily hardcoded them in different places because apparently the ID - display name mapping should be set somewhere in the UI.

### DIFF
--- a/Applications/Spire/Include/Spire/KeyBindings/ConjunctionTerm.hpp
+++ b/Applications/Spire/Include/Spire/KeyBindings/ConjunctionTerm.hpp
@@ -17,8 +17,8 @@ namespace Spire {
       */
       explicit ConjunctionTerm(std::vector<std::shared_ptr<Term>> terms);
 
-      virtual bool is_satisfied(const std::vector<KeyBindings::Tag>&
-        tags) const override;
+      bool is_satisfied(const std::vector<KeyBindings::Tag>& tags) const
+        override;
 
     private:
       std::vector<std::shared_ptr<Term>> m_terms;
--- a/Applications/Spire/Include/Spire/KeyBindings/ConjunctionTerm.hpp
+++ b/Applications/Spire/Include/Spire/KeyBindings/ConjunctionTerm.hpp
@@ -7,16 +7,13 @@
 
 namespace Spire {
 
-  /*! \class ConjunctionTerm
-      \brief Evaluates to true iff all given terms evaluate to true.
-  */
+  //! Evaluates to true iff there is no term in a list that evaluates to false.
   class ConjunctionTerm : public Term {
     public:
 
       //! Constructs a ConjunctionTerm.
       /*!
-        \param terms The terms to evaluate for tags. Should contain at least
-                     1 term.
+        \param terms The terms to evaluate for tags.
       */
       explicit ConjunctionTerm(std::vector<std::shared_ptr<Term>> terms);
 
--- a/Applications/Spire/Include/Spire/KeyBindings/ConjunctionTerm.hpp
+++ b/Applications/Spire/Include/Spire/KeyBindings/ConjunctionTerm.hpp
@@ -7,7 +7,7 @@
 
 namespace Spire {
 
-  //! Evaluates to true iff there is no term in a list that evaluates to false.
+  //! Is satisfied iff there is no term in a list that evaluates to false.
   class ConjunctionTerm : public Term {
     public:
 
--- a/Applications/Spire/Include/Spire/KeyBindings/ConjunctionTerm.hpp
+++ b/Applications/Spire/Include/Spire/KeyBindings/ConjunctionTerm.hpp
@@ -1,0 +1,28 @@
+#ifndef SPIRE_CONJUNCTION_TERM_HPP
+#define SPIRE_CONJUNCTION_TERM_HPP
+#include <memory>
+#include <vector>
+#include "Spire/KeyBindings/KeyBindings.hpp"
+#include "Spire/KeyBindings/Term.hpp"
+
+namespace Spire {
+
+  /*! \class ConjunctionTerm
+      \brief Evaluates to true iff all given terms evaluate to true.
+  */
+  class ConjunctionTerm : public Term {
+    public:
+
+      //! Constructs a ConjunctionTerm.
+      /*!
+        \param terms The terms to evaluate for tags. Should contain at least
+                     1 term.
+      */
+      explicit ConjunctionTerm(std::vector<std::shared_ptr<Term>> terms);
+
+      virtual bool is_satisfied(const std::vector<KeyBindings::Tag>&
+        tags) const override;
+  };
+}
+
+#endif
--- a/Applications/Spire/Include/Spire/KeyBindings/ConjunctionTerm.hpp
+++ b/Applications/Spire/Include/Spire/KeyBindings/ConjunctionTerm.hpp
@@ -19,6 +19,9 @@ namespace Spire {
 
       virtual bool is_satisfied(const std::vector<KeyBindings::Tag>&
         tags) const override;
+
+    private:
+      std::vector<std::shared_ptr<Term>> m_terms;
   };
 }
 
--- a/Applications/Spire/Include/Spire/KeyBindings/KeyBindings.hpp
+++ b/Applications/Spire/Include/Spire/KeyBindings/KeyBindings.hpp
@@ -42,24 +42,48 @@ namespace Details {
       class Tag {
         public:
 
-          //! The type of a variant used to store all possible tag values.
+          //! The variant of all possible types of tag values.
           using Type = typename Details::AddOptionalToVariant<
             typename Details::AddOptionalToVariant<
             typename Details::AddOptionalToVariant<
+            typename Details::AddOptionalToVariant<
             typename Details::ToVariantOfOptionals<Nexus::Tag::Type>::Type,
-            Nexus::Side>::Type, Nexus::OrderType>::Type, Nexus::Region>::Type;
+            Nexus::Side>::Type, Nexus::OrderType>::Type,
+            Nexus::Region>::Type, Nexus::TimeInForce>::Type;
 
           //! Constructs a Tag.
           /*!
             \param key The key of the tag.
             \param value The value of the tag.
           */
-          Tag(int key, Type value);
+          template<typename T>
+          Tag(int key, T value);
+
+          //! Constructs a Tag.
+          /*!
+            \param key The key of the tag.
+            \param value The value of the tag.
+          */
+          template<typename T>
+          Tag(int key, boost::optional<T> value);
 
           //! Returns the key of the tag.
           int get_key() const;
 
           //! Returns the value of the tag.
+          /*!
+            \tparam T The type of tag's value.
+            \return The value of the tag, boost::none if the type of the value
+                    is not T.
+          */
+          template<typename T>
+          boost::optional<T> get_value() const;
+
+          //! Returns the value of the tag.
+          /*!
+            \return The value of the tag contained in a variant of all possible
+                    types.
+          */
           Type get_value() const;
 
           //! Checks whether two tags are equal.
@@ -67,16 +91,57 @@ namespace Details {
 
           //! Checks whether two tags are not equal.
           bool operator !=(const Tag& other) const;
+
+        private:
+          friend std::ostream& operator <<(std::ostream& out,
+            const KeyBindings::Tag& tag);
+
+          int m_key;
+          Type m_value;
       };
 
       //! Stores a description of an order submission action.
-      struct OrderAction {
+      class OrderAction {
+        public:
 
-        //! The name of the action.
-        std::string m_name;
+          //! Constructs an unnamed OrderAction with no tags.
+          OrderAction() = default;
 
-        //! The list of tags.
-        std::vector<Tag> m_tags;
+          //! Constructs an OrderAction.
+          /*!
+            \param name The name of the action.
+            \param tags The list of the tags.
+          */
+          OrderAction(std::string name, std::vector<Tag> tags);
+
+          //! Returns the name of the action.
+          std::string get_name() const;
+
+          //! Renames the action.
+          /*!
+            \param name The new name of the action.
+          */
+          void set_name(std::string name);
+
+          //! Returns the tag with the key provided.
+          /*!
+            \return The tag if present, boost::none otherwise.
+          */
+          boost::optional<Tag> get_tag(int tag_key) const;
+
+          //! Adds a new tag to the list, or updates a value of the existing
+          //! tag if one with the same key is already stored.
+          /*!
+            \param tag The tag to set.
+          */
+          void set_tag(Tag tag);
+
+          //! Returns a list of all tags.
+          std::vector<Tag> get_tags() const;
+
+        private:
+          std::string m_name;
+          std::vector<Tag> m_tags;
       };
 
       //! Represents an order cancellation action.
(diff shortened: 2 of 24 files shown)


Applications/Spire/Source/KeyBindings/KeyBindings.cpp
```diff
@@ -23,17 +79,21 @@ bool KeyBindings::CancelActionBinding::operator !=(
...
   auto region = Region::Global();
-  auto action1 = KeyBindings::OrderAction{"", OrderType::LIMIT, Side::BID, {},
-    100, {}};
+  auto action1 = KeyBindings::OrderAction{"", {Tag(40, OrderType(
+    OrderType::LIMIT)), Tag(54, Side(Side::BID)), Tag(59,
+    boost::optional<TimeInForce>()), Tag(38, 100)}};
   bindings.set(Qt::Key_F5, region, action1);
-  auto action2 = KeyBindings::OrderAction{"", OrderType::LIMIT, Side::ASK, {},
-    100, {}};
+  auto action2 = KeyBindings::OrderAction{"", {Tag(40, OrderType(
```
andrei-sa, line 86: Unfortunately, the type of something like `OrderType::LIMIT` is technically not `OrderType`, it is `OrderTypeEnumMembers::Type` due to the definition of `BEAM_ENUM`. That makes `tag.get_value<OrderType>` fail should such a value be passed directly into the variant. It seems quite unsafe because one is running a risk of misinterpreting the value of a tag, but I do not know a better way to handle it and implicit conversions in general.

Applications/Spire/Include/Spire/KeyBindings/ConjunctionTerm.hpp
```diff
@@ -0,0 +1,28 @@
+#ifndef SPIRE_CONJUNCTION_TERM_HPP
+#define SPIRE_CONJUNCTION_TERM_HPP
+#include <memory>
+#include <vector>
+#include "Spire/KeyBindings/KeyBindings.hpp"
+#include "Spire/KeyBindings/Term.hpp"
+
+namespace Spire {
+
+  //! Evaluates to true iff there is no term in a list that evaluates to false.
```
kranar, line 10: Replace the "Evaluates to ..." with "Is satisfied iff ..."

Applications/Spire/Include/Spire/KeyBindings/ConjunctionTerm.hpp
```diff
@@ -0,0 +1,28 @@
...
+  class ConjunctionTerm : public Term {
+    public:
+
+      //! Constructs a ConjunctionTerm.
+      /*!
+        \param terms The terms to evaluate for tags.
+      */
+      explicit ConjunctionTerm(std::vector<std::shared_ptr<Term>> terms);
+
+      virtual bool is_satisfied(const std::vector<KeyBindings::Tag>&
```
kranar, line 20: Remove ```virtual```, we only mark as ```virtual``` in the base class and all other classes mark as ```override```.

Applications/Spire/Include/Spire/KeyBindings/KeyBindings.hpp
```diff
@@ -24,58 +24,127 @@ namespace Details {
...
+
+          //! Constructs an OrderAction.
+          /*!
+            \param name The name of the action.
+            \param tags The list of the tags.
+          */
+          OrderAction(std::string name, std::vector<Tag> tags);
+
+          //! Returns the name of the action.
+          std::string get_name() const;
```
kranar, line 118: Return ```const &```.

Applications/Spire/Include/Spire/KeyBindings/KeyBindings.hpp
```diff
@@ -24,58 +24,127 @@ namespace Details {
...
+
+          //! Adds a new tag to the list, or updates a value of the existing
+          //! tag if one with the same key is already stored.
+          /*!
+            \param tag The tag to set.
+          */
+          void set_tag(Tag tag);
+
+          //! Returns a list of all tags.
+          std::vector<Tag> get_tags() const;
```
kranar, line 140: Return ```const &```

Applications/Spire/Include/Spire/KeyBindings/KeyBindings.hpp
```diff
@@ -227,11 +290,39 @@ namespace Details {
       std::vector<KeyBindingMapping> m_bindings;
   };
 
+  template<typename T>
+  KeyBindings::Tag::Tag(int key, T value)
+    : Tag(key, boost::optional<T>(std::move(value))) {}
```
kranar, line 295: Whenever possible, use deduction guides, so ```boost::optional(std::move(value))``` is fine.

Applications/Spire/Include/Spire/KeyBindings/KeyBindings.hpp
```diff
@@ -227,11 +290,39 @@ namespace Details {
...
+
+  template<typename T>
+  boost::optional<T> KeyBindings::Tag::get_value() const {
+    return std::visit(
+      [] (auto& value) {
+        if constexpr(std::is_same_v<std::decay_t<decltype(value)>,
+            boost::optional<T>>) {
+          return value;
+        } else {
+          return boost::optional<T>();
```
kranar, line 310: Don't think a silent failure works well, I'd rather an exception/abort. What do you think?

Applications/Spire/Include/Spire/KeyBindings/TagModel.hpp
```diff
@@ -0,0 +1,53 @@
...
+#define SPIRE_TAG_MODEL_HPP
+#include <memory>
+#include <string>
+#include <vector>
+#include "Spire/KeyBindings/KeyBindings.hpp"
+#include "Spire/KeyBindings/Term.hpp"
+
+namespace Spire {
+
+	//! Stores key binding tags with their requirements.
```
kranar, line 11: Fix indentation.

Applications/Spire/Include/Spire/KeyBindings/TagModel.hpp
```diff
@@ -0,0 +1,53 @@
...
+      //! Returns a list of tags enabled for a set of prior tags.
+      /*!
+        \param prior_tags A list of already set tags.
+      */
+      std::vector<KeyBindings::Tag> load_enabled_tags(
+        const std::vector<KeyBindings::Tag>& prior_tags) const;
+
+    private:
+      std::vector<Schema> m_schemas;
+	};
```
kranar, line 50: Indentation.

Applications/Spire/Include/Spire/KeyBindings/TagModel.hpp
```diff
@@ -0,0 +1,53 @@
...
+      /*!
+        \param schemas A list of Schemas of the tags to store.
+      */
+      explicit TagModel(std::vector<Schema> schemas);
+      
+      //! Returns the display name of the tag with a given key.
+      /*!
+        \param tag_key The key of the tag.
+      */
+      std::string get_tag_name(int tag_key) const;
```
kranar, line 39: Return ```const &```.

Applications/Spire/Include/Spire/KeyBindings/TagModel.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+      //! Constructs a TagModel.
+      /*!
+        \param schemas A list of Schemas of the tags to store.
+      */
+      explicit TagModel(std::vector<Schema> schemas);
+      
+      //! Returns the display name of the tag with a given key.
+      /*!
+        \param tag_key The key of the tag.
```
kranar, line 37: ```tag_key``` -> ```key```

Applications/Spire/Include/Spire/KeyBindings/TagModel.hpp
```diff
@@ -0,0 +1,53 @@
...
+      */
+      std::string get_tag_name(int tag_key) const;
+      
+      //! Returns a list of tags enabled for a set of prior tags.
+      /*!
+        \param prior_tags A list of already set tags.
+      */
+      std::vector<KeyBindings::Tag> load_enabled_tags(
```
kranar, line 45: ```load_enabled_tags``` -> ```load_satisfied_tags```

```prior_tags``` -> ```tags```

File: Applications/Spire/Include/Spire/KeyBindings/TagPresentTerm.hpp [hunk 13]
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Spire {
+
+  //! Checks whether a list of tags contains a tag with a selected key.
+  class TagPresentTerm : public Term {
+    public:
+
+      //! Constructs a TagPresentTerm.
+      /*!
+        \param tag_key The key of the tag that should be present.
```
kranar, line 15: ```tag_key``` -> ```key```.

Applications/Spire/Include/Spire/KeyBindings/TagPresentTerm.hpp
```diff
@@ -0,0 +1,27 @@
...
+      */
+      explicit TagPresentTerm(int tag_key);
+
+      virtual bool is_satisfied(const std::vector<KeyBindings::Tag>&
+        tags) const override;
+
+    private:
+      int m_tag_key;
```
kranar, line 23: ```m_tag_key``` -> ```m_key```.

Applications/Spire/Source/KeyBindings/KeyBindings.cpp
kranar: Add overloaded constructor that takes an ```OrderType::Type``` and a ```Side::Type```.

Applications/Spire/Source/KeyBindings/KeyBindings.cpp
```diff
@@ -129,18 +189,51 @@ std::vector<KeyBindings::ActionBinding>
...
+    const boost::optional<Region>& region) {
+  if(!region) {
+    return out << "(None)";
+  }
+  out << '(' << region->GetName() << " (";
+  for(auto country_code : region->GetCountries()) {
+    out << ", " << country_code;
+  }
+  out << ") (";
+  for(auto security : region->GetSecurities()) {
```
kranar, line 202: ```auto&```

Applications/Spire/Source/KeyBindings/RegionTerm.cpp
```diff
@@ -0,0 +1,22 @@
...
+  auto it = std::find_if(tags.begin(), tags.end(), [] (auto& tag) {
+    return tag.get_key() == 55;
+  });
+  if(it != tags.end()) {
+    auto value = it->get_value<Region>();
+    if(value) {
+      return *value <= m_region;
+    }
+  }
+  return m_region.IsGlobal();
```
kranar, line 21: If no region is present this term shall not be satisfied.

Applications/Spire/Source/KeyBindings/TagModel.cpp
```diff
@@ -0,0 +1,24 @@
+#include "Spire/KeyBindings/TagModel.hpp"
+#include <algorithm>
+
+using namespace Spire;
+
+TagModel::TagModel(std::vector<Schema> schemas)
+  : m_schemas(std::move(schemas)) {}
+
+std::string TagModel::get_tag_name(int tag_key) const {
+  return std::find_if(m_schemas.begin(), m_schemas.end(), [=] (auto& schema) {
```
kranar, line 10: Capture by reference.

Applications/Spire/Include/Spire/KeyBindings/KeyBindings.hpp
andrei-sa: I can, but as there are public setters for name and tags I thought it might be considered unsafe.
Should I also return `boost::optional<const Tag&>` from `get_tag` then?

Applications/Spire/Include/Spire/KeyBindings/KeyBindings.hpp
andrei-sa: It won't compile, which I think is due to the fact that `boost::optional` supports references.

Applications/Spire/Include/Spire/KeyBindings/KeyBindings.hpp
andrei-sa: Yes, returning a nullopt is somewhat random and can really conceal issues. Especially for implicitly convertible types, e.g. in:

```
auto tag = Tag(1, 10);
auto value = tag.get_value<double>();
```

There isn't a lot of exception handling code in Spire, so I am not sure about the convention in this regard. As an option, I can throw `std::bad_cast` here.

Applications/Spire/Include/Spire/KeyBindings/KeyBindings.hpp
kranar: Gotcha, that sucks.

Applications/Spire/Include/Spire/KeyBindings/KeyBindings.hpp
kranar: Yes, ```optional<T&>``` is the reason our codebase doesn't use ```std::optional```.

Applications/Spire/Source/KeyBindings/TagModel.cpp
andrei-sa: So, capture by reference even applies to captures less than a pointer size, right?

kranar: Good observation but yes, the choice of capture will be based on lifetime rather than optimization and we'll defer to the compiler to optimize it away through inlining.

Applications/Spire/Source/KeyBindings/KeyBindings.cpp
andrei-sa: I think I can avoid API bloat using a slightly different approach. Instead of adding more constructors, I add such types into the variant type list, so that it is capable of storing both `OrderType` and `OrderTypeEnumMembers::Type`. To allow seamless casts like this:

```
auto tag = Tag(1, Side::BID);
auto val1 = tag.get_value<OrderType>(); // OK
auto val2 = tag.get_value<OrderTypeEnumMembers::Type>(); // OK
```

I then enable implicit (and optionally explicit as well) conversions in` Tag::get_value<T>` by converting `boost::optional<StoredType>` to `boost::optional<QueryType>`, which is easily done.

The side effect is then that queries like below will also succeed:

```
auto tag = Tag(1, 10);
auto value tag.get_value<double>();
```

which might be either desired or undesired depending on our preference. I think it is more user-friendly if any convertible (at least implicitly) type can be user as a tag's value during initialization, and the set of such types can change without the need to modify `KeyBindings`.

andrei-sa: Disregard the last part about modifying `KeyBindings`, one still has to add new convertible types to the variant. However, failing to do so will result in a compilation error, making it quite safe.

Applications/Spire/Source/KeyBindings/RegionTerm.cpp
andrei-sa: Does it also apply if a region tag is present but its value is not set? Or should it be interpreted as a global region?

Applications/Spire/Source/KeyBindings/KeyBindings.cpp
kranar: I do like you're thinking on this issue and your solution is very friendly/permissive. Unfortunately I get worried about permissive and flexible code in C++ because without the right abstraction built directly into the language, permissive code in C++ almost always ends up breaking in a subtle and disasterous way at the worst possible time.

It will not be desirable to ```get_value<OrderType::Type>```, only ```get_value<OrderType>```. It also should not be permissible to store an ```int``` and get back a ```double```. Every tag is required to work with a specific type and may only work with that specific type.

You can add the constructors for ```Side::Type``` and ```OrderType::Type``` or you can require the user to have to wrap those values explicitly in which case ```Tag(1, Side::BID)``` is a compiler error, and it's not really the end of the world.

Applications/Spire/Source/KeyBindings/RegionTerm.cpp
kranar: We might change this later depending on feedback from users, but for now I suspect if a region is present but unset, it shall not satisfy this term.

Applications/Spire/Source/KeyBindings/KeyBindings.cpp
andrei-sa: Makes sense. The case with `Side` and `OrderType` is not that bad due to compiler errors.
What I think is more error-prone is something like `Quantity`, because both double and `Quantity` are accepted by the variant, but querying querying `Quantity` will fail if I store a `double`. In fact, I did this mistake already in `get_default_key_bindings` and test cases (which I forgot to commit, will include along with the fixes). I could implement tag dispatch or something to require explicit specification of a type, but this adds even more boilerplate and does not solve the problem completely. Anyway, with cast failures, it will at least manifest itself in runtime, so I think I will keep it as is and avoid adding extra constructors so that a user is prepared to explicitly cast the types on their own.

PR #480, opened 2020-07-23
Local group page model

spiredevang:


Applications/WebPortal/WebApp/library/source/pages/group_page/group_model.ts
```diff
@@ -0,0 +1,11 @@
+import * as Beam from 'beam';
+
+/** Interface for a model representing a single group. */
+export abstract class GroupModel {
+
+  /** The group this model represents. */
+  public abstract get group(): Beam.DirectoryEntry;
+
+  /** Loads this model. */
+  public abstract async load(): Promise<void>;
```
kranar, line 10: What is the purpose of this model? It's not doing anything.

spiresheep: We're going to be fleshing out the group page some more. I thought it would be a good idea to have a model for the page that we can build on with future pages.

kranar: Sure but this model will need to do something before it can be merged to master. Right now it has no functionality.

spiresheep: Then let's shelve it for now and then we'll use this when we're working on the controller?

kranar: Sure, you can keep this branch and resubmit the PR later.

PR #482, opened 2020-07-24
Uniform drop down

spireken:
The demo is called GenericDropDown.

Part 2 of this case is updating the windows to use the new drop down widgets, so I'll update the new class names to use the old names when I replace them.


Applications/Spire/Include/Spire/Ui/DropDownItem.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+  //! Represents an item used in a DropDownList.
+  class DropDownItem : public QWidget {
+    public:
+
+      //! Signals that the item was highlighted with the mouse.
+      /*!
+        \param value The value of the DropDownItem.
+      */
+      using HighlightedSignal = Signal<void (QVariant value)>;
```
andrei-sa, line 18: Any reason to take the variant by copy here and in `SelectedSignal`?

Applications/Spire/Include/Spire/Ui/DropDownWindow.hpp
```diff
@@ -0,0 +1,46 @@
...
+                                  otherwise.
+        \param parent The parent widget. Used to determine the position of the
+                      DropDownWindow.
+      */
+      explicit DropDownWindow(bool is_click_activated,
+        QWidget* parent = nullptr);
+
+      virtual ~DropDownWindow() = default;
+
+      //! Sets the widget to display in the DropDownWindow. Does nothing if a
```
andrei-sa, line 26: Try to reflect the "do nothing if" behavior in the method's name.

Applications/Spire/Include/Spire/Ui/FilteredDropDownMenu.hpp
```diff
@@ -0,0 +1,69 @@
...
+        \param value The value of the selected item.
+      */
+      using SelectedSignal = Signal<void (const QVariant& value)>;
+
+      //! Constructs a FilteredDropDownMenu.
+      /*!
+        \param items The initial items to display.
+        \param parent The parent widget.
+      */
+      explicit FilteredDropDownMenu(const std::vector<QVariant>& items,
```
andrei-sa, line 26: Use copy and move where possible.

Applications/Spire/Include/Spire/Ui/FilteredDropDownMenu.hpp
```diff
@@ -0,0 +1,69 @@
...
+      bool eventFilter(QObject* watched, QEvent* event) override;
+      void focusInEvent(QFocusEvent* event) override;
+      void keyPressEvent(QKeyEvent* event) override;
+      void mousePressEvent(QMouseEvent* event) override;
+      void paintEvent(QPaintEvent* event) override;
+
+    private:
+      mutable SelectedSignal m_selected_signal;
+      DropDownList* m_menu_list;
+      std::vector<QVariant> m_items;
```
andrei-sa, line 49: In general, prefer placing the members copied from constructors before the rest (unless violates other conventions, like, I suppose, mutable first).

Applications/Spire/Include/Spire/Ui/FilteredDropDownMenu.hpp
```diff
@@ -0,0 +1,69 @@
...
+      QVariant m_current_item;
+      QVariant m_last_activated_item;
+      bool m_was_click_focused;
+      CustomVariantItemDelegate m_item_delegate;
+      boost::signals2::scoped_connection m_list_selection_connection;
+
+      const std::vector<DropDownItem*> create_widget_items(
```
andrei-sa, line 56: Don't remove `const` by copy.

Applications/Spire/Include/Spire/Ui/FilteredDropDownMenu.hpp
```diff
@@ -0,0 +1,69 @@
...
+        const std::vector<QVariant>& items);
+      const std::vector<DropDownItem*> create_widget_items(
```
andrei-sa, line 58: Same here.

Applications/Spire/Include/Spire/Ui/StaticDropDownMenu.hpp
```diff
@@ -0,0 +1,106 @@
...
+      /*!
+        \param items The items to display.
+        \param display_text The fixed input label text to display.
+        \param parent The parent widget.
+      */
+      StaticDropDownMenu(const std::vector<QVariant>& items,
+        const QString& display_text, QWidget* parent = nullptr);
+
+      //! Returns the number of items in the list.
+      virtual int item_count() const;
```
andrei-sa, line 47: Not sure if we code natural numbers with signed integers in public methods, please confirm.

Applications/Spire/Include/Spire/Ui/TextInputWidget.hpp
```diff
@@ -0,0 +1,20 @@
...
+
+  //! Represents a Spire-style line edit.
+  class TextInputWidget : public QLineEdit {
+    public:
+
+      //! Constructs a TextInputWidget.
+      /*!
+        \param parent The parent widget.
+      */
+      TextInputWidget(QWidget* parent = nullptr);
```
andrei-sa, line 16: `explicit`

Applications/Spire/Source/Toolbar/ToolbarWindow.cpp
```diff
@@ -12,6 +12,18 @@ using namespace boost;
 using namespace boost::signals2;
 using namespace Spire;
 
+namespace {
+  auto ICON_SIZE() {
```
andrei-sa, line 16: I think there was a request to not capitalize constant-like functions by Kamal in one of the previous PRs, please refer to it.

Applications/Spire/Source/Ui/DropDownItem.cpp
```diff
@@ -0,0 +1,89 @@
...
+  }
+}
+
+DropDownItem::DropDownItem(const QVariant& value, QWidget* parent)
+  : DropDownItem(value, {}, parent) {}
+
+DropDownItem::DropDownItem(const QVariant& value, const QImage& icon,
+    QWidget* parent)
+    : QWidget(parent),
+      m_value(value),
```
andrei-sa, line 27: Use copy and move for both `value` and `icon`.

Applications/Spire/Source/Ui/DropDownItem.cpp
```diff
@@ -0,0 +1,89 @@
...
+    const SelectedSignal::slot_type& slot) const {
+  return m_selected_signal.connect(slot);
+}
+
+connection DropDownItem::connect_selected_signal(
+    const SelectedSignal::slot_type& slot) const {
+  return m_selected_signal.connect(slot);
+}
+
+const QVariant& DropDownItem::get_value() const {
```
andrei-sa, line 77: Unless `m_value` is never modified, returning a copy is likely safer.

Applications/Spire/Source/Ui/DropDownList.cpp
```diff
@@ -0,0 +1,247 @@
...
+  set_widget(m_scroll_area);
+  set_items(items);
+  parent->installEventFilter(this);
+}
+
+bool DropDownList::eventFilter(QObject* watched, QEvent* event) {
+  if(watched == parent()) {
+    if(event->type() == QEvent::KeyPress) {
+      auto e = static_cast<QKeyEvent*>(event);
+      if(e->key() == Qt::Key_Up) {
```
andrei-sa, line 37: Use `switch` with fallthrough for the last case.

Applications/Spire/Source/Ui/DropDownList.cpp
```diff
@@ -0,0 +1,247 @@
...
+    auto item = m_layout->itemAt(*m_highlight_index);
+    if(item != nullptr) {
+      static_cast<DropDownItem*>(item->widget())->reset_highlight();
+    }
+    m_highlight_index = boost::none;
+  }
+}
+
+void DropDownList::keyPressEvent(QKeyEvent* event) {
+  if(event->key() == Qt::Key_Up) {
```
andrei-sa, line 71: Use `switch`.

Applications/Spire/Source/Ui/DropDownList.cpp
```diff
@@ -0,0 +1,247 @@
...
+  if(m_layout->count() - 1 < index) {
+    return QVariant();
+  }
+  return get_widget(index)->get_value();
+}
+
+void DropDownList::insert_item(DropDownItem* item) {
+  m_item_selected_connections.push_back(std::move(
+    item->connect_selected_signal(
+      [=, index = m_layout->count()] (auto value) {
```
andrei-sa, line 120: What if two insertions are happening concurrently? Won't the same item be selected in both cases? If concurrent insertions are not allowed, why copy `index` in the first place?

Applications/Spire/Source/Ui/DropDownList.cpp
```diff
@@ -0,0 +1,247 @@
...
+  m_layout->insertWidget(m_layout->count(), item);
+  update_height();
+}
+
+int DropDownList::item_count() const {
+  return m_layout->count();
+}
+
+void DropDownList::remove_item(int index) {
+  if(index > m_layout->count() - 1 || index < 0) {
```
andrei-sa, line 132: This is probably another place where you should pass an unsigned integer as an argument unless, of course, negative indexes are logically possible.

Applications/Spire/Source/Ui/DropDownList.cpp
```diff
@@ -0,0 +1,247 @@
...
+    return;
+  }
+  m_item_selected_connections.pop_back();
+  for(auto i = std::size_t(index); i < m_item_selected_connections.size();
+      ++i) {
+    m_item_selected_connections[i] =
+      std::move(get_widget(i + 1)->connect_selected_signal(
+      [=, index = i] (auto value) {
```
andrei-sa, line 140: Remove `index` alias.

Applications/Spire/Source/Ui/DropDownList.cpp
```diff
@@ -0,0 +1,247 @@
...
+void DropDownList::set_items(std::vector<DropDownItem*> items) {
+  while(auto item = m_layout->takeAt(0)) {
+    delete item->widget();
+    delete item;
+  }
+  m_item_selected_connections.clear();
+  for(auto i = std::size_t(0); i < items.size(); ++i) {
+    m_layout->addWidget(items[i]);
+    m_item_selected_connections.push_back(std::move(
+      items[i]->connect_selected_signal([=, index = i] (auto value) {
```
andrei-sa, line 170: Remove `index = i`.

Applications/Spire/Source/Ui/DropDownWindow.cpp
```diff
@@ -0,0 +1,121 @@
...
+  return QWidget::event(event);
+}
+
+bool DropDownWindow::eventFilter(QObject* watched, QEvent* event) {
+  if(watched == m_widget) {
+    if(event->type() == QEvent::Resize) {
+      resize(m_widget->width() + 2, m_widget->height() + 2);
+    }
+  } else if(watched == parent()) {
+    if(event->type() == QEvent::Move) {
```
andrei-sa, line 39: Use switch on `event->type()`.

Applications/Spire/Source/Ui/FilteredDropDownMenu.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+const std::vector<DropDownItem*> FilteredDropDownMenu::create_widget_items(
+    const std::vector<QVariant>& items) {
+  return create_widget_items(items, "");
+}
+
+const std::vector<DropDownItem*> FilteredDropDownMenu::create_widget_items(
+    const std::vector<QVariant>& items, const QString& filter_text) {
+  auto widget_items = std::vector<DropDownItem*>();
+  widget_items.reserve(items.size());
```
andrei-sa, line 122: Don't reserve capacity unless the filter is supposed to be true almost all of the times.

Applications/Spire/Source/Ui/FilteredDropDownMenu.cpp
```diff
@@ -0,0 +1,207 @@
...
+  painter.drawText(QPoint(highlight_x_pos,
+    (height() / 2) + (metrics.ascent() / 2) - 1), highlight_text);
+}
+
+void FilteredDropDownMenu::on_editing_finished() {
+  if(text().isEmpty()) {
+    setText(m_item_delegate.displayText(m_current_item));
+    return;
+  }
+  if(m_last_activated_item.isValid()) {
```
andrei-sa, line 162: I think` else if `would be more readable than multiple `if` blocks with `return`.

Applications/Spire/Source/Ui/StaticDropDownMenu.cpp
```diff
@@ -0,0 +1,206 @@
...
+  m_menu_list->insert_item(item);
+}
+
+void StaticDropDownMenu::remove_item(int index) {
+  m_menu_list->remove_item(index);
+}
+
+void StaticDropDownMenu::set_items(const std::vector<QVariant>& items) {
+  auto widget_items = std::vector<DropDownItem*>();
+  widget_items.reserve(items.size());
```
andrei-sa, line 61: Prefer creating a sized vector followed with `std::transform` or a for loop.

Applications/Spire/Source/Ui/StaticDropDownMenu.cpp
```diff
@@ -0,0 +1,206 @@
...
+    auto item_widget = new DropDownItem(item, this);
+    item_widget->setFixedHeight(scale_height(20));
+    widget_items.push_back(item_widget);
+  }
+  m_menu_list->set_items(widget_items);
+  m_menu_list->setFixedWidth(width());
+}
+
+const QVariant& StaticDropDownMenu::get_current_item() const {
+  return m_current_item;
```
andrei-sa, line 72: Returning a copy might be safer if `m_current_item` can be modified.

Applications/Spire/Include/Spire/Ui/DropDownItem.hpp
spireken: The issue I was having was that the items can be removed from the list after they're selected, so DropDownItem's value might not exist after the signal is triggered.

Applications/Spire/Include/Spire/Ui/DropDownWindow.hpp
kranar: Why does this method do nothing if the widget has been set? Invariants like these should be part of the constructor.

kranar: If this method really can't reset the widget, then name it to ```initialize_widget```.

Applications/Spire/Source/Toolbar/ToolbarWindow.cpp
kranar: Nope, constants and constant-like functions should be capitalized.

Applications/Spire/Include/Spire/Ui/DropDownWindow.hpp
spireken: I wanted the DropDownList to be a DropDownWindow, but in order to have the ScrollArea be the widget for the DropDownList I had to delay setting the widget on the base class DropDownWindow until it's created by the DropDownList.

I was trying to avoid composing objects in certain cases like I mentioned in the case notes (issues with filters and layouts), but I think in this case I'll modify this to have the DropDownList just inherit from QWidget.

Applications/Spire/Source/Ui/DropDownList.cpp
spireken: By fallthrough do you mean omit the default case?

andrei-sa: Fallthrough means absense of `break` statement at the end of a case. You can structure this control flow in the following way:
```
case A:
  // action1
  break;
case B:
  // action2
  break;
case C:
case D:
  // action3
  break;
```

Applications/Spire/Source/Ui/DropDownList.cpp
spireken: I don't think two insertions could happen concurrently. The issue was that when the lambda is called it was referencing the current m_layout->count(), instead of the index when the lambda was created.

Applications/Spire/Include/Spire/Toolbar/ToolbarMenu.hpp
```diff
@@ -24,7 +24,7 @@ namespace Spire {
       */
       void add(const QString& text, const QImage& icon);
 
-      void remove_item(int index) override;
+      void remove_item(unsigned int index) override;
```
andrei-sa, line 27: For type `unsigned int`, I think the preferred spelling is `unsigned`. However, this type is rarely used. The most common unsigned type is the pointer size type as it is the correct way to represent array indexes, for example.
Having said that, I have noticed that Qt uses the signed `int` type for most logically unsigned methods. You may want to revert back to `int` or use `std::size_t`, whichever requires fewer casts and enables fewer uses with undefined behavior.

kranar: Use ```int```.

PR #484, opened 2020-07-31
Text input widget

spireken:


spireken:
Demos are TextInputWidget for the text input, BookViewUiTester's properties dialog for the hex input, and the KeyBindingsUiTester for the Name and Quantity columns.

Applications/Spire/Source/KeyBindingsUiTester/main.cpp
```diff
@@ -18,11 +18,11 @@ int main(int argc, char** argv) {
     DefaultCountries::US()));
   auto action1 = KeyBindings::OrderAction{"Name 1",
```
andrei-sa, line 19: Don't use uniform initialization here as `OrderAction` now has a constructor.

Applications/Spire/Source/Ui/ColorSelectorHexInputWidget.cpp
```diff
@@ -70,28 +68,30 @@ connection ColorSelectorHexInputWidget::connect_color_signal(
...
-    if(e->key() == Qt::Key_Enter || e->key() == Qt::Key_Return) {
-      auto text = m_line_edit->text();
-      if(text.length() == 6 || text.length() == 3) {
-        auto color = QColor();
-        color.setNamedColor(QString("#%1").arg(m_line_edit->text()));
-        m_color_name = color_name(color);
-        m_color_signal(color);
-        if(text.length() == 3) {
-          m_line_edit->setText(m_color_name);
+    switch(e->key()) {
```
andrei-sa, line 73: Use `if` instead as there is only a single branch here.
Normally, one would opt for `switch` only if there are more than two different branches (or expected to be in the future)

Applications/Spire/Source/Ui/TextInputWidget.cpp
```diff
@@ -0,0 +1,85 @@
...
+  font.setPixelSize(scale_height(12));
+  auto metrics = QFontMetrics(font);
+  setText(metrics.elidedText(text(),
+    Qt::ElideRight, width() - scale_width(16)));
+  QLineEdit::focusOutEvent(event);
+}
+
+void TextInputWidget::keyPressEvent(QKeyEvent* event) {
+  switch(event->key()) {
+    case Qt::Key_Up:
```
andrei-sa, line 40: You can probably do something like:
```
    case Qt::Key_Delete:
      setText("");
    case Qt::Key_Enter:
    case Qt::Key_Return:
      Q_EMIT editingFinished();
    case Qt::Key_Up:
    case Qt::Key_Down:
      return;
```

PR #485, opened 2020-07-31
Static drop down menu

spireken:
The demo is called StaticDropDownMenu. This case just implements the menu; there's another case to add it to the windows.

### DIFF
--- a/Applications/Spire/Include/Spire/Ui/DropDownItem.hpp
+++ b/Applications/Spire/Include/Spire/Ui/DropDownItem.hpp
@@ -1,0 +1,73 @@
+#ifndef SPIRE_DROP_DOWN_ITEM_HPP
+#define SPIRE_DROP_DOWN_ITEM_HPP
+#include <QImage>
+#include <QWidget>
+#include "Spire/Spire/Spire.hpp"
+#include "Spire/Ui/CustomQtVariants.hpp"
+
+namespace Spire {
+
+  //! Represents an item used in a DropDownList.
+  class DropDownItem : public QWidget {
+    public:
+
+      //! Signals that the item was highlighted with the mouse.
+      /*!
+        \param value The value of the DropDownItem.
+      */
+      using HighlightedSignal = Signal<void (QVariant value)>;
+
+      //! Signals that the item was selected.
+      /*!
+        \param value The value of the DropDownItem.
+      */
+      using SelectedSignal = Signal<void (QVariant value)>;
+
+      //! Constructs a text DropDownItem.
+      /*!
+        \param value The value represented by the DropDownItem.
+        \param parent The parent widget.
+      */
+      explicit DropDownItem(QVariant value, QWidget* parent = nullptr);
+
+      //! Consstructs a text DropDownItem with an icon.
+      /*!
+        \param value The value represented by the DropDownItem.
+        \param icon The icon to display beside the value.
+        \param parent The parent widget.
+      */
+      DropDownItem(QVariant value, QImage icon, QWidget* parent = nullptr);
+
+      //! Returns the item's value.
+      QVariant get_value() const;
+
+      //! Displays the item's highlight style.
+      void set_highlight();
+
+      //! Removes the item's highlight style.
+      void reset_highlight();
+
+      //! Connects a slot to the highlighted signal.
+      boost::signals2::connection connect_highlighted_signal(
+        const HighlightedSignal::slot_type& slot) const;
+      
+      //! Connects a slot to the selected signal.
+      boost::signals2::connection connect_selected_signal(
+        const SelectedSignal::slot_type& slot) const;
+
+    protected:
+      void enterEvent(QEvent* event) override;
+      void mousePressEvent(QMouseEvent* event) override;
+      void paintEvent(QPaintEvent* event) override;
+
+    private:
+      QVariant m_value;
+      QImage m_icon;
+      mutable HighlightedSignal m_highlighted_signal;
+      mutable SelectedSignal m_selected_signal;
+      bool m_is_highlighted;
+      CustomVariantItemDelegate m_item_delegate;
+  };
+}
+
+#endif
--- a/Applications/Spire/Include/Spire/Ui/DropDownItem.hpp
+++ b/Applications/Spire/Include/Spire/Ui/DropDownItem.hpp
@@ -39,7 +39,7 @@ namespace Spire {
       DropDownItem(QVariant value, QImage icon, QWidget* parent = nullptr);
 
       //! Returns the item's value.
-      QVariant get_value() const;
+      const QVariant& get_value() const;
 
       //! Displays the item's highlight style.
       void set_highlight();
--- a/Applications/Spire/Include/Spire/Ui/DropDownItem.hpp
+++ b/Applications/Spire/Include/Spire/Ui/DropDownItem.hpp
@@ -23,14 +23,14 @@ namespace Spire {
       */
       using SelectedSignal = Signal<void (QVariant value)>;
 
-      //! Constructs a text DropDownItem.
+      //! Constructs a DropDownItem.
       /*!
         \param value The value represented by the DropDownItem.
         \param parent The parent widget.
       */
       explicit DropDownItem(QVariant value, QWidget* parent = nullptr);
 
-      //! Consstructs a text DropDownItem with an icon.
+      //! Consstructs a DropDownItem with an icon.
       /*!
         \param value The value represented by the DropDownItem.
         \param icon The icon to display beside the value.
--- a/Applications/Spire/Include/Spire/Ui/DropDownList.hpp
+++ b/Applications/Spire/Include/Spire/Ui/DropDownList.hpp
@@ -50,6 +50,14 @@ namespace Spire {
       DropDownList(std::vector<DropDownItem*> items,
         bool is_click_activated, QWidget* parent = nullptr);
 
+      //! Activates the next item in the list without sending an activated
+      //! signal. Activates the first item if no item is currently activated.
+      void activate_next();
+
+      //! Activates the previous item in the list without sending an activated
+      //! signal. Activates the last item if no item is currently activated.
+      void activate_previous();
+
       //! Returns the value of the item at the given index.
       /*!
         \param index The index of the item.
@@ -120,8 +128,6 @@ namespace Spire {
         m_item_selected_connections;
 
       DropDownItem* get_widget(int index);
-      void focus_next();
-      void focus_previous();
       void set_highlight(int index);
       void scroll_to_highlight();
       void update_height();
--- a/Applications/Spire/Include/Spire/Ui/DropDownList.hpp
+++ b/Applications/Spire/Include/Spire/Ui/DropDownList.hpp
@@ -132,6 +132,7 @@ namespace Spire {
       void scroll_to_highlight();
       void update_height();
       void on_item_selected(QVariant value, int index);
+      void on_item_selected(QVariant value, DropDownItem* item);
   };
 }
 
--- a/Applications/Spire/Include/Spire/Ui/DropDownList.hpp
+++ b/Applications/Spire/Include/Spire/Ui/DropDownList.hpp
@@ -1,6 +1,7 @@
 #ifndef SPIRE_DROP_DOWN_LIST_HPP
 #define SPIRE_DROP_DOWN_LIST_HPP
 #include <QVBoxLayout>
+#include "Beam/SignalHandling/ConnectionGroup.hpp"
 #include "Spire/Spire/Spire.hpp"
 #include "Spire/Ui/CustomQtVariants.hpp"
 #include "Spire/Ui/DropDownItem.hpp"
@@ -40,12 +41,11 @@ namespace Spire {
       //! Constructs a DropDownList.
       /*!
         \param items The items to display.
-        \param is_click_activated True if the DropDownList should be shown
-                                  or hidden in response to parent widget
-                                  mouse/keyboard activation events, false
-                                  otherwise.
+        \param is_click_activated True iff the DropDownList should be shown
+               or hidden in response to parent widget mouse/keyboard activation
+               events.
         \param parent The parent widget. Used to determine the position of the
-                      DropDownList.
+               DropDownList.
       */
       DropDownList(std::vector<DropDownItem*> items,
         bool is_click_activated, QWidget* parent = nullptr);
@@ -62,7 +62,7 @@ namespace Spire {
       /*!
         \param index The index of the item.
       */
-      QVariant get_value(int index);
+      const QVariant& get_value(int index);
 
       //! Appends an item to the list.
       /*!
@@ -83,7 +83,7 @@ namespace Spire {
       //! text.
       /*!
         \param text The text to match against the item.
-        \return True if an item was highlighted, false if no match was found.
+        \return True iff an item was highlighted.
       */
       bool set_highlight(const QString& text);
 
@@ -91,7 +91,7 @@ namespace Spire {
       /*!
         \param items The new items to display.
       */
-      void set_items(std::vector<DropDownItem*> items);
+      void set_items(const std::vector<DropDownItem*>& items);
 
       //! Connects a slot to the activated signal.
       boost::signals2::connection connect_activated_signal(
@@ -111,7 +111,6 @@ namespace Spire {
 
     protected:
       bool eventFilter(QObject* watched, QEvent* event) override;
-      void hideEvent(QHideEvent* event) override;
       void keyPressEvent(QKeyEvent* event) override;
 
     private:
@@ -124,10 +123,9 @@ namespace Spire {
       ScrollArea* m_scroll_area;
       boost::optional<int> m_highlight_index;
       CustomVariantItemDelegate m_item_delegate;
-      std::vector<boost::signals2::scoped_connection>
-        m_item_selected_connections;
+      Beam::SignalHandling::ConnectionGroup m_item_selected_connections;
 
-      DropDownItem* get_widget(int index);
+      DropDownItem* get_item(int index);
       void set_highlight(int index);
       void scroll_to_highlight();
       void update_height();
--- a/Applications/Spire/Include/Spire/Ui/DropDownList.hpp
+++ b/Applications/Spire/Include/Spire/Ui/DropDownList.hpp
@@ -1,7 +1,7 @@
 #ifndef SPIRE_DROP_DOWN_LIST_HPP
 #define SPIRE_DROP_DOWN_LIST_HPP
+#include <Beam/SignalHandling/ConnectionGroup.hpp>
 #include <QVBoxLayout>
-#include "Beam/SignalHandling/ConnectionGroup.hpp"
 #include "Spire/Spire/Spire.hpp"
 #include "Spire/Ui/CustomQtVariants.hpp"
 #include "Spire/Ui/DropDownItem.hpp"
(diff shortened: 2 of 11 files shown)


Applications/Spire/Source/Ui/DropDownList.cpp
```diff
@@ -0,0 +1,259 @@
...
+int DropDownList::item_count() const {
+  return m_layout->count();
+}
+
+void DropDownList::remove_item(int index) {
+  if(index > m_layout->count() - 1) {
+    return;
+  }
+  m_item_selected_connections.pop_back();
+  for(auto i = std::size_t(index); i < m_item_selected_connections.size();
```
andrei-sa, line 148: Consider using `std::list` for `m_item_selected_connections`. List iterators won't be invalidated by insertions and deletions, so you may be able to avoid the array index trouble altogether.

spireken: I'm not sure how iterators would simplify this, the API for both removing and signalling item selection is done with indexes.

andrei-sa: You don't have to use indexes in the implementation just because the API works with them.

The problem with the existing code is that, while list items (connections) do not depend on each other, you have to update all the items on every write operation because array indexes do not completely identify the locations of the elements as you move them around. The solution to this problem is to fix the locations of the elements, capture those fixed locations in your lambdas, and move around handles rather than the elements themselves.

For example, consider storing `std::unique_ptr<boost::signals2::scoped_connection>` instead of `boost::signals2::scoped_connection` directly in `m_item_selected_connections`. On line 132, you could then capture the fixed address of the connection using `std::unique_ptr::get` so that you would never need to update the signal function. Of course, you would still need to keep some code to entertain the index API, but it does not have to be as operationally complex.

spireken: The connections are returned by value; wouldn't they get deleted when this function returns?

spireken:
@andrei-sa I couldn't figure out exactly what you had in mind for managing the connections, but I came up with a method to store the connections that doesn't require overwriting all of them with every removal, while still allowing the index of the selected item to be determined.

Applications/Spire/Source/Ui/DropDownList.cpp
```diff
@@ -0,0 +1,255 @@
...
+void DropDownList::set_items(std::vector<DropDownItem*> items) {
+  while(auto item = m_layout->takeAt(0)) {
+    delete item->widget();
+    delete item;
+  }
+  m_item_selected_connections.clear();
+  for(auto i = std::size_t(0); i < items.size(); ++i) {
+    m_layout->addWidget(items[i]);
+    m_item_selected_connections.push_back(items[i]->connect_selected_signal(
+      [=] (auto value) {
```
andrei-sa, line 174: Capture a single pointer instead of copying the entire array, e.g 
`[item = items[i]] (auto value) { ... }`

Applications/Spire/Include/Spire/Ui/DropDownItem.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+  //! Represents an item used in a DropDownList.
+  class DropDownItem : public QWidget {
+    public:
+
+      //! Signals that the item was highlighted with the mouse.
+      /*!
+        \param value The value of the DropDownItem.
+      */
+      using HighlightedSignal = Signal<void (QVariant value)>;
```
kranar, line 18: Can you explain again why a ```const &``` isn't being used for these signals?

Applications/Spire/Include/Spire/Ui/DropDownItem.hpp
```diff
@@ -0,0 +1,73 @@
...
+      //! Consstructs a DropDownItem with an icon.
+      /*!
+        \param value The value represented by the DropDownItem.
+        \param icon The icon to display beside the value.
+        \param parent The parent widget.
+      */
+      DropDownItem(QVariant value, QImage icon, QWidget* parent = nullptr);
+
+      //! Returns the item's value.
+      QVariant get_value() const;
```
kranar, line 42: Similarly, why not return a ```const &```?

Applications/Spire/Include/Spire/Ui/DropDownList.hpp
```diff
@@ -0,0 +1,139 @@
...
+      /*!
+        \param value The item's value.
+      */
+      using ValueSelectedSignal = Signal<void (const QVariant& value)>;
+
+      //! Constructs a DropDownList.
+      /*!
+        \param items The items to display.
+        \param is_click_activated True if the DropDownList should be shown
+                                  or hidden in response to parent widget
```
kranar, line 44: Going forward, continue the line below the name of the parameter.

Applications/Spire/Include/Spire/Ui/DropDownList.hpp
```diff
@@ -0,0 +1,139 @@
...
+                                  mouse/keyboard activation events, false
+                                  otherwise.
```
kranar, line 46: Instead of saying "True if X, false otherwise." use the convention "True iff X." as per:

https://en.wikipedia.org/wiki/If_and_only_if

Applications/Spire/Include/Spire/Ui/DropDownList.hpp
```diff
@@ -0,0 +1,139 @@
...
+      mutable ActivatedSignal m_activated_signal;
+      mutable HighlightedSignal m_highlighted_signal;
+      mutable IndexSelectedSignal m_index_selected_signal;
+      mutable ValueSelectedSignal m_value_selected_signal;
+      int m_max_displayed_items;
+      QVBoxLayout* m_layout;
+      ScrollArea* m_scroll_area;
+      boost::optional<int> m_highlight_index;
+      CustomVariantItemDelegate m_item_delegate;
+      std::vector<boost::signals2::scoped_connection>
```
kranar, line 127: Since this happens a lot, look into ```Beam::SignalHandling::ConnectionGroup``` found at ```Beam/SignalHandling/ConnectionGroup.hpp```

Applications/Spire/Include/Spire/Ui/DropDownWindow.hpp
```diff
@@ -0,0 +1,45 @@
...
+                                  or hidden in response to parent widget
+                                  mouse/keyboard activation events, false
+                                  otherwise.
+        \param parent The parent widget. Used to determine the position of the
+                      DropDownWindow.
+      */
+      explicit DropDownWindow(bool is_click_activated,
+        QWidget* parent = nullptr);
+
+      virtual ~DropDownWindow() = default;
```
kranar, line 24: What happens if you remove this destructor?

Applications/Spire/Include/Spire/Ui/ScrollArea.hpp
```diff
@@ -24,6 +24,8 @@ namespace Spire {
       */
       explicit ScrollArea(bool is_dynamic, QWidget* parent = nullptr);
 
+      virtual ~ScrollArea() = default;
```
kranar, line 27: Similarly, what happens if this destructor is removed?

Applications/Spire/Include/Spire/Ui/StaticDropDownMenu.hpp
```diff
@@ -0,0 +1,106 @@
...
+      //! Constructs a StaticDropDownMenu with a fixed input label.
+      /*!
+        \param items The items to display.
+        \param display_text The fixed input label text to display.
+        \param parent The parent widget.
+      */
+      StaticDropDownMenu(std::vector<QVariant> items,
+        const QString& display_text, QWidget* parent = nullptr);
+
+      virtual ~StaticDropDownMenu() = default;
```
kranar, line 44: Look to remove it.

Applications/Spire/Resources/Resources.qrc
```diff
@@ -109,5 +109,6 @@
         <file>Icons/text_effects/underline-active.svg</file>
         <file>Icons/text_effects/underline-hover.svg</file>
         <file>Icons/text_effects/underline-inactive.svg</file>
+        <file>Icons/arrow-down-grey.svg</file>
```
kranar, line 112: As a suggestion, would it be helpful to order these alphabetically? Is there a logical order to these resources?

Applications/Spire/Source/Ui/DropDownList.cpp
```diff
@@ -0,0 +1,255 @@
...
+          }
+          break;
+      }
+    }
+  }
+  return DropDownWindow::eventFilter(watched, event);
+}
+
+void DropDownList::hideEvent(QHideEvent* event) {
+  m_scroll_area->verticalScrollBar()->setValue(0);
```
kranar, line 67: Why is the value set to 0 and why are all the items removed?

Applications/Spire/Source/Ui/DropDownList.cpp
```diff
@@ -0,0 +1,255 @@
...
+  m_layout = new QVBoxLayout(main_widget);
+  m_layout->setContentsMargins({});
+  m_layout->setSpacing(0);
+  m_scroll_area->setWidget(main_widget);
+  initialize_widget(m_scroll_area);
+  set_items(items);
+  parent->installEventFilter(this);
+}
+
+
```
kranar, line 33: Remove vertical whitespace.

Applications/Spire/Source/Ui/DropDownList.cpp
```diff
@@ -0,0 +1,255 @@
...
+    }
+  }
+  return DropDownWindow::eventFilter(watched, event);
+}
+
+void DropDownList::hideEvent(QHideEvent* event) {
+  m_scroll_area->verticalScrollBar()->setValue(0);
+  if(m_highlight_index) {
+    auto item = m_layout->itemAt(*m_highlight_index);
+    if(item != nullptr) {
```
kranar, line 70: It is permissible, going forward, to use "falsy" in conditional checks, ie ```if(item) { ... }```. In the past the preference was to be explicit, but going forward it is preferable to use "falsy". You need not make the change for this PR however.

Applications/Spire/Source/Ui/DropDownList.cpp
```diff
@@ -0,0 +1,255 @@
...
+}
+
+connection DropDownList::connect_value_selected_signal(
+    const ValueSelectedSignal::slot_type& slot) const {
+  return m_value_selected_signal.connect(slot);
+}
+
+QVariant DropDownList::get_value(int index) {
+  if(m_layout->count() - 1 < index) {
+    return QVariant();
```
kranar, line 124: What you do to support returning a ```const &``` is along the lines of

    if(m_layout->count() - 1 < index) {
      static auto SENTINEL = QVariant();
      return SENTINEL;
    }
    return get_widget(index)->get_value();

Applications/Spire/Source/Ui/DropDownList.cpp
```diff
@@ -0,0 +1,255 @@
...
+  }
+  return get_widget(index)->get_value();
+}
+
+void DropDownList::insert_item(DropDownItem* item) {
+  m_item_selected_connections.push_back(item->connect_selected_signal(
+    [=] (auto value) {
```
kranar, line 131: You will likely end up taking an ```auto& value```, and then removing the ```std::move``` on line 132. This will happen when the signal is changed to pass a ```const &``` instead of passing by value.

Applications/Spire/Source/Ui/DropDownList.cpp
```diff
@@ -0,0 +1,255 @@
...
+  return false;
+}
+
+void DropDownList::set_items(std::vector<DropDownItem*> items) {
+  while(auto item = m_layout->takeAt(0)) {
+    delete item->widget();
+    delete item;
+  }
+  m_item_selected_connections.clear();
+  for(auto i = std::size_t(0); i < items.size(); ++i) {
```
kranar, line 171: ```for(auto& item : items)```

Applications/Spire/Source/Ui/DropDownList.cpp
```diff
@@ -0,0 +1,255 @@
...
+    m_layout->addWidget(items[i]);
+    m_item_selected_connections.push_back(items[i]->connect_selected_signal(
+      [=, item = items[i]] (auto value) {
```
kranar, line 174: See above about pass by reference/move.

Applications/Spire/Source/Ui/DropDownList.cpp
```diff
@@ -0,0 +1,255 @@
...
+    if(m_item_delegate.displayText(get_value(i)).startsWith(text,
+        Qt::CaseInsensitive)) {
+      set_highlight(i);
+      return true;
+    }
+  }
+  return false;
+}
+
+void DropDownList::set_items(std::vector<DropDownItem*> items) {
```
kranar, line 165: Doesn't look like the implementation needs a copy of ```items```.

Applications/Spire/Source/Ui/DropDownList.cpp
```diff
@@ -0,0 +1,255 @@
...
+  if(m_layout->count() > 0) {
+    update_height();
+    static_cast<DropDownItem*>(m_layout->itemAt(0)->widget())->set_highlight();
+    m_highlight_index = 0;
+  } else {
+    hide();
+  }
+}
+
+DropDownItem* DropDownList::get_widget(int index) {
```
kranar, line 187: Use a symmetric naming convention. If you use ```set_items``` as the setter, then use ```get_item``` as the getter.

Applications/Spire/Source/Ui/DropDownList.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+void DropDownList::activate_next() {
+  if(m_layout->count() == 0) {
+    return;
+  }
+  if(!m_highlight_index) {
+    set_highlight(0);
+    return;
+  }
+  set_highlight((*m_highlight_index) + 1);
```
kranar, line 199: Don't need the extra brackets, ```(*m_highlight_index + 1)```.

Applications/Spire/Source/Ui/DropDownList.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+void DropDownList::activate_previous() {
+  if(m_layout->count() == 0) {
+    return;
+  }
+  if(!m_highlight_index) {
+    set_highlight(m_layout->count() - 1);
+    return;
+  }
+  set_highlight((*m_highlight_index) - 1);
```
kranar, line 210: See above.

Applications/Spire/Source/Ui/DropDownList.cpp
```diff
@@ -0,0 +1,255 @@
...
+}
+
+void DropDownList::set_highlight(int index) {
+  if(m_highlight_index) {
+    get_widget(*m_highlight_index)->reset_highlight();
+  }
+  if(index < 0) {
+    index = m_layout->count() - 1;
```
kranar, line 218: Not a fan of this approach. You are trying to accommodate ```activate_previous``` and ```activate_next``` with this implementation. Instead, ```set_highlight``` should strictly apply a highlight on the item at the specified ```index```. It is up to ```activate_XXX``` to perform the "wrapping" logic or any other logic they need to determine what the "next" item is. It's not the job of ```set_highlight``` to perform this.

Applications/Spire/Source/Ui/DropDownWindow.cpp
```diff
@@ -0,0 +1,131 @@
...
+    scale_width(1), scale_height(1));
+  parent->installEventFilter(this);
+  parent->window()->installEventFilter(this);
+  hide();
+}
+
+bool DropDownWindow::event(QEvent* event) {
+  if(event->type() == QEvent::WindowDeactivate &&
+      focusWidget() != nullptr &&
+      !isAncestorOf(focusWidget())) {
```
kranar, line 27: Fit line 27 on line 26.

Applications/Spire/Source/Ui/DropDownWindow.cpp
```diff
@@ -0,0 +1,131 @@
...
+    hide();
+  }
+  return QWidget::event(event);
+}
+
+bool DropDownWindow::eventFilter(QObject* watched, QEvent* event) {
+  if(watched == m_widget) {
+    if(event->type() == QEvent::Resize) {
+      resize(m_widget->width() + 2, m_widget->height() + 2);
```
kranar, line 36: Briefly explain this, why 2 pixels? Why to both the width and height?

Applications/Spire/Source/Ui/StaticDropDownMenu.cpp
```diff
@@ -0,0 +1,212 @@
...
+}
+
+void StaticDropDownMenu::remove_item(int index) {
+  m_menu_list->remove_item(index);
+}
+
+void StaticDropDownMenu::set_items(std::vector<QVariant> items) {
+  auto widget_items = std::vector<DropDownItem*>(items.size());
+  std::transform(items.begin(), items.end(), widget_items.begin(),
+    [=] (const auto& item) {
```
kranar, line 62: Capture by ```&``` if the lambda is guaranteed to not outlive what it captures.

Applications/Spire/Include/Spire/Ui/DropDownItem.hpp
spireken: The issue was that the item might be removed from the list in response to the signal, so anything else listening to the signal may receive a reference to an object that was destroyed.

Applications/Spire/Include/Spire/Ui/DropDownWindow.hpp
spireken: Aren't virtual destructors required when inheriting from another class to ensure that the base class is destructed properly?

Applications/Spire/Include/Spire/Ui/DropDownItem.hpp
kranar: That's fine, what you do in this case is something along the lines of:

    auto value = somevalue;
    m_highlighted_signal(value);

In other words, whatever calls the signal makes a single copy and then passes a reference to that copy to the signal. This avoids every receiver from having to make a copy.

Applications/Spire/Resources/Resources.qrc
spireken: I typically edit this file through Qt Creator, so it appears the resources just get added in chronological order.

I don't find modifying the resources file cumbersome, but it might be a good idea to add more prefixes if you're expecting to add a lot of resources in the future, so the resources look like:

Icons/BookView/file1.png
Icons/SecurityInput/file2.png
Icons/UI/file3.png
etc...

Applications/Spire/Source/Ui/DropDownList.cpp
spireken: I'm not seeing where the list items get removed, but the old default was to have the list scroll back to the top and remove the current highlight. I'm not sure if the filtered drop down menu will use this, but I'll remove it.

Applications/Spire/Include/Spire/Ui/DropDownWindow.hpp
kranar: Good point, however, only the top-most class needs to define a virtual destructor. In this case the top-most class is ```QObject``` and it does define a virtual destructor. As such, all descendants of ```QObject``` will inherit that destructor and so you only need one if you intend to extend its behavior.

Applications/Spire/Resources/Resources.qrc
kranar: Okay, something to think about for another task then.

Applications/Spire/Source/Ui/DropDownList.cpp
kranar: Okay thanks for clarifying.

Applications/Spire/Source/Ui/DropDownWindow.cpp
spireken: The DropDownWindow has a border on all sides, so it's an allowance to have the border visible.

kranar: Makes sense.

PR #490, opened 2020-08-10
Quantity input

spireken:
The demo is called QuantityInputWidget. The two labels show when the change and submit signals are triggered, and they also take focus. The acceptable range is [-9999, 9999]. The disabled input on the right is to demonstrate the disabled style.

### DIFF
--- a/Applications/Spire/Include/Spire/Ui/DecimalInputWidget.hpp
+++ b/Applications/Spire/Include/Spire/Ui/DecimalInputWidget.hpp
@@ -14,13 +14,15 @@ namespace Spire {
     protected:
       void focusInEvent(QFocusEvent* event) override;
       void keyPressEvent(QKeyEvent* event) override;
+      void mousePressEvent(QMouseEvent* event) override;
 
     private:
       double m_value;
       int m_last_cursor_pos;
 
-      void on_editing_finished();
+      void add_step(int step, Qt::KeyboardModifiers modifiers);
       void revert_cursor();
+      void on_editing_finished();
       void on_text_edited(const QString& text);
       void on_value_changed(double value);
   };
--- a/Applications/Spire/Include/Spire/Ui/DecimalInputWidget.hpp
+++ b/Applications/Spire/Include/Spire/Ui/DecimalInputWidget.hpp
@@ -12,14 +12,15 @@ namespace Spire {
       QString textFromValue(double value) const override;
 
     protected:
+      void focusInEvent(QFocusEvent* event) override;
       void keyPressEvent(QKeyEvent* event) override;
 
     private:
       double m_value;
       int m_last_cursor_pos;
 
-      void on_cursor_position_changed(int old_pos, int new_pos);
       void on_editing_finished();
+      void revert_cursor();
       void on_text_edited(const QString& text);
       void on_value_changed(double value);
   };
--- a/Applications/Spire/Include/Spire/Ui/DecimalInputWidget.hpp
+++ b/Applications/Spire/Include/Spire/Ui/DecimalInputWidget.hpp
@@ -45,6 +45,7 @@ namespace Spire {
 
       void add_step(int step, Qt::KeyboardModifiers modifiers);
       void revert_cursor();
+      void set_stylesheet(bool is_up_disabled, bool is_down_disabled);
       void on_text_edited(const QString& text);
       void on_value_changed(double value);
   };
--- a/Applications/Spire/Include/Spire/Ui/DecimalInputWidget.hpp
+++ b/Applications/Spire/Include/Spire/Ui/DecimalInputWidget.hpp
@@ -39,6 +39,7 @@ namespace Spire {
     private:
       mutable ValueSignal m_change_signal;
       mutable ValueSignal m_submit_signal;
+      double m_last_valid_value;
       int m_last_cursor_pos;
       bool m_has_first_click;
 
--- a/Applications/Spire/Include/Spire/Ui/DecimalInputWidget.hpp
+++ b/Applications/Spire/Include/Spire/Ui/DecimalInputWidget.hpp
@@ -19,8 +19,6 @@ namespace Spire {
       */
       explicit DecimalInputWidget(double value, QWidget* parent = nullptr);
 
-      QString textFromValue(double value) const override;
-
       //! Connects a slot to the value change signal.
       boost::signals2::connection connect_change_signal(
         const ValueSignal::slot_type& slot) const;
@@ -35,6 +33,7 @@ namespace Spire {
       void focusOutEvent(QFocusEvent* event) override;
       void keyPressEvent(QKeyEvent* event) override;
       void mousePressEvent(QMouseEvent* event) override;
+      QString textFromValue(double value) const override;
 
     private:
       mutable ValueSignal m_change_signal;
--- a/Applications/Spire/Include/Spire/Ui/DecimalInputWidget.hpp
+++ b/Applications/Spire/Include/Spire/Ui/DecimalInputWidget.hpp
@@ -1,0 +1,28 @@
+#ifndef SPIRE_DECIMAL_INPUT_WIDGET_HPP
+#define SPIRE_DECIMAL_INPUT_WIDGET_HPP
+#include <QDoubleSpinBox>
+
+namespace Spire {
+
+  class DecimalInputWidget : public QDoubleSpinBox {
+    public:
+
+      explicit DecimalInputWidget(double value, QWidget* parent = nullptr);
+
+      QString textFromValue(double value) const override;
+
+    protected:
+      void keyPressEvent(QKeyEvent* event) override;
+
+    private:
+      double m_value;
+      int m_last_cursor_pos;
+
+      void on_cursor_position_changed(int old_pos, int new_pos);
+      void on_editing_finished();
+      void on_text_edited(const QString& text);
+      void on_value_changed(double value);
+  };
+}
+
+#endif
--- a/Applications/Spire/Include/Spire/Ui/DecimalInputWidget.hpp
+++ b/Applications/Spire/Include/Spire/Ui/DecimalInputWidget.hpp
@@ -1,16 +1,12 @@
 #ifndef SPIRE_DECIMAL_INPUT_WIDGET_HPP
 #define SPIRE_DECIMAL_INPUT_WIDGET_HPP
 #include <QDoubleSpinBox>
-#include "Spire/Spire/Spire.hpp"
 
 namespace Spire {
 
   //! Represents a widget for displaying and modifying decimal numbers.
   class DecimalInputWidget : public QDoubleSpinBox {
     public:
-
-      //! Signals a user interaction with the value.
-      using ValueSignal = Signal<void (double value)>;
 
       //! Constructs a DecimalInputWidget.
       /*!
@@ -19,13 +15,7 @@ namespace Spire {
       */
       explicit DecimalInputWidget(double value, QWidget* parent = nullptr);
 
-      //! Connects a slot to the value change signal.
-      boost::signals2::connection connect_change_signal(
-        const ValueSignal::slot_type& slot) const;
-
-      //! Connects a slot to the value submit signal.
-      boost::signals2::connection connect_submit_signal(
-        const ValueSignal::slot_type& slot) const;
+      QString textFromValue(double value) const override;
 
     protected:
       bool eventFilter(QObject* watched, QEvent* event) override;
@@ -33,11 +23,8 @@ namespace Spire {
       void focusOutEvent(QFocusEvent* event) override;
       void keyPressEvent(QKeyEvent* event) override;
       void mousePressEvent(QMouseEvent* event) override;
-      QString textFromValue(double value) const override;
 
     private:
-      mutable ValueSignal m_change_signal;
-      mutable ValueSignal m_submit_signal;
       double m_last_valid_value;
       int m_last_cursor_pos;
       bool m_has_first_click;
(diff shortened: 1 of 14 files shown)


Applications/Spire/Include/Spire/Ui/IntegerInputWidget.hpp
```diff
@@ -1,29 +1,36 @@
...
       //! Constructs a ValueInputWidget.
       /*
-        \param min_value The lowest value that can be entered.
-        \param max_value The highest value that can be entered.
+        \param value The initial value to display.
         \param parent The parent widget.
       */
-      IntegerInputWidget(int min_value, int max_value,
-        QWidget* parent = nullptr);
+      IntegerInputWidget(int value, QWidget* parent = nullptr);
```
andrei-sa, line 19: `explicit`

Applications/Spire/Source/Ui/DecimalInputWidget.cpp
```diff
@@ -0,0 +1,269 @@
...
+  set_stylesheet(false, false);
+  setContextMenuPolicy(Qt::NoContextMenu);
+  connect(this, QOverload<double>::of(&QDoubleSpinBox::valueChanged), this,
+    &DecimalInputWidget::on_value_changed);
+  connect(lineEdit(), &QLineEdit::textEdited, this,
+    &DecimalInputWidget::on_text_edited);
+  lineEdit()->installEventFilter(this);
+}
+
+QString DecimalInputWidget::textFromValue(double value) const {
```
andrei-sa, line 48: This method does not use `this` and can be static.

Applications/Spire/Source/Ui/DecimalInputWidget.cpp
```diff
@@ -0,0 +1,269 @@
...
+  if(value == 0.0) {
```
andrei-sa, line 49: Use literals like `0.` unless requested otherwise.

Applications/Spire/Source/Ui/DecimalInputWidget.cpp
```diff
@@ -0,0 +1,269 @@
...
+    return "0";
+  }
+  QString str = QString::number(value, 'f', decimals());
```
andrei-sa, line 52: Use `auto`.

Applications/Spire/Source/Ui/DecimalInputWidget.cpp
```diff
@@ -0,0 +1,269 @@
...
+      lineEdit()->setText(textFromValue(value()));
+      m_last_valid_value = value();
+      m_submit_signal(m_last_valid_value);
+      return;
+    case Qt::Key_Up:
+      add_step(1, event->modifiers());
+      return;
+    case Qt::Key_Down:
+      add_step(-1, event->modifiers());
+      return;
```
andrei-sa, line 123: Just to confirm, should all these branches bypass the call `QDoubleSpinBox::keyPressEvent(event)` below?

Applications/Spire/Source/Ui/DecimalInputWidget.cpp
spireken: The textFromValue() method is overriding a base class method. I did change it to be protected to match the base class' declaration.

Applications/Spire/Source/Ui/DecimalInputWidget.cpp
spireken: Yes, the base class keyPressEvent() will have some kind of side effect (like highlighting all of the text) that I'm trying to avoid.

Applications/Spire/Source/Ui/DecimalInputWidget.cpp
```diff
@@ -46,10 +46,10 @@ DecimalInputWidget::DecimalInputWidget(double value, QWidget* parent)
 }
 
 QString DecimalInputWidget::textFromValue(double value) const {
-  if(value == 0.0) {
+  if(value == 0) {
```
andrei-sa, line 49: 0 is an integer literal. When comparing a value against a literal, always try to supply the same type of the literal as that of the value.
"0." (zero period) is a double literal and is a short form of "0.0".

spireken: If the output is the same I'd prefer to keep the trailing zero to make it clearer.

andrei-sa: Okay, restore the original double literal. If trailing zeros are prohibited (I do not know if they are), Kamal will probably let you know.

Applications/Spire/Source/Ui/IntegerInputWidget.cpp
```diff
@@ -8,82 +8,31 @@
 using namespace boost::signals2;
 using namespace Spire;
 
-namespace {
-  auto BUTTON_X() {
-    static auto x = scale_width(12);
-    return x;
-  }
+IntegerInputWidget::IntegerInputWidget(int value, QWidget* parent)
```
kranar, line 11: What do you think about using composition instead of inheritance for this?

spireken: Wouldn't it have the same problem if it was using a DecimalInputWidget?

kranar: Can you rephrase this question?

spireken: I'm guessing based on you asking about using composition versus inheritance that there is a way to reuse a common numeric input for each of the numeric types. I'm not sure how I can modify the default QSpinBox implementations to create a generic numeric input that can be reused.

I can create a numeric input that uses Quantity as the base type that can be reused (assuming Quantity can express all our required numeric types), but that would require redoing some work regarding editing and fixing up the text in the widget.

kranar: ```QDoubleSpinBox``` can be used for all the widgets using composition instead of inheritance.

kranar: You can inherit from ```QDoubleSpinBox``` to implement the ```DecimalInputWidget``` as you have done, and then reuse it for the ```QuantityInputWidget``` and ```IntegerInputWidget```.

spireken: I don't understand how that's different from inheritance in this case. The Quantity input for instance would still have to convert a double to a quantity (and vice-versa) if it was using the DecimalInputWidget.

kranar: No, you can not operate on the ```double```, you would need to operate on the ```QString``` to parse it into a ```Quantity```.

kranar: https://github.com/spiretrading/nexus/blob/3151c02be77001d09b15b2fcc5a30d8572328f12/Nexus/Include/Nexus/Definitions/Quantity.hpp#L37

kranar: Similarly when you implement the ```MoneyInputWidget``` in the future, you will also parse it into a ```Money```.

spireken: In order to call setMinimum(double), would I have to do something like:

`
auto qty = Quantity(1.234567)
auto str = item_delegate.displayText(qty);
setMinimum(str.toDouble());
`

spireken: Actually, displayText just casts the Quantity to a double anyway.

Applications/Spire/Include/Spire/Ui/DecimalInputWidget.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+      //! Constructs a DecimalInputWidget.
+      /*!
+        \param value The initial value to display.
+        \param parent The parent widget.
+      */
+      explicit DecimalInputWidget(double value, QWidget* parent = nullptr);
+
+      //! Connects a slot to the value change signal.
+      boost::signals2::connection connect_change_signal(
```
kranar, line 23: You may be able to get rid of these signals and stick to using ```QDoubleSpinBox```'s signals. All this widget does is style a ```QDoubleSpinBox``` correct?

Applications/Spire/Include/Spire/Ui/IntegerInputWidget.hpp
```diff
@@ -1,29 +1,36 @@
 #ifndef SPIRE_INTEGER_INPUT_WIDGET_HPP
 #define SPIRE_INTEGER_INPUT_WIDGET_HPP
-#include <QSpinBox>
+#include "Spire/Ui/DecimalInputWidget.hpp"
 
 namespace Spire {
 
   //! Represents a widget for inputting whole numbers.
-  class IntegerInputWidget : public QSpinBox {
+  class IntegerInputWidget : public DecimalInputWidget {
```
kranar, line 8: I still have hesitations that this inheritance is the right choice, but you can leave it for now.

Applications/Spire/Include/Spire/Ui/QuantityInputWidget.hpp
```diff
@@ -1,48 +1,61 @@
...
-#include "Spire/Ui/TextInputWidget.hpp"
-#include "Spire/Ui/Ui.hpp"
+#include "Spire/Ui/DecimalInputWidget.hpp"
 
 namespace Spire {
 
-  //! Displays an input box that accepts Quantities.
-  class QuantityInputWidget : public TextInputWidget {
+  //! Represents a widget for displaying and modifying Quantities.
+  class QuantityInputWidget : public QWidget {
```
kranar, line 9: Have you considered inheriting from ```QAbstractSpinBox```?

Applications/Spire/Source/Ui/QuantityInputWidget.cpp
```diff
@@ -1,46 +1,56 @@
...
+    const ValueSignal::slot_type& slot) const {
+  return m_submit_signal.connect(slot);
 }
 
-void QuantityInputWidget::on_line_edit_committed() {
-  if(!text().isEmpty()) {
-    m_committed_signal(*Quantity::FromValue(text().toStdString()));
-  }
+void QuantityInputWidget::on_editing_finished() {
+  m_submit_signal(*Quantity::FromValue(m_input_widget->text().toStdString()));
```
kranar, line 51: You need to validate anything that you parse.

Applications/Spire/Source/Ui/QuantityInputWidget.cpp
```diff
@@ -1,46 +1,56 @@
...
+}
+
+void QuantityInputWidget::set_maximum(Nexus::Quantity maximum) {
+  m_input_widget->setMaximum(static_cast<double>(maximum));
 }
 
 void QuantityInputWidget::set_value(Quantity value) {
-  setText(m_item_delegate.displayText(
-    QVariant::fromValue(Truncate(value, 0)), m_locale));
+  m_input_widget->setValue(static_cast<double>(value));
```
kranar, line 37: This isn't possible, not all ```Quantity```'s can be expressed as a ```double```. This is a big issue because even a common value like ```1.1``` is a valid ```Quantity``` but not a valid ```double```.

Applications/Spire/Include/Spire/Ui/DecimalInputWidget.hpp
spireken: I might be able to get rid of the signals, but there's a few changes related to editing that go beyond just a style change.

kranar: Okay, keep all the logic then, remove the signals.

Applications/Spire/Include/Spire/Ui/QuantityInputWidget.hpp
spireken: I did, but it involves defining the valueFromText, textFromValue, and validate functions because the default implementations do nothing, among other tasks to get it to work. I thought a double could represent the types we need so modifying the QDoubleSpinBox would be the most straightforward way to implement this.

kranar: Okay sounds good.

Applications/Spire/Source/Ui/QuantityInputWidget.cpp
spireken: I borrowed this from the CustomVariantItemDelegate::displayText(...) function that converts a Quantity to a string. Would I have to go Quantity -> QString -> double to convert this properly?

Also, how is 1.1 not a valid double? Or is it just the context?

kranar: There is no way to represent the value ```1.1``` as a double.

I can look into this a bit later today but I was hoping you could set the value of the ```QDoubleSpinBox``` using a ```QString``` instead of a ```double```.

spireken: I can create a getter to access the underlying QLineEdit to set the current text, but there's still the issue of setting the range. There is a private implementation (QAbstractSpinBoxPrivate) that has a setRange() function that takes QVariants.

kranar: Setting the range isn't a concern since there's no use case where a range is set to more than 15 decimal places.

spireken: Sorry, the range for the spin box means the acceptable input range from the minimum to maximum value.

Would it be possible to just use the conversion functions given by the QDoubleSpinBox? Converting the Quantity to a QString then passing it to QDoubleSpinBox::valueFromText() would give the double that the input would return from the given text.

kranar: Yep I got what you meant by range, it's not an issue for our use case.

Your solution of using ```valueFromText``` sounds good. Go with it.

spireken:
The QDoubleSpinBox::valueFromText() function bounds the value to the current range, which is why I 'faked' an unlimited range in the helper function.

PR #493, opened 2020-08-27
Filtered menu

spireken:
The demo is the FilteredDropDownDemo.

There was a discrepancy between the given padding for the TextInputWidget's stylesheet and the actual padding when Qt draws it, so the text would move from when the filtered menu text was being edited to when it was committed. The constants in the TextInputWidget's anonymous namespace come from QLineEdit's internals that are used to draw the text consistently. I imagine this is the same issue that is causing the KeyBindingItemDelegate's text to shift between editing and committing.


Applications/Spire/Source/Ui/FilteredDropDownMenu.cpp
```diff
@@ -0,0 +1,214 @@
...
+  }
+}
+
+void FilteredDropDownMenu::keyPressEvent(QKeyEvent* event) {
+  switch(event->key()) {
+    case Qt::Key_Delete:
+      setText("");
+      return;
+    case Qt::Key_Escape:
+      setText(m_item_delegate.displayText(m_current_item));
```
kranar, line 67: Is it the case that pressing ESC performs everything that pressing ENTER performs?

Applications/Spire/Source/Ui/FilteredDropDownMenu.cpp
```diff
@@ -0,0 +1,214 @@
...
+  if(m_was_click_focused) {
+    selectAll();
+    m_was_click_focused = false;
+  }
+}
+
+void FilteredDropDownMenu::paintEvent(QPaintEvent* event) {
+  TextInputWidget::paintEvent(event);
+  if(!text().isEmpty()) {
+    auto item = m_menu_list->get_value(0);
```
kranar, line 96: Can you use ```auto&```? What is the return type of ```get_value```?

Applications/Spire/Source/Ui/FilteredDropDownMenu.cpp
```diff
@@ -0,0 +1,214 @@
...
+  }
+  m_menu_list->hide();
+  set_items(m_items);
+  update();
+}
+
+void FilteredDropDownMenu::on_text_edited(const QString& text) {
+  m_last_activated_item = QVariant();
+  if(text.isEmpty()) {
+    m_menu_list->set_items(std::move(create_widget_items(m_items)));
```
kranar, line 203: Get rid of the ```std::move```.

Applications/Spire/Source/Ui/FilteredDropDownMenu.cpp
```diff
@@ -0,0 +1,214 @@
...
+    m_menu_list->show();
+    return;
+  }
+  auto items = std::move(create_widget_items(m_items, text));
```
kranar, line 207: Get rid of the ```std::move```.

Applications/Spire/Source/Ui/FilteredDropDownMenu.cpp
spireken: It isn't a problem to use auto& here because get_value(...) returns a const reference that will be valid for as long as item is being used.

PR #496, opened 2020-09-01
Filtered type widgets

spireken:
The demo is called FilteredTypeInputs.


kranar:
This branch will be closed and deleted in the future. Most of these widgets shall be reimplemented one-by-one.

PR #500, opened 2020-09-26
Side combo box

spireken:
Same demo, SideInputWidget.


Applications/Spire/Include/Spire/Ui/SideInputWidget.hpp
```diff
@@ -0,0 +1,35 @@
+#ifndef SPIRE_SIDE_INPUT_WIDGET_HPP
+#define SPIRE_SIDE_INPUT_WIDGET_HPP
+#include <QWidget>
+#include "Nexus/Definitions/Side.hpp"
+#include "Spire/Spire/Spire.hpp"
+
+namespace Spire {
+
+  //! Represents a widget for selecting Side values.
+  class SideInputWidget : public QWidget {
```
kranar, line 10: Let's start using descriptive names instead of generic terms. If this is a ComboBox then call it a ```SideComboBox```, if it's a drop down list then ```SideDropDownList```, etc...

Applications/Spire/Include/Spire/Ui/SideInputWidget.hpp
```diff
@@ -0,0 +1,35 @@
...
+    public:
+
+      //! Signals that a Side was selected.
+      /*!
+        \param side The selected Side.
+      */
+      using SideSignal = Signal<void (const Nexus::Side& side)>;
```
kranar, line 17: A signal indicates an action was performed and hence should be named after a verb. ```UpdateSignal``` is more descriptive than ```SideSignal```.

kranar: Pass ```Nexus::Side``` by value, refer to:

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-conventional

Applications/Spire/Include/Spire/Ui/SideInputWidget.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+      //! Constructs a SideInputWidget.
+      /*!
+        \param parent The parent widget.
+      */
+      SideInputWidget(QWidget* parent = nullptr);
```
kranar, line 23: Mark as explicit.

Applications/Spire/Source/Ui/SideInputWidget.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include "Spire/Ui/StaticDropDownMenu.hpp"
+
+using namespace boost::signals2;
+using namespace Nexus;
+using namespace Spire;
+
+SideInputWidget::SideInputWidget(QWidget* parent)
+    : QWidget(parent) {
+  auto menu = new StaticDropDownMenu({QVariant::fromValue<Side>(Side::ASK),
+        QVariant::fromValue<Side>(Side::BID)}, this);
```
kranar, line 12: Fix indentation.

kranar, line 12: There is no need to explicitly instantiate function templates in situations where they can be deduced based on the function arguments.

kranar (changes requested):

Applications/Spire/Source/Ui/SideInputWidget.cpp
spireken: In this case I might have to because the type is being deduced as: Nexus::SideEnumMembers::Type, which isn't registered as a QVariant type.

kranar: Good, good.

PR #501, opened 2020-09-28
Security input updates

spireken:


Applications/Spire/Include/Spire/SecurityInput/SecurityInfoWidget.hpp
```diff
@@ -1,25 +1,16 @@
...
-#include <QWidget>
 #include "Nexus/Definitions/SecurityInfo.hpp"
 #include "Spire/SecurityInput/SecurityInput.hpp"
+#include "Spire/Ui/DropDownItem.hpp"
 
 namespace Spire {
 
   //! A convenience widget to display a security's name and ticker symbol.
-  class SecurityInfoWidget : public QWidget {
+  class SecurityInfoWidget : public DropDownItem {
```
kranar, line 11: Sticking to the theme of using descriptive names, rename this to ```SecurityInfoItem```. Let's do our best to avoid using generic terms like ```Widget``` unless it truly is a widget that is dissimilar to anything else.

Applications/Spire/Include/Spire/Ui/TextInputWidget.hpp
```diff
@@ -22,6 +22,9 @@ namespace Spire {
       */
       explicit TextInputWidget(QString text, QWidget* parent = nullptr);
 
+      //! Returns the left text padding.
+      int get_padding() const;
+
       //! Sets the input's style to the table cell style.
       void set_cell_style();
```
kranar, line 29: Fix up these ```set_xxx_style``` methods:

    /** The styles available to render the TextInputWidget. */
    enum class Style {

      /** Render using the default style. */
      DEFAULT,

      /** Render using the table cell style. */
      CELL
    };

    void set_style(Style style);

Applications/Spire/Source/SecurityInput/SecurityInfoWidget.cpp
```diff
@@ -10,9 +11,8 @@ using namespace Nexus;
 using namespace Spire;
 
 SecurityInfoWidget::SecurityInfoWidget(SecurityInfo info, QWidget* parent)
-    : QWidget(parent),
-      m_info(std::move(info)),
-      m_is_highlighted(false) {
+    : DropDownItem(QVariant::fromValue<Security>(info.m_security), (parent)),
```
kranar, line 14: Avoid explicit template instantiation.

Applications/Spire/Source/SecurityInput/SecurityInputDialog.cpp
```diff
@@ -115,11 +118,24 @@ void SecurityInputDialog::mouseReleaseEvent(QMouseEvent* event) {
...
+  m_layout->addWidget(m_security_line_edit);
+  m_security_line_edit->show();
+  m_security_line_edit->setFocus();
+}
+
+void SecurityInputDialog::reject_dialog() {
+  disconnect(m_security_line_edit,
+    &SecurityInputLineEdit::editingFinished, nullptr, nullptr);
+  reject();
+  static_cast<QWidget*>(parent())->activateWindow();
```
kranar, line 138: Use ```parentWidget()``` over ```static_cast<QWidget*>(...)```.

kranar, line 138: What guarantee is there that a parent exists?

Applications/Spire/Source/SecurityInput/SecurityInputDialog.cpp
```diff
@@ -56,31 +54,36 @@ const Security& SecurityInputDialog::get_security() const noexcept {
...
+        return true;
+      case QEvent::KeyPress:
+        auto e = static_cast<QKeyEvent*>(event);
+        if(e->key() == Qt::Key_Escape) {
+          reject_dialog();
+          return true;
+        }
+        break;
+    }
+  } else if (watched == m_security_list) {
```
kranar, line 82: No space after ```if```.

Applications/Spire/Source/SecurityInput/SecurityInputLineEdit.cpp
```diff
@@ -164,14 +151,29 @@ void SecurityInputLineEdit::on_text_edited() {
         return std::vector<SecurityInfo>();
       }
     }();
-    m_securities->set_list(completions);
-    if(completions.empty()) {
+    auto widget_items = std::vector<DropDownItem*>(completions.size());
+    std::transform(completions.begin(), completions.end(),
+      widget_items.begin(), [&] (const auto& item) {
+        auto item_widget = new SecurityInfoWidget(item, this);
```
kranar, line 157: Remove ```item_widget``` entirely.

Applications/Spire/Source/SecurityInput/SecurityInputLineEdit.cpp
```diff
@@ -68,93 +61,87 @@ const Nexus::Security& SecurityInputLineEdit::get_security() const {
...
   Q_EMIT editingFinished();
-  m_commit_signal(security);
 }
 
 void SecurityInputLineEdit::on_text_edited() {
+  if(text().isEmpty()) {
+    m_is_suggestion_disabled = false;
+  }
   m_completions = m_model->autocomplete(text().toStdString());
   m_completions.then([=] (auto result) {
```
kranar, line 146: Merge line 146 and 145:

    m_completions = m_model->autocomplete(...).then([=] (auto result) { ... });

Applications/Spire/Source/Ui/RealSpinBoxModel.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+namespace {
+  auto to_real(const QVariant& value) {
+    static auto locale = [] {
+      auto locale = QLocale();
+      locale.setNumberOptions(locale.numberOptions().setFlag(
+        QLocale::OmitGroupSeparator, true));
+      return locale;
+    }();
+    return CustomVariantItemDelegate().displayText(value,
```
kranar, line 14: Use a static instance of ```CustomVariantItemDelegate``` just like you use a static instance of ```locale```.

Applications/Spire/Source/Ui/RealSpinBoxModel.cpp
```diff
@@ -0,0 +1,33 @@
...
+      locale).toStdString().c_str();
```
kranar, line 15: This function is called ```to_real``` but it returns a ```const char*```. Apart from being undefined behavior, the name of the function is not reflective of what this function actually does.

kranar (changes requested):

Applications/Spire/Source/SecurityInput/SecurityInputDialog.cpp
spireken: There's no hard guarantee, but given our coding standards it should always have one. I can remove the default parameter to ensure it always a parent, but I'm not sure why this would be a special case.

kranar: Perhaps a better question is why does this dialog call ```activateWindow``` on the parent? What would happen otherwise?

spireken: I added this when after closing the SecurityInputDialog the tester window would receive focus instead of the parent Time and Sales window.

kranar: Best to look into that issue directly. My understanding is that when a window closes, its parent is what gets activated by default.

Applications/Spire/Source/Ui/RealSpinBoxModel.cpp
spireken: Regarding the name, how is this function different from the other to_real(...) functions? Is it just semantics because the return type is const char*?

kranar: A function named ```to_X``` should return an object of type ```X```.

Applications/Spire/Source/SecurityInput/SecurityInputDialog.cpp
spireken: I can't reproduce it now, either on accepting or rejecting the SecurityInputDialog. It may have been a transient issue when updating the windows/widgets.

spireken: What I'm seeing is that Qt processes the system event and receives information on which window to activate next based on that event. I'll look into this more tomorrow.

spireken:
I spent about half of today looking into potential window activation issues when the SecurityInputDialog is closed, but couldn't find anything conclusive. It was likely related to something I was doing while updating and debugging the widgets, but I can no longer reproduce it.

kranar:
Good good.

PR #504, opened 2020-09-30
Added OrderStatusComboBox implementation

spireken:
Demo is called OrderStatusComboBox.

Would it make sense to insert the values manually, so they can be put in alphabetical order?

### DIFF
--- a/Applications/Spire/Source/Ui/OrderStatusComboBox.cpp
+++ b/Applications/Spire/Source/Ui/OrderStatusComboBox.cpp
@@ -1,0 +1,34 @@
+#include "Spire/Ui/OrderStatusComboBox.hpp"
+#include <QHBoxLayout>
+#include "Beam/Collections/EnumIterator.hpp"
+#include "Spire/Ui/StaticDropDownMenu.hpp"
+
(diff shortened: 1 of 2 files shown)


Applications/Spire/Source/Ui/OrderStatusComboBox.cpp
```diff
@@ -0,0 +1,34 @@
+#include "Spire/Ui/OrderStatusComboBox.hpp"
+#include <QHBoxLayout>
+#include "Beam/Collections/EnumIterator.hpp"
```
kranar, line 3: Treat Beam as a third party library, use ```<...>``` to include it.

spireken: So alphabetically it now is included before <QHBoxLayout>?

kranar: Included before what?

spireken: Mean to say that would `<QHBoxLayout>` be included after Beam includes?

PR #507, opened 2020-10-07
Drop down usage

spireken:
The changes for this PR are:

- Removing the old drop down classes and replacing them with the new ones.
- Updating the usage of the current menus to use the new menus.
- Fixing input issues related to the RealSpinBox class when the minimum is greater than zero, and fixing style issues by adding separate images.
- Adding a ComboBoxEditor to wrap the static menu for use as a combo box in a table cell.
- Updating the OrderType, Side, and TimeInForce delegates to use the ComboBoxEditor.

The Book View properties window demos the drop down menu usage (font menu) and the spin box. The Key Bindings demo shows the updated column delegates.


kranar:
These changes far exceed the description of the task on Asana. The task on Asana says only to update existing widgets to use the new drop down menu. This task introduces several breaking API changes, introduces new classes that have never been reviewed before, and overall is far too big of a change to incorporate into a single PR.

If you need to add something new or break the API, then first propose those changes in a new task or subtask, have it reviewed/approved and submit a PR for it. It is very rare that a PR will be accepted with breaking API changes that have not first been discussed.

PR #508, opened 2020-10-08
Menu updates

spireken:
I updated and used the Book View tester to demo the font family (properties dialog) using the static menu.

I updated the Key Bindings demo as well, but the changes here to the InputFieldEditor (which is going to be removed) and the delegates are temporary until the Key Bindings table is updated to use the combo boxes.

Also, I believe most of the changes here were already approved previously, but renaming the drop down classes (e.g., DropDownList to DropDownMenuList) makes it appear as if they were new changes.


kranar:
What task does this PR correspond to?

spireken:
This one: https://app.asana.com/0/1169210941620249/1186371420022201

This PR is just for the menus, so there will be a second PR for this case to update a few other classes (ColorSelector being one of them).

The Key Bindings table updates should be made into a separate case because the combo boxes will need a few updates to support table/cell editing.

kranar:
I don't see how the issue has been resolved. This PR is doing way too many things. The task says that all that is being done is to update a set of existing widgets to use the new drop down menu and yet this PR contains tons of breaking API changes, new classes and a host of things that are far beyond the scope of the task in Asana. There are far too many issues with the PR to accept it.

If you're introducing a new API, then propose the API first and once it's approved implement it and submit a PR for it.

spireken:
I can move the InputFieldEditor and Key Bindings delegate changes to another PR which might help to simplify this, but those changes are only temporary until the combo boxes are added to the table.

The only change I made to the menu class' APIs here was to add the set_current_item() method. It appears like I made a lot of changes to the APIs but that's only because the DropDownMenuX classes are being diffed against the previous versions, and not against the DropDownX classes that they were originally called.

PR #509, opened 2020-10-09
Real spin box min fix

spireken:
The previous implementation always relied on the current text to determine a valid value, but this wasn't sufficient for 'incomplete' values in the case of the font size input where the user couldn't input 1-5 when trying to input, say, 12.

I added a BookViewUiTester-SpinBox demo to demonstrate the fix. The value signal will only trigger when a valid value is input but will accept partial values that may otherwise be invalid.


kranar:
I will review this down the road, a task has been assigned to you regarding the UI Viewer. Work on that and get it done so that we have a general framework for testing UI components more thoroughly, once that's done you will incorporate this widget into the UI viewer and then submit a PR for it.

spireken:
I added a UiViewer-SpinBoxMinFix demo to use the UiViewer instead of the Book View as a demo. You can use a UiViewer from another branch to compare the fixed version to the previous.

Applications/Spire/Include/Spire/Ui/RealSpinBox.hpp
```diff
@@ -98,7 +98,8 @@ namespace Spire {
       QString display_string(Real value);
       QStyle::SubControl get_current_control(const QPoint& mouse_pos);
       boost::optional<Real> get_value(const QString& text) const;
-      bool is_valid(const QString& string);
+      bool is_valid_input(const QString& string);
```
kranar, line 101: Mark methods as ```const```

Applications/Spire/Source/Ui/RealSpinBox.cpp
```diff
@@ -31,6 +31,41 @@ namespace {
       return a.compare(b) == -1;
     });
   }
+
+  auto integer_digit_count(RealSpinBox::Real value) {
+    if(value.compare(std::int64_t(10)) == -1) {
```
kranar, line 36: Use a loop that repeatedly divides ```value``` by 10 until ```value < 1```.

Applications/Spire/Include/Spire/Ui/RealSpinBox.hpp
```diff
@@ -98,7 +98,8 @@ namespace Spire {
...
+      bool is_valid_value(const QString& string);
```
kranar, line 102: What is the difference been a value value and a valid input?

kranar (changes requested):

Applications/Spire/Include/Spire/Ui/RealSpinBox.hpp
spireken: is_valid_input(...) checks if the input text itself has the *potential* to make a valid number (i.e., accepts 'partial' input), but also prevents things like entering a negative sign when the minimum is >= 0.

is_valid_value(...) is the final check to see if the input text makes a valid value within the bounds of the SpinBoxModel.

spireken: To answer the question more directly: a valid value is valid with respect to the model, and a valid input is valid with respect to the UX, but may or may not be acceptable as a submitted value.

spireken:
> What happens if you remove the `std::int64_t` and express `value.compare(1)`?

The compiler can't differentiate between these types for a suitable overloaded method to call: long double, boost::ulong_long_type, and boost::long_long_type.

kranar:
That's so sad, oh well, keep as is then.

spireken:
Are there any more changes for this PR?

Applications/Spire/Include/Spire/Ui/RealSpinBox.hpp
```diff
@@ -98,7 +98,8 @@ namespace Spire {
       QString display_string(Real value);
       QStyle::SubControl get_current_control(const QPoint& mouse_pos);
       boost::optional<Real> get_value(const QString& text) const;
-      bool is_valid(const QString& string);
+      bool is_valid_input(const QString& string) const;
```
kranar, line 101: The technical term you're looking for is "well formed" which briefly speaking means that the input is syntactically valid but may or may not be semantically valid.

Hence rename ```is_valid_input``` to ```is_well_formed``` and ```is_valid_value``` to ```is_valid```.

PR #512, opened 2020-10-13
Flat button test widget

spireken:
Along with the UiViewer demo I updated the Book View and Time and Sales demos for the CheckBox updates. The CheckBox creation code was basically copy-pasted everywhere it was used.


spireken:
I forgot the constructor test.

Applications/Spire/Include/Spire/Ui/FlatButton.hpp
```diff
@@ -34,17 +34,14 @@ namespace Spire {
...
+        \param label The label text.
         \param parent The parent widget to the FlatButton.
       */
       explicit FlatButton(const QString& label, QWidget* parent = nullptr);
 
       //! Sets the FlatButton's text.
       /*!
-        \param text The text to set. If the text is empty, the button is
-                    considered a color button and is drawn with white padding.
+        \param text The text to set.
```
kranar, line 44: The name of the parameter should be the same as the noun, ie. rename ```text``` to ```label```.

Applications/Spire/Source/UiViewer/FlatButtonTestWidget.cpp
```diff
@@ -0,0 +1,84 @@
...
+using namespace Spire;
+
+namespace {
+  auto BUTTON_SIZE() {
+    static auto size = scale(100, 26);
+    return size;
+  }
+
+  auto make_labeled_button(const QString& label, QWidget* parent) {
+    auto button = new FlatButton(label, parent);
```
kranar, line 15: Why is so much work being done in this function instead of in the constructor for the ```FlatButton```?

kranar (changes requested):

Applications/Spire/Source/UiViewer/FlatButtonTestWidget.cpp
spireken: So far we have a couple of different styles for FlatButton so the default style isn't really set. I could make the style in the spec the default that would be set in the constructor.

kranar: For this task, only implement the button specified in the ui_kit/button.xd. If no class exists implementing that component then write a new one for it.

spireken: Or have a make_flat_button(...) function like the I did for the CheckBox.

kranar: Yes that works too.

kranar (approved):

PR #525, opened 2020-10-22
Security test widget

spireken:


Applications/Spire/Source/UiViewer/SecurityWidgetTestWidget.cpp
```diff
@@ -0,0 +1,71 @@
...
+  m_layout->addWidget(info_label, 0, 0, 1, 2);
+  m_dark_theme_check_box = make_check_box(tr("Dark Theme"), this);
+  m_layout->addWidget(m_dark_theme_check_box, 1, 0);
+  auto reset_button = make_flat_button(tr("Reset"), this);
+  reset_button->setFixedSize(scale(100, 26));
+  reset_button->connect_clicked_signal([=] { on_reset_button(); });
+  m_layout->addWidget(reset_button, 1, 1);
+  m_status_label = new QLabel(this);
+  m_layout->addWidget(m_status_label, 2, 0, 1, 2);
+  m_model.add(SecurityInfo(
```
kranar, line 26: Write a function to return test model that is used across all of these widgets. Include that function in ```Spire/SecurityInput/LocalSecurityInputModel.hpp```

spireken: Do you want this applied to every widget, or just the test widgets?

kranar: Do what you think is best.

Applications/Spire/Include/Spire/SecurityInput/LocalSecurityInputModel.hpp
```diff
@@ -25,6 +26,8 @@ namespace Spire {
     private:
       rtv::Trie<char, Nexus::SecurityInfo> m_securities;
   };
+
+  Beam::Ref<const LocalSecurityInputModel> get_local_security_input_test_model();
```
kranar, line 30: ```Ref``` is used as a parameter to indicate lack of ownership, similar to things like ```std::string_view```. It should not be used as a return type.

kranar:
The model can not be marked ```const``` so it's fine as is.

spireken:
Aside from the creation of the model's SecurityInfo instances which were copied, I had to write this function from scratch.

As far as I can tell it implements a solution where the model will be instantiated once, have its lifetime managed implicitly by it being static, while still being able to be returned as a reference from a free function. It would be nice if it could be const, but given the restrictions on the other APIs and knowing that it's typically not going to be shared this implementation is a fair compromise without changing other APIs.

PR #528, opened 2020-10-23
Static test widget

spireken:


spireken:
The menu items can be set or the menu widget reset completely with a list of CSV values from the input.

PR #534, opened 2020-10-30
Date test widget

spireken:


Applications/Spire/Include/Spire/Ui/DateInputWidget.hpp
```diff
@@ -10,6 +10,12 @@ namespace Spire {
   class DateInputWidget : public QLabel {
     public:
 
+      //! Signal type for date selection changes.
+      /*!
+        \param date The selected date.
+      */
+      using SelectedSignal = Signal<void (boost::posix_time::ptime date)>;
```
kranar, line 17: For a date, can you stick to using ```boost::gregorian::date```?

kranar (changes requested):

Applications/Spire/Include/Spire/Ui/DateInputWidget.hpp
spireken: I think that a date is more appropriate, but if I recall originally the requirement was to use a ptime because that's what the APIs interacting with this DateInputWidget would expect.

kranar: That is entirely possible, but go with ```boost::gregorian::date```.

spireken:
Can I push an amended commit to fix the commit message?

kranar:
Yes but it's almost never worth it.

PR #542, opened 2020-11-04
Key bindings combo boxes

spireken:
I originally named this branch key_bindings_combo_boxes because I thought the PR would be minimal, but these changes are currenly only to get the OrderTypeComboBox working in the Key Bindings task keys table. If this gets approved I'll submit another (or multple) PR for the rest of the combo boxes/delegates.


Applications/Spire/Include/Spire/Ui/OrderTypeComboBox.hpp
```diff
@@ -16,12 +17,25 @@ namespace Spire {
...
         \param parent The parent widget.
       */
       explicit OrderTypeComboBox(QWidget* parent = nullptr);
 
+      //! Returns the currently selected OrderType.
+      Nexus::OrderType get_current_order_type() const;
+
+      //! Returns the currently activated OrderType, or the currently selected
+      //! OrderType if there is no currently activated OrderType.
+      Nexus::OrderType get_last_order_type() const;
```
kranar, line 31: This method looks suspicious, can you go into some detail why you have it?

Applications/Spire/Include/Spire/Ui/StaticDropDownMenu.hpp
```diff
@@ -62,10 +72,26 @@ namespace Spire {
       //! Returns the currently selected item.
       virtual QVariant get_current_item() const;
 
+      //! Returns the currently activated item, or the currently selected item
+      //! if there is no currently activated item.
+      virtual QVariant get_last_item() const;
```
kranar, line 77: Same here, don't like the look of this method.

Applications/Spire/Include/Spire/Ui/OrderTypeComboBox.hpp
spireken: The StaticDropDownMenu triggers a selection signal when it loses focus, and if there is an item that was activated (using up/down or charcter key press), that item gets submitted as the selected item. The KeyBindingItemDelegate closes the editor (combo box) when the tab key is pressed, so the editor doesn't confirm the currently activated item when it loses focus.

This method provides a way to determine the intent of the user when the editor is closed. Either they explicitly selected an item and the function will return the current item, or the function will return the last activated item when the editor is closed.

An alternative would be to have the get_current_item() behave like the get_last_item(), but there may be inconsistencies if get_current_item() is called while the user is currently selecting a value. In this case, get_current_item() would return the currently activated item even though the user hasn't confirmed it.

Another alternative, assuming the editor is closed before setModelData is called, would be to have the StaticDropDownMenu's last activated item committed on a closeEvent, if the menu has a cell style.

kranar: Okay one issue is a matter of terminology, a ComboBox has a current and an active item.

When a ComboBox loses focus, the active item becomes the current item.

spireken: That's true, but from a technical perspective when the editor is closed it doesn't receive a focus out event, so the active item doesn't get confirmed as the current item.

Without modifying the API, my second suggestion would probably be the best way to fix this. To prevent potential issues with both a focus out and close event triggering a double selected signal, the active item would become the current item on a close event only when the ComboBox has the cell style applied to it.

kranar: Okay, when a ComboBox is closed, the active item becomes the current item as well. Will that help fix it?

spireken: After testing, the delegate's setModelData is called before the call to the ComboBox's closeEvent.

kranar: That's good news right?

kranar: Oh shoot that's bad news isn't it?

kranar: At any rate, feel free to think up an alternative solution and if you don't think there's a reasonable solution, then go with your proposal of having ```get_current_item``` behave like ```get_last_item```.

spireken: I'm looking into other events to see if one of them can be used as a trigger to update the current item. Ideally I'd remove get_last_item() and just have get_current_item() as it currently is, but if I can't find a way to encapsulate this in the ComboBox then I'll update get_current_item() to work like get_last_item().

kranar: Agood... ... agood...

spireken: Looking into this, the editor is closed entirely within the signal/slot system and the events for the Qt event system aren't sent until after the data is committed.

I don't see a way to implement this aside from updating get_current_item() to use the last active item.

kranar: Go for it.

kranar:
My understanding is that this PR should be closed until the input fields are changed to no longer enter "edit mode" on focus. Let me know if that's correct.

spireken:
This branch is to add the first combo box into the table, and I updated the base delegate to simplify the editing logic. I would argue that the tab edit mode should be addressed in another case or a separate branch on this case ('Update Windows to Use New Drop Down Classes").

The tab edit changes should be relatively straightforward, but it adds more things to test in this branch.

Applications/Spire/Include/Spire/Ui/StaticDropDownMenu.hpp
```diff
@@ -1,17 +1,27 @@
...
 
 namespace Spire {
 
   //! Represents a DropDownMenu that isn't modified by user interactions.
-  class StaticDropDownMenu : public QWidget {
+  class StaticDropDownMenu : public QLineEdit {
     public:
 
+      //! The styles available to render the StaticDropDownMenu.
+      enum class Style {
```
kranar, line 16: Look into removing this, widgets should be rendered independently of whether they are part of a table or not.

spireken: I can look into manipulating the editor geometry outside of the bounds of the current cell, but if that doesn't work I'm not sure what else I could do aside from some impractical workaround.

kranar: Submit a demo using only the default style so I can see what the issue is.

spireken: I updated the OrderType column to use the default style. Along with the border issue, another issue is that the combo box needs to know if it's in a cell, and if it is then it needs to show the list when it's displayed.

spireken: So I can manipulate the editor geometry and it looks like the cell's padding was updated to 8px, so there should be no need to have a custom cell style.

I'll still need a way to set the list as always visible for cell combo boxes, so I'll probably have to add at least a constructor that takes that as a boolean.

spireken:
I think these changes will work nicely. The other columns are using the InputFieldEditor, which is being removed, so the placement of their editors will be off until they're updated to use the combo boxes. The key sequence column's editor will also be off until it's updated to use the KeySequenceInputField.

These changes affect the Name column's editor, so I removed the cell style from the TextInputWidget in this PR to have it work correctly as an editor.

PR #547, opened 2020-11-16
Icon button updates

spireken:
The updated icon buttons (aside from the UiViewer test widget) are in the ToolbarWindow, KeyBindingsTableView (delete row button), LoginWindow (close button), and the title bar buttons.

The ChartWindow had it's toggle buttons update, too.

The builds are in a folder called 'icon_updates', in the demos folder.


Applications/Spire/Include/Spire/Ui/IconButton.hpp
```diff
@@ -22,48 +22,41 @@ namespace Spire {
...
-      //! Constructs an IconButton with a default icon and a hover icon.
+      //! Sets the icon to display.
       /*!
-        \param icon The icon shown when the button is not hovered.
-        \param hover_icon The icon shown when the button is hovered.
-        \param parent The parent QWidget to the IconButton.
+        \param icon The icon image.
       */
-      IconButton(QImage icon, QImage hover_icon, QWidget* parent = nullptr);
+      void set_icon(const QImage& icon);
```
kranar, line 29: Why is this setter here instead of using the constructor?

Applications/Spire/Include/Spire/Ui/IconButton.hpp
```diff
@@ -22,48 +22,41 @@ namespace Spire {
...
+        \param color The hover color.
       */
-      void set_hover_style(const QString& stylesheet);
-
-      //! Returns the icon displayed.
-      const QImage& get_icon() const;
-
-      //! Sets the icon to display.
-      void set_icon(QImage icon);
+      void set_hover_color(const QColor& color);
```
kranar, line 47: Minimize the number of setters, instead have a constructor that takes these values and even better is to have a Styles struct that defines these along with sensible default values.

    struct Properties {
      QColor m_backgroundColor;
      QColor m_hoverColor;
      QColor m_blurColor;
    };

kranar (changes requested):

Applications/Spire/Include/Spire/Ui/IconButton.hpp
spireken: This is used in the case of the TitleBar, which is constructed with a default Spire icon but can later be replaced.

kranar: If you want to replace it, then construct a new ```IconButton```.

spireken: What's the benefit of replacing the button when all that's really required is to replace the icon?

kranar: Fair question, the answer is to prefer a small API or at the very least, an API with as few setters as possible.

Applications/Spire/Source/Ui/TitleBar.cpp
```diff
@@ -20,31 +20,33 @@ namespace {
...
-  auto apply_button_style(IconButton* button) {
-    button->setFocusPolicy(Qt::FocusPolicy::NoFocus);
-    auto button_style = button->get_style();
-    button_style.m_default_color = "#333333";
-    button_style.m_hover_color = "#333333";
-    button_style.m_blur_color = "#D0D0D0";
-    button->set_style(button_style);
-    button->setFixedSize(BUTTON_SIZE());
+  auto BUTTON_STYLE() {
+    static auto style = IconButton::Style();
```
kranar, line 24: Either make this ```static``` and return a ```const&``` or return a value.

Applications/Spire/Source/Ui/ToggleButton.cpp
```diff
@@ -56,23 +53,33 @@ connection ToggleButton::connect_clicked_signal(
...
-    button_style.m_default_color = "#1FD37A";
-    button_style.m_hover_color = "#1FD37A";
-  } else {
-    button_style.m_default_color = "#7F5EEC";
-    button_style.m_hover_color = "#4B23A0";
-  }
-  m_icon_button->set_style(button_style);
+void ToggleButton::update_button() {
+  delete_later(m_icon_button);
+  auto style = [&] {
```
kranar, line 61: Remove IIFE, it doesn't provide any benefit here.

kranar (changes requested):

Applications/Spire/Include/Spire/Ui/ToggleButton.hpp
```diff
@@ -15,51 +15,24 @@ namespace Spire {
...
-        \param toggle_icon The icon shown when the button is in a toggled
-                            state.
-        \param hover_icon The icon shown when the button is hovered or receives
-                          keyboard focus.
-        \param disable_icon The icon shown when the button is disabled.
-        \param parent The parent QWidget to the button
-      */
-      ToggleButton(QImage icon, QImage toggle_icon, QImage hover_icon,
-        QImage disabled_icon, QWidget* parent = nullptr);
+      ToggleButton(QImage icon, QWidget* parent = nullptr);
```
kranar, line 21: ```explicit```

kranar (changes requested):

PR #549, opened 2020-11-20
Drop down updates

spireken:
I forgot about this branch, so conveniently these changes were already implemented. I think I originally meant to add more changes so it's called 'drop_down_updates', but it only updates the ColorSelectorButton to use the new DropDownWindow. 

The UiViewer shows that the space and down keys will show the drop down window for the ColorSelectorButton.


Applications/Spire/Include/Spire/Ui/ColorSelectorButton.hpp
```diff
@@ -39,18 +39,14 @@ namespace Spire {
 
     protected:
       bool eventFilter(QObject* watched, QEvent* event) override;
```
kranar, line 41: Make constructor  ```explicit```.

kranar (changes requested):

kranar (approved):

PR #550, opened 2020-11-23
Trend line list updates

spireken:
I added a ChartingUiTester-TrendList demo.


Applications/Spire/Include/Spire/Ui/DropDownItem.hpp
```diff
@@ -38,24 +38,26 @@ namespace Spire {
       */
       DropDownItem(QVariant value, QImage icon, QWidget* parent = nullptr);
 
+      virtual ~DropDownItem() = default;
```
kranar, line 41: Remove this.

Applications/Spire/Include/Spire/Ui/StaticDropDownMenu.hpp
```diff
@@ -41,6 +41,8 @@ namespace Spire {
       StaticDropDownMenu(std::vector<QVariant> items,
         const QString& display_text, QWidget* parent = nullptr);
 
+      virtual ~StaticDropDownMenu() = default;
```
kranar, line 44: Why was this added?

kranar (changes requested):

Applications/Spire/Include/Spire/Ui/StaticDropDownMenu.hpp
spireken: Right, forgot about the base class' virtual destructor.

Applications/Spire/Include/Spire/Ui/DropDownItem.hpp
```diff
@@ -39,23 +39,23 @@ namespace Spire {
...
+      virtual void reset_highlight();
 
       //! Connects a slot to the highlighted signal.
-      boost::signals2::connection connect_highlighted_signal(
+      virtual boost::signals2::connection connect_highlighted_signal(
         const HighlightedSignal::slot_type& slot) const;
       
       //! Connects a slot to the selected signal.
-      boost::signals2::connection connect_selected_signal(
+      virtual boost::signals2::connection connect_selected_signal(
```
kranar, line 58: Why are these now virtual? In general I do not like to see changes to an API that have not been discussed/approved.

Applications/Spire/Source/Charting/StyleDropDownMenuItem.cpp
```diff
@@ -1,69 +1,40 @@
...
+StyleDropDownMenuItem::StyleDropDownMenuItem(TrendLineStyle style,
+  QWidget* parent)
+  : DropDownItem(QVariant::fromValue(style), parent),
+    m_style(style) {}
 
-void StyleDropDownMenuItem::mouseReleaseEvent(QMouseEvent* event) {
-  if(event->button() == Qt::LeftButton) {
-    m_selected_signal(m_style);
-  }
+// TODO: still required?
```
kranar, line 25: Resolve this TODO

kranar (changes requested):

Applications/Spire/Include/Spire/Ui/DropDownItem.hpp
spireken: I subclass the DropDownItem for items that require custom widgets or painting. Doesn't that require the base class to have virtual methods when used polymorphically?

spireken: I would argue they should have been virtual to begin with, but the other widget that subclasses the DropDownItem, SecurityInfoItem, doesn't override any of the base class methods (aside from the protected Qt event functions).

kranar: Don't need these to be ```virtual```.

spireken: Why?

kranar: The default is non-```virtual```, if method is declared to be ```virtual``` there needs to be a justification but I don't see one.

spireken: So is the standard we declare functions as virtual on an as-needed basis?

kranar: Yep! That's exactly it.

kranar (approved):

PR #551, opened 2020-11-23
Color selector updates

spireken:
I updated the UiViewer demo.

### DIFF
--- a/Applications/Spire/Include/Spire/Ui/ColorSelectorHexInputWidget.hpp
+++ b/Applications/Spire/Include/Spire/Ui/ColorSelectorHexInputWidget.hpp
@@ -28,15 +28,20 @@ namespace Spire {
       */
       void set_color(const QColor& color);
 
-      //! Connects a slot to the color signal.
-      boost::signals2::connection connect_color_signal(
+      //! Connects a slot to the color modified signal.
+      boost::signals2::connection connect_modified_signal(
+        const ColorSignal::slot_type& slot) const;
+
+      //! Connects a slot to the color selection signal.
+      boost::signals2::connection connect_selected_signal(
         const ColorSignal::slot_type& slot) const;
 
     protected:
       bool eventFilter(QObject* watched, QEvent* event) override;
 
     private:
-      mutable ColorSignal m_color_signal;
+      mutable ColorSignal m_modified_signal;
+      mutable ColorSignal m_selected_signal;
       TextInputWidget* m_text_input;
       QString m_color_name;
 
--- a/Applications/Spire/Include/Spire/Ui/ColorSelectorHexInputWidget.hpp
+++ b/Applications/Spire/Include/Spire/Ui/ColorSelectorHexInputWidget.hpp
@@ -28,8 +28,8 @@ namespace Spire {
       */
       void set_color(const QColor& color);
 
-      //! Connects a slot to the color modified signal.
-      boost::signals2::connection connect_modified_signal(
+      //! Connects a slot to the color changed signal.
+      boost::signals2::connection connect_changed_signal(
         const ColorSignal::slot_type& slot) const;
 
       //! Connects a slot to the color selection signal.
@@ -40,7 +40,7 @@ namespace Spire {
       bool eventFilter(QObject* watched, QEvent* event) override;
 
     private:
-      mutable ColorSignal m_modified_signal;
+      mutable ColorSignal m_changed_signal;
       mutable ColorSignal m_selected_signal;
       TextInputWidget* m_text_input;
       QString m_color_name;
--- a/Applications/Spire/Source/Ui/ColorSelectorButton.cpp
+++ b/Applications/Spire/Source/Ui/ColorSelectorButton.cpp
@@ -13,15 +13,19 @@ ColorSelectorButton::ColorSelectorButton(const QColor& current_color,
   setFocusPolicy(Qt::StrongFocus);
   setAttribute(Qt::WA_Hover);
   m_selector_widget = new ColorSelectorDropDown(current_color, this);
-  m_color_connection = m_selector_widget->connect_color_signal(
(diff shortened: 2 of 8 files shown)


Applications/Spire/Include/Spire/Ui/ColorSelectorHexInputWidget.hpp
```diff
@@ -28,15 +28,20 @@ namespace Spire {
       */
       void set_color(const QColor& color);
 
-      //! Connects a slot to the color signal.
-      boost::signals2::connection connect_color_signal(
+      //! Connects a slot to the color modified signal.
```
kranar, line 31: I don't know of any signals that use the word modified. Can you look to rename this signal to bring it more in line with how Qt names it signals?

Applications/Spire/Source/Ui/ColorSelectorButton.cpp
```diff
@@ -44,6 +45,13 @@ bool ColorSelectorButton::eventFilter(QObject* watched, QEvent* event) {
       m_selector_widget->activateWindow();
       m_selector_widget->setFocus();
     }
+  } else if(watched == this) {
+    if(event->type() == QEvent::KeyPress) {
+      auto e = static_cast<QKeyEvent*>(event);
+      if(e->key() == Qt::Key_Down) {
```
kranar, line 51: What is this section used for?

Applications/Spire/Source/Ui/ColorSelectorDropDown.cpp
```diff
@@ -143,10 +143,11 @@ ColorSelectorDropDown::ColorSelectorDropDown(const QColor& current_color,
...
-  m_hex_input->connect_color_signal([=] (const auto& color) {
-    m_color_value_slider->set_color(color);
-    m_color_hue_slider->set_color(color);
-    on_color_selected(color);
+  m_hex_input->connect_modified_signal([=] (const auto& color) {
+    on_color_hex_updated(color);
+  });
+  m_hex_input->connect_selected_signal([=] (const auto& color) {
+    on_color_hex_updated(color);
     window()->hide();
```
kranar, line 151: Avoid having windows hide or show themselves. Sometimes it's necessary if the window is a top-level window but when you can avoid doing it it's preferable. Instead have the button hide the window when it receives a selected signal.

Applications/Spire/Source/Ui/ColorSelectorHexInputWidget.cpp
```diff
@@ -61,22 +61,28 @@ void ColorSelectorHexInputWidget::set_color(const QColor& color) {
...
-    if(e->key() == Qt::Key_Enter || e->key() == Qt::Key_Return) {
+    if(e->key() == Qt::Key_Enter || e->key() == Qt::Key_Return ||
+        e->key() == Qt::Key_Space) {
       auto text = m_text_input->text();
       if(text.length() == 6 || text.length() == 3) {
         auto color = QColor();
         color.setNamedColor(QString("#%1").arg(m_text_input->text()));
         m_color_name = color_name(color);
-        m_color_signal(color);
+        m_selected_signal(color);
```
kranar, line 85: Only signal once you've committed all state. A signal introduces re-entrancy so any state not committed before the signal can end up being inconsistent afterwards.

Applications/Spire/Source/Ui/FlatButton.cpp
```diff
@@ -155,21 +157,30 @@ void FlatButton::enable_button() {
 }
 
 QString FlatButton::get_stylesheet_properties(const Style& s) {
+  auto [margin, border_color] = [&] {
+    if(m_label->text().isEmpty()) {
+      return std::make_pair(2, QColor(Qt::transparent));
```
kranar, line 162: Use ```std::pair```

kranar (changes requested):

Applications/Spire/Include/Spire/Ui/ColorSelectorHexInputWidget.hpp
spireken: Qt uses 'edited' for user changes, and 'changed' for both user changes and changes made programmatically. I used modified for the Key Bindings table signals, so I can update those as well.

Applications/Spire/Source/Ui/ColorSelectorButton.cpp
spireken: The default behavior of the DropDownWindow is to show itself when its parent receives a down key press. The expectation was that most widgets that use a DropDownWindow would have the down key as an activation option. If the down key is reserved for the list I can move the down key event filter to the list and remove this code.

kranar: Oh I see, okay keep this then.

kranar: Also if that's the case then feel free to have the down arrow key also open up the color window as well.

spireken: Should I ask Darryl to update the user flow?

kranar: Yep, go for it.

Applications/Spire/Include/Spire/Ui/ColorSelectorHexInputWidget.hpp
spireken: I think I'll go with Changed because it's more generic.

Applications/Spire/Source/Ui/ColorSelectorDropDown.cpp
spireken: I'll have to add a Changed signal to the ColorSelectorDropDown. The drop down is what differentiates between change and a selection signal, so it was more straightforward to close the window from there.

PR #559, opened 2020-12-15
Scroll bar fix

spireken:


Applications/Spire/Source/Ui/ScrollBarStyle.cpp
```diff
@@ -0,0 +1,152 @@
...
+ScrollBarStyle::ScrollBarStyle(QWidget* parent)
+    : m_horizontal_handle_height(scale_height(13)),
+      m_vertical_handle_width(scale_width(13)) {
+  setParent(parent);
+}
+
+void ScrollBarStyle::drawComplexControl(QStyle::ComplexControl control,
+    const QStyleOptionComplex* option, QPainter* painter,
+    const QWidget *widget) const {
+  if(control == CC_ScrollBar) {
```
kranar, line 29: ```QStyle::CC_ScrollBar```

Applications/Spire/Source/Ui/ScrollBarStyle.cpp
```diff
@@ -0,0 +1,152 @@
...
+    painter->fillRect(option->rect, Qt::white);
+    auto slider_option = QStyleOptionSlider();
+    slider_option.rect = subControlRect(CC_ScrollBar, &slider_option,
```
kranar, line 32: Same here.

Applications/Spire/Source/Ui/ScrollBarStyle.cpp
```diff
@@ -0,0 +1,152 @@
...
+      SC_ScrollBarSlider, widget);
```
kranar, line 33: Scope all the enums explicitly.

Applications/Spire/Source/Ui/ScrollBarStyle.cpp
```diff
@@ -0,0 +1,152 @@
...
+    drawControl(CE_ScrollBarSlider, &slider_option, painter, widget);
+    return;
+  }
+  QProxyStyle::drawComplexControl(control, option, painter, widget);
+}
+
+void ScrollBarStyle::drawControl(QStyle::ControlElement element,
+    const QStyleOption* option, QPainter* painter,
+    const QWidget *widget) const {
+  if(element == CE_ScrollBarSlider) {
```
kranar, line 43: See above about scoping enums.

Applications/Spire/Source/Ui/ScrollBarStyle.cpp
```diff
@@ -0,0 +1,152 @@
...
+  }
+  return QProxyStyle::subControlRect(control, option, sub_control, widget);
+}
+
+void ScrollBarStyle::set_horizontal_scroll_bar_height(int height) {
+  m_horizontal_handle_height = height;
+}
+
+void ScrollBarStyle::set_vertical_scroll_bar_width(int width) {
+  m_vertical_handle_width = width;
```
kranar, line 124: Does an update need to be performed? I'm concerned that a state is being changed here but no associated side effect is produced. This could be fine but it looks fishy.

kranar (changes requested):
Looks like a decent improvement!

Applications/Spire/Source/Ui/ScrollBarStyle.cpp
spireken: It's possible that we could set a style on an existing scroll bar (like a QListWidget's scroll bar) and we would have to call update() in that case.

In practice only the ScrollArea calls these methods so this isn't an issue because the ScrollArea is calling update() after calling these, but, to keep it working predictably it's probably best to have the style call update on the given parent QWidget.

kranar: No it's fine then, keep it as is.

kranar (approved):

PR #561, opened 2020-12-16
UI viewer fixes2

spireken:
I added a UiViewer-TabOrderFix demo. This fixes the issue where resetting a widget would distort the expected tab order.

If there was more than one widget that would have to be declared in the header, then I just declared a single lambda to capture the widgets in the constructor instead.


Applications/Spire/Source/UiViewer/ColorSelectorButtonTestWidget.cpp
```diff
@@ -44,6 +44,13 @@ ColorSelectorButtonTestWidget::ColorSelectorButtonTestWidget(QWidget* parent)
   create_color_button->connect_clicked_signal([=] {
     on_create_color_button_color(create_color_input->text());
   });
+  m_reset_tab_order = [=] () {
```
kranar, line 47: Why is this a lambda expression instead of added to ```on_create_color_button_color```?

spireken: Missed it at the time.

spireken: I would have to duplicate the tab order code in this case, once in the lambda and again at the end of the constructor.

kranar: Then use a member function.

kranar (approved):

PR #567, opened 2021-01-05
Restored previously removed event filter for TimeAndSalesTableView th…

spireken:
I added a TimeAndSalesUiTester-FlickerFix demo. The TimeAndSalesUiTester-CursorResizeRegion demo can be used to see the original flicker issue.

To reiterate the issue:

1. Run the TimeAndSalesUiTester-CursorResizeRegion demo.
2. Type 'm' and select a security.
3. Drag the vertical scroll bar up and down quickly.
4. The table's header will disappear and reappear, causing a 'flicker' effect.

The more data there is in the table, the more obvious the issue becomes, so it helps to scroll to the bottom of the table a few times first to have the table load more test data.

### DIFF
--- a/Applications/Spire/Source/TimeAndSales/TimeAndSalesTableView.cpp
+++ b/Applications/Spire/Source/TimeAndSales/TimeAndSalesTableView.cpp
@@ -90,7 +90,6 @@ TimeAndSalesTableView::TimeAndSalesTableView(QWidget* parent)
     })");
   m_table->setItemDelegate(new ItemPaddingDelegate(scale_width(5),
     new CustomVariantItemDelegate(), this));
-  m_table->installEventFilter(this);
   m_layout->addWidget(m_table);
   setWidget(main_widget);
 }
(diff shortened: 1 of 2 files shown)


Applications/Spire/Source/TimeAndSales/TimeAndSalesTableView.cpp
```diff
@@ -131,6 +132,13 @@ void TimeAndSalesTableView::set_properties(
   }
 }
 
+bool TimeAndSalesTableView::eventFilter(QObject* watched, QEvent* event) {
+  if(watched == m_table && event->type() == QEvent::Paint) {
+    m_header->viewport()->update();
```
kranar, line 137: This might work but I'd prefer to avoid using event filters as much as possible. When does a table repaint itself? Even better would be to determine when does a table repainting require the header's viewport to be updated. Is it really on every single table paint event or is there perhaps another causal relationship you can make use of that doesn't require an event filter, for example a when a row is added, removed, or the table is scrolled.

Applications/Spire/Source/TimeAndSales/TimeAndSalesTableView.cpp
```diff
@@ -215,6 +207,7 @@ void TimeAndSalesTableView::on_vertical_slider_value_changed(
   }
   m_model->set_row_visible(m_table->rowAt(
     m_table->visibleRegion().boundingRect().bottom()));
+  m_header->viewport()->update();
```
kranar, line 210: Oh much better and it makes a lot more sense, it directly ties the visual update to a change in the data model.

PR #570, opened 2021-01-07
Tab bar padding

spireken:
I added TabWidgetPadding demos for the Key Bindings and Book View (properties) windows.

### DIFF
--- a/Applications/Spire/Include/Spire/Ui/TabWidget.hpp
+++ b/Applications/Spire/Include/Spire/Ui/TabWidget.hpp
@@ -9,17 +9,17 @@ namespace Spire {
   class TabWidget : public QTabWidget {
     public:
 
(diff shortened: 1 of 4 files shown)


Applications/Spire/Include/Spire/Ui/TabWidget.hpp
```diff
@@ -9,12 +9,30 @@ namespace Spire {
...
     public:
 
-      //! Constructs a TabWidget.
+      //! Represents the supported tab content padding styles.
+      enum class PaddingStyle {
+
+        //! The content has no padding.
+        NONE,
+
+        //! The default content padding.
```
kranar, line 18: Don't name an enum ```DEFAULT```, instead give it a name ```FOO``` that describes what it does and then the default value (for example in the constructor) can be use ```FOO```.

PR #574, opened 2021-01-22
Toggle button updates

spireken:
I merged the Icon and Toggle buttons, so the IconButton's 'checkable' property can be tested in the UiViewer in place of the ToggleButton.


kranar (approved):

Applications/Spire/Source/UiViewer/StandardUiProfiles.cpp
```diff
@@ -126,16 +126,30 @@ UiProfile Spire::make_icon_button_profile() {
...
       auto& tooltip = get<QString>("tooltip", profile.get_properties());
       button->setToolTip(tooltip.get());
-      apply_widget_properties(button, profile.get_properties());
       tooltip.connect_changed_signal([=] (const auto& value) {
         button->setToolTip(value);
       });
+      auto& checkable = get<bool>("checkable", profile.get_properties());
+      checkable.connect_changed_signal([=] (auto is_checkable) {
+        button->setCheckable(is_checkable);
+        button->update();
```
shaobinspire, line 144: When the checkable is set to false, the checked which is true should be reset to false.

spireken: @kranar recommended in the case notes that I don't add too much complexity to the tester: "You can program any arbitrary logic you want, but I would recommend against it."

It's a small inconsistency with the tester, but it's not caused by an issue with the button itself.

shaobinspire: I see.

shaobinspire (approved):

PR #575, opened 2021-01-26
Tooltip

spireken:
In order to apply the UiViewer's test parameters, the 'Rebuild' button must be pressed because the changes aren't reflected in the test Tooltip automatically.


kranar:
Fix the following bug:

1. Place top-left of the mouse cursor at the bottom pixel of the "Hover me!" label.
2. Wait for a few milliseconds.
3. Expected behavior: Tooltip pops up and remains visible until the mouse is no longer hovering over the label.
4.  Actual behavior: Tooltip pops up and disappears repeatedly, resulting in flickering.

PR #576, opened 2021-01-27
Text box

shaobinspire:
The new textbox is added submit and current signals, and it also truncates the text with ellipses when it is read-only or not on focus.

### DIFF
--- a/Applications/Spire/Include/Spire/Ui/TextBox.hpp
+++ b/Applications/Spire/Include/Spire/Ui/TextBox.hpp
@@ -29,7 +29,7 @@ namespace Spire {
       explicit TextBox(const QString& text, QWidget* parent = nullptr);
 
       //! Gets the text.
-      QString get_text() const;
+      const QString& get_text() const;
 
       //! Sets the text.
       void set_text(const QString& text);
--- a/Applications/Spire/Include/Spire/Ui/TextBox.hpp
+++ b/Applications/Spire/Include/Spire/Ui/TextBox.hpp
@@ -54,6 +54,7 @@ namespace Spire {
       mutable CurrentSignal m_current_signal;
       mutable SubmitSignal m_submit_signal;
       QString m_text;
+      QString m_submitted_text;
 
       QString text() const = delete;
       void setText(const QString&) = delete;
--- a/Applications/Spire/Include/Spire/Ui/TextBox.hpp
+++ b/Applications/Spire/Include/Spire/Ui/TextBox.hpp
@@ -1,0 +1,67 @@
+#ifndef SPIRE_TEXT_BOX_HPP
+#define SPIRE_TEXT_BOX_HPP
+#include <QLineEdit>
+#include "Spire/Ui/Ui.hpp"
+
+namespace Spire {
+
+  //! Displays a one-line text box.
+  class TextBox : public QLineEdit {
+    public:
+
+      //! Signals that the current text changed.
+      using CurrentSignal = Signal<void (const QString& text)>;
+
+      //! Signals that the text is submitted.
+      using SubmitSignal = Signal<void (const QString& text)>;
+
+      //! Constructs a text box without a default text.
+      /*!
+        \param parent The parent widget.
+      */
+      explicit TextBox(QWidget* parent = nullptr);
+
+      //! Constructs a text box with a default text.
+      /*!
+        \param text The text to display.
+        \param parent The parent widget.
+      */
+      explicit TextBox(const QString& text, QWidget* parent = nullptr);
+
+      //! Gets the text.
+      QString get_text() const;
+
+      //! Sets the text.
+      void set_text(const QString& text);
+
+      //! Connects a slot to the current signal.
+      boost::signals2::connection connect_current_signal(
+        const CurrentSignal::slot_type& slot) const;
+
+      //! Connects a slot to the submit signal.
+      boost::signals2::connection connect_submit_signal(
+        const SubmitSignal::slot_type& slot) const;
+
+    protected:
+      void changeEvent(QEvent* event) override;
+      void focusInEvent(QFocusEvent* event) override;
+      void focusOutEvent(QFocusEvent* event) override;
+      void keyPressEvent(QKeyEvent* event) override;
+      void resizeEvent(QResizeEvent* event) override;
+      QSize sizeHint() const override;
+
+    private:
+      mutable CurrentSignal m_current_signal;
+      mutable SubmitSignal m_submit_signal;
+      QString m_text;
+
+      QString text() const = delete;
+      void setText(const QString&) = delete;
+      void on_editing_finished();
+      void on_selection_changed();
+      void on_text_edited(const QString& text);
+      void elide_text();
+  };
+}
+
+#endif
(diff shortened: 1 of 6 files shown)


spireken (changes requested):

Applications/Spire/Include/Spire/Ui/TextBox.hpp
```diff
@@ -0,0 +1,68 @@
...
+#include "Spire/Ui/Ui.hpp"
+
+namespace Spire {
+
+  //! Displays a one-line text box.
+  class TextBox : public QLineEdit {
+    public:
+
+      //! Signals that the current text changed.
+      using CurrentSignal = Signal<void (const QString& text)>;
```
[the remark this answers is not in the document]
shaobinspire: Good question. I thought about this question and tried to do it. The current signal can be replaced by the QLineEdit::textEdited signal, but I can’t find a way to replace the submit signal. Actually, there is a QLineEdit::editingFinished signal which is similar to the submit signal. The reason that I can’t use it directly is that the signal can be emitted when the textbox is read-only. I suspect it’s a Qt’s bug. As a result, I have to filter the signal manually in this particular case. For consistency, I chose the Spire signal for these two signals.  Do you have any good ideas for the editingFinished signal?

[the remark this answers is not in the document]
spireken: If it makes the implementation unnecessarily complicated it's probably fine to leave it as it is.

spireken (approved):

Applications/Spire/Include/Spire/Ui/TextBox.hpp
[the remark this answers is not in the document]
shaobinspire: Why do we have to use boost signals? We can define Qt signals in the class.

Applications/Spire/Source/UiViewer/StandardUiProfiles.cpp
```diff
@@ -142,3 +143,35 @@ UiProfile Spire::make_icon_button_profile() {
     });
   return profile;
 }
+
+UiProfile Spire::make_text_box_profile() {
+  auto properties = std::vector<std::shared_ptr<UiProperty>>();
+  populate_widget_properties(properties);
+  properties.push_back(make_standard_bool_property("read-only"));
+  properties.push_back(make_standard_qstring_property("contents"));
+  properties.push_back(make_standard_qstring_property("placeholder",
```
kranar, line 152: Remove anything that isn't in the spec, so this ```placeholder``` should be removed.

Applications/Spire/Source/UiViewer/StandardUiProfiles.cpp
```diff
@@ -142,3 +143,35 @@ UiProfile Spire::make_icon_button_profile() {
     });
   return profile;
 }
+
+UiProfile Spire::make_text_box_profile() {
+  auto properties = std::vector<std::shared_ptr<UiProperty>>();
+  populate_widget_properties(properties);
+  properties.push_back(make_standard_bool_property("read-only"));
+  properties.push_back(make_standard_qstring_property("contents"));
```
kranar, line 151: Prefer using names that are in the model, so instead of ```contents```, rename this to ```submission```.

Applications/Spire/Source/UiViewer/StandardUiProfiles.cpp
```diff
@@ -142,3 +143,35 @@ UiProfile Spire::make_icon_button_profile() {
     });
   return profile;
 }
+
+UiProfile Spire::make_text_box_profile() {
+  auto properties = std::vector<std::shared_ptr<UiProperty>>();
+  populate_widget_properties(properties);
+  properties.push_back(make_standard_bool_property("read-only"));
```
kranar, line 150: ```read_only```

Applications/Spire/Include/Spire/Ui/TextBox.hpp
```diff
@@ -0,0 +1,68 @@
...
+      void resizeEvent(QResizeEvent* event) override;
+      QSize sizeHint() const override;
+
+    private:
+      mutable CurrentSignal m_current_signal;
+      mutable SubmitSignal m_submit_signal;
+      QString m_text;
+      QString m_submitted_text;
+
+      QString text() const = delete;
```
kranar, line 59: There's no point deleting these, they can be accessed through the base class.

Applications/Spire/Include/Spire/Ui/TextBox.hpp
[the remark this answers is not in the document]
spireken: I don't think we use Qt's meta-object compiler when building Spire, but @kranar will know for sure.

Applications/Spire/Source/Ui/TextBox.cpp
```diff
@@ -0,0 +1,157 @@
...
+using namespace Spire;
+
+TextBox::TextBox(QWidget* parent)
+  : TextBox("", parent) {}
+
+TextBox::TextBox(const QString& text, QWidget* parent)
+    : QLineEdit(text, parent),
+      m_text(text),
+      m_submitted_text(text) {
+  setObjectName("text_box");
```
kranar, line 16: Rename ```text_box``` to ```TextBox```.

Applications/Spire/Source/Ui/TextBox.cpp
```diff
@@ -0,0 +1,157 @@
...
+    #text_box:disabled {
+      background-color: #F5F5F5;
+      border-color: #C8C8C8;
+      color: #C8C8C8;
+    }
+    #text_box:read-only:disabled {
+      background-color: #00000000;
+      border: none;
+    })")
+    .arg(scale_width(1))
```
kranar, line 49: Don't start a new line with an operator.

    ).
    arg(...).
    arg(...).
    arg(...);

Applications/Spire/Include/Spire/Ui/TextBox.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+      //! Constructs a text box with a default text.
+      /*!
+        \param text The text to display.
+        \param parent The parent widget.
+      */
+      explicit TextBox(const QString& text, QWidget* parent = nullptr);
+
+      //! Gets the text.
+      QString get_text() const;
```
kranar, line 32: Return by const reference.

kranar (changes requested):
This is excellent.

Applications/Spire/Include/Spire/Ui/TextBox.hpp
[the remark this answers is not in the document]
kranar: We use Qt because it's basically the only option for cross platform UIs in C++, not because it's particularly good. When there is an option to avoid using Qt to implement a feature, such as the standard library or another third party library, we prefer to use it.

Applications/Spire/Include/Spire/Ui/TextBox.hpp
shaobinspire: I don't want the user to access this function, because calling these two functions can make logical errors. The QLineEdit doesn’t support ellipses by default, and I use setText to change the displayed text in order to display ellipsis when overflow happens. If a user calls QLineEdit::setText, the textbox will lose a chance to elide text; if QLineEdit::text is called, the textbox may return a string with ellipses instead of the original string.

kranar: Okay feel free to leave it then.

kranar (approved):

PR #577, opened 2021-02-01
Checkbox

shaobinspire:
Re-implemented the checkbox according to the new specs.

### DIFF
--- a/Applications/Spire/Source/Ui/Checkbox.cpp
+++ b/Applications/Spire/Source/Ui/Checkbox.cpp
@@ -3,49 +3,20 @@
 #include <QFontMetrics>
 #include <QMouseEvent>
 #include <QPainter>
+#include <QStylePainter>
 #include "Spire/Spire/Dimensions.hpp"
 #include "Spire/Ui/Ui.hpp"
 
 using namespace Spire;
 
 namespace {
-  auto BOX_SIZE() {
-    static auto size = scale(16, 16);
-    return size;
-  }
-
-  auto PADDING() {
-    static auto padding = scale_width(8);
-    return padding;
-  }
-
   const auto& CHECK_ICON() {
-    static auto icon = imageFromSvg(":/Icons/check.svg", scale(10, 8));
+    static auto icon = imageFromSvg(":/Icons/check.svg", scale(16, 16));
     return icon;
-  }
-
-  const auto& BOX_ENABLED_COLOR() {
-    static auto color = QColor("#FFFFFF");
-    return color;
-  }
-
-  const auto& BOX_DISABLED_COLOR() {
-    static auto color = QColor("#F5F5F5");
-    return color;
-  }
-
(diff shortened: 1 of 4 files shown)


Applications/Spire/Source/Ui/Checkbox.cpp
```diff
@@ -56,99 +27,98 @@ Checkbox::Checkbox(QWidget* parent)
...
+  if(isChecked()) {
+    auto icon = QPixmap::fromImage(CHECK_ICON());
+    if(!isEnabled()) {
+      auto image_painter = QPainter(&icon);
+      image_painter.setCompositionMode(QPainter::CompositionMode_SourceIn);
+      image_painter.fillRect(icon.rect(), CHECK_DISABLED_COLOR());
+    }
+    painter.drawItemPixmap(indicator_rect, Qt::AlignCenter, icon);
+  }
+  auto alignment = Qt::AlignLeft | Qt::AlignVCenter | Qt::TextShowMnemonic;
```
spireken, line 106: It looks like this is a constant so you can declare it in the anonymous namespace.

Applications/Spire/Source/Ui/Checkbox.cpp
```diff
@@ -56,99 +27,98 @@ Checkbox::Checkbox(QWidget* parent)
...
+    #Checkbox:disabled {
+      color: #C8C8C8;
+    }
+    #Checkbox::indicator:disabled {
+      background-color: #F5F5F5;
+    }
+    #Checkbox::indicator:read-only:disabled {
+      background-color: #00000000;
+      border-color: #00000000;
+    })").arg(scale_width(12)).arg(scale_width(1)).
```
spireken, line 64: I think this case is an exception to the operator rule. You can put the period on line 64 before the call to arg(...) on line 65.

spireken (changes requested):
Looks really good.

One issue is that if I change the overall width (say to 150 pixels), clicking on the label doesn't check/uncheck the Checkbox, and the hover affect doesn't show.

Applications/Spire/Source/Ui/Checkbox.cpp
kranar: I need to update the style guide, but the ```.``` should go at the end of the line instead of at the beginning.

Applications/Spire/Source/Ui/Checkbox.cpp
kranar: Similarly for this, constants should be declared as local as possible. Only put it in the anonymous namespace if it's used in multiple functions within the same file.

Applications/Spire/Source/Ui/Checkbox.cpp
spireken: Okay, we'll have to update this as we come across it because it's pretty prevalent.

kranar: Yep, very true.

PR #579, opened 2021-02-03
Fix the Uiviewer scaling issue

shaobinspire:


spireken (approved):

Applications/Spire/Include/Spire/Spire/Dimensions.hpp
```diff
@@ -52,6 +52,35 @@ namespace Spire {
     \return The translated point.
   */
   QPoint translate(const QPoint& p) noexcept;
+
+  //! Scales the pixel width at the system's current DPI to 96 DPI.
```
kranar, line 56: Fix the documentation.

kranar (changes requested):

Applications/Spire/Include/Spire/Spire/Dimensions.hpp
shaobinspire: Do you mean to replace the word scales to the word unscales?

kranar: More than just that, the documentation is incorrect, the function doesn't do what the documentation says it does.

kranar (approved):
The documentation is fine as is.

PR #580, opened 2021-02-04
Border scaling

spireken:


Applications/Spire/Source/Ui/IconButton.cpp
```diff
@@ -64,27 +62,29 @@ void IconButton::keyPressEvent(QKeyEvent* event) {
...
-  }
+  painter.fillRect(rect(), get_background_color());
   auto icon = QPixmap::fromImage(m_icon);
   auto image_painter = QPainter(&icon);
   image_painter.setCompositionMode(QPainter::CompositionMode_SourceIn);
   image_painter.fillRect(icon.rect(), get_current_icon_color());
   painter.drawPixmap((width() - icon.width()) / 2,
     (height() - icon.height()) / 2, icon);
   if(hasFocus()) {
-    painter.setPen({QColor("#4B23A0"), static_cast<double>(scale_width(1))});
```
shaobinspire, line 79: Why is this color not put in IconButton::Style:?

Applications/Spire/Source/Ui/IconButton.cpp
```diff
@@ -64,27 +62,29 @@ void IconButton::keyPressEvent(QKeyEvent* event) {
...
-    painter.drawRect(rect().adjusted(0, 0, -scale_width(1), -scale_height(1)));
+    draw_border(rect(), "#4B23A0", &painter);
```
shaobinspire, line 73: Why is this color not put in IconButton::Style:?

spireken: For now it's essentially a constant. I don't see a reason why we couldn't add it to Style, but until there's a need for I'll leave it out.

shaobinspire: I don’t understand. The background color can be specified through passing IconButton::Style to the constructor when creating an Iconbutton, but why is the border color a constant?

spireken: The border color is currently a constant because there's no instances that need a different border color. It's likely that the font selector's buttons will need options to customize this, but we generally only add things as they're required.

In this case adding a member to the Style struct with a default value is trivial, so I don't think leaving it out until it's needed is a big deal.

shaobinspire (approved):

Applications/Spire/Source/Ui/ColorSelectorButton.cpp
```diff
@@ -61,20 +66,30 @@ bool ColorSelectorButton::eventFilter(QObject* watched, QEvent* event) {
...
+  adjust_style_rect(style_rect);
+  painter.fillRect(style_rect, m_current_color);
 }
 
 QSize	ColorSelectorButton::sizeHint() const {
   return scale(100, 26);
 }
 
+void ColorSelectorButton::adjust_style_rect(QRect& rect) const {
+  rect.adjust(scale_width(1), scale_height(1), -scale_width(1),
```
kranar, line 82: If a method doesn't make use of ```this```, either explicitly or implicitly, then it shouldn't be a method.

Applications/Spire/Source/Ui/Ui.cpp
```diff
@@ -7,6 +7,17 @@
 
 using namespace Spire;
 
+void Spire::draw_border(const QRect& region, const QColor& color,
+    QPainter* painter) {
+  painter->save();
+  auto border_region = QRegion(region).subtracted(region.adjusted(
+    scale_width(1), scale_height(1), -scale_width(1), -scale_height(1)));
+  for(const auto& rect : border_region) {
```
kranar, line 15: Remove the ```const```, ```auto& rect```. Just a matter of coding style preference.

kranar (changes requested):
Looks good.

Applications/Spire/Source/Ui/ColorSelectorButton.cpp
spireken: It shouldn't be a member method, or a method at all?

kranar: What is a member method?

spireken: Oh, I see. I was using 'function' and 'method' interchangeably.

PR #584, opened 2021-02-11
textbox padding

shaobinspire:


Applications/Spire/Source/Ui/TextBox.cpp
```diff
@@ -57,44 +93,15 @@ TextBox::TextBox(QWidget* parent)
 TextBox::TextBox(const QString& text, QWidget* parent)
     : QLineEdit(text, parent),
       m_text(text),
-      m_submitted_text(text) {
+      m_submitted_text(text),
+      m_padding({8, 8}) {
```
spireken, line 97: I would scale these values and remove the scaling from the stylesheet's arguments. Typically whenever we set pixel values they're scaled by the caller, and whenever they're received in a setter method they're assumed to be scaled already.

spireken (changes requested):
Looks good.

spireken (approved):

kranar (approved):

PR #585, opened 2021-02-12
Window fixes

spireken:


spireken:
Darryl reported an issue on a vertical monitor orientation.

PR #589, opened 2021-03-02
Text box

shaobinspire:


Applications/Spire/Source/Ui/TextBox.cpp
```diff
@@ -1,155 +1,95 @@
...
+TextBox::TextBox(const QString& current, QWidget* parent)
+    : QWidget(parent),
+      m_current(current) {
+  m_line_edit = new QLineEdit(m_current, this);
+  m_line_edit->setFrame(false);
+  m_line_edit->setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Expanding);
+  auto layout = new QHBoxLayout(this);
+  layout->setContentsMargins(0, 0, 0, 0);
+  layout->addWidget(m_line_edit);
+  setLayout(layout);
```
spireken, line 59: You don't have to manually set the layout here, because constructing the layout with a parent automatically sets the layout as the parent's main layout: https://doc.qt.io/qt-5/qhboxlayout.html#QHBoxLayout-1

spireken (changes requested):

spireken (approved):

Applications/Spire/Source/Ui/TextBox.cpp
```diff
@@ -162,101 +97,106 @@ connection TextBox::connect_submit_signal(
...
+    m_submit_signal(m_submission);
   }
 }
 
 void TextBox::on_selection_changed() {
-  if(isReadOnly()) {
-    auto text = selectedText();
+  if(is_read_only()) {
+    auto text = m_line_edit->selectedText();
     if(text.endsWith(QChar(0x2026)) || text.endsWith(QStringLiteral("..."))) {
```
kranar, line 163: What happens if the actual current text ends with a "..."? The user can type in something like "Hello...".

kranar (changes requested):
This is solid work overall.

Applications/Spire/Source/Ui/TextBox.cpp
shaobinspire: Great catch. When the text is “Hello…” and the textbox is read-only, it will trigger the selection change event infinitely. It has no problem in non-read-only mode. I am going to fix the bug.

PR #590, opened 2021-03-04
Button

shaobinspire:
The text color in the hover state doesn't not satisfying the specs of the FlatButton.


Applications/Spire/Source/Ui/Button.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+connection Button::connect_clicked_signal(
+    const ClickedSignal::slot_type& slot) const {
+  return m_clicked_signal.connect(slot);
+}
+
+bool Button::eventFilter(QObject* watched, QEvent* event) {
+  if(watched == m_component) {
+    switch(event->type()) {
+    case QEvent::MouseButtonPress:
```
kranar, line 46: Give the component a chance to handle these events and only if the component doesn't handle them, then the Button can handle it.

kranar (changes requested):
This is excellent work, looks like several corner cases were taken care of properly.

kranar (commented):
Overall this is good, let's meet tomorrow on Zoom to go over it a bit more.

shaobinspire (commented):
When the text of the button is overflowed and you click the button, the text is slightly off.

kranar:
Looks like that happens because the border pushes the text, the text is off just by focusing the button without clicking on it. Can you look into why that happens? Does it happen for TextBox?

kranar:
It looks like it happens because the border changes from 0x to 1px. To fix it, use a consistent border size but make it transparent.

kranar:
Shaobin, did you identify any other issue other than the border?

shaobinspire:
> Shaobin, did you identify any other issue other than the border?

Yes, I found a bug of TextBox. The TextBox can't elide text when it focuses out, because you removed the eventFilter. The FocusInEvent and focusOutEvent is textbox's not qlineedit's.

PR #591, opened 2021-03-04
Window shadow

spireken:


Applications/Spire/Source/Ui/Window.cpp
```diff
@@ -163,14 +180,10 @@ bool Window::nativeEvent(const QByteArray& eventType, void* message,
     return true;
   } else if(msg->message == WM_SIZE) {
     if(msg->wParam == SIZE_MAXIMIZED) {
-      auto abs_pos = mapToGlobal(m_title_bar->pos());
-      auto pos = QGuiApplication::screenAt(mapToGlobal({window()->width() / 2,
-        window()->height() / 2}))->geometry().topLeft();
-      pos = QPoint(std::abs(abs_pos.x() - pos.x()),
-        std::abs(abs_pos.y() - pos.y()));
-      setContentsMargins(pos.x(), pos.y(), pos.x(), pos.y());
+      layout()->setContentsMargins({});
```
shaobinspire, line 183: Why do we need to change margins when maximizing or restorting the window?

spireken: We simulate a border by setting the Window's background color and then setting the content layout's margins to make a small part of the background visible. When the Window is maximized we remove these margins to hide the border, then restore them when the Window is restored so that the border is visible again.

If we added a stylesheet for the Window border we would still need margins to make it visible (because of the Window body), so it seems easier to just define the border using the Window background color.

Applications/Spire/Source/Ui/Window.cpp
```diff
@@ -87,14 +87,31 @@ bool Window::nativeEvent(const QByteArray& eventType, void* message,
...
-        GetSystemMetrics(SM_CXFRAME) + GetSystemMetrics(SM_CXPADDEDBORDER);
-      requested_client_area.bottom -=
-        GetSystemMetrics(SM_CYFRAME) + GetSystemMetrics(SM_CXPADDEDBORDER);
+      auto hwnd = reinterpret_cast<HWND>(effectiveWinId());
+      auto placement = WINDOWPLACEMENT{};
+      if (GetWindowPlacement(hwnd, &placement)) {
+        auto& rect = reinterpret_cast<NCCALCSIZE_PARAMS*>(
+          msg->lParam)->rgrc[0];
+        if(IsZoomed(hwnd) || (IsIconic(hwnd) &&
+            placement.flags & WPF_RESTORETOMAXIMIZED)) {
```
shaobinspire, line 96: Why deal with the minimized window?

spireken: In this case I'm checking to see if the Window is minimized, but will be maximized when un-minimized. In this case I still want the Window to have the non-client area sized to the screen's 'working area' (excluding the task bar) size.

shaobinspire (approved):
Looks good!

kranar (approved):
I'm pretty impressed by this, shows you can dig down to a low level and work out nitty gritty issues. Well done.

PR #594, opened 2021-03-24
Implemented the ScrollBar.

shaobinspire:


Applications/Spire/Include/Spire/Ui/ScrollBar.hpp
```diff
@@ -0,0 +1,94 @@
...
+#define SPIRE_SCROLL_BAR_HPP
+#include <QScrollBar>
+#include "Spire/Styles/StyledWidget.hpp"
+#include "Spire/Ui/Ui.hpp"
+
+namespace Spire {
+namespace Styles {
+
+  /** Sets the width. */
+  using Width = BasicProperty<int, struct WidthTag>;
```
spireken, line 11: Is it possible to avoid duplicating this? It seems like these size properties would be common to all StyledWidgets.

Applications/Spire/Include/Spire/UiViewer/StandardUiProfiles.hpp
```diff
@@ -30,6 +31,15 @@ namespace Spire {
 
   //! Returns a UiProfile for the Tooltip.
   UiProfile make_tooltip_profile();
+
+  //! Returns a UiProfile for the ScrollBar.
+  UiProfile make_scroll_bar_profile(Qt::Orientation orientation);
```
spireken, line 36: I think these are supposed to be declared alphabetically.

Applications/Spire/Source/UiViewer/UiViewerWindow.cpp
```diff
@@ -94,6 +94,8 @@ UiViewerWindow::UiViewerWindow(QWidget* parent)
   add(make_icon_button_profile());
   add(make_text_box_profile());
   add(make_tooltip_profile());
+  add(make_vertical_scroll_bar_profile());
```
spireken, line 97: It may be preferable to combine the ScrollBar test widgets into one and have the orientation be a settable property. Whoever is testing it will have to press rebuild to cause this to take effect, but at least it condenses the number of test widgets a little bit.

spireken (changes requested):
Looks very good.

Applications/Spire/Include/Spire/Ui/ScrollBar.hpp
shaobinspire: Yes, they should be in a place where all common properties are declared. What is a better place? I find lots of properties are in Box.hpp, so do I put them to Box.hpp?

kranar:
Remove the size l for now as a style. That will need an upcoming feature to
be supported properly.

spireken (approved):

Applications/Spire/Source/Ui/ScrollBar.cpp
```diff
@@ -0,0 +1,123 @@
...
+void ScrollBar::set_range(const Range& range) {
+  m_scroll_bar->setRange(range.m_start, range.m_end);
+}
+
+void ScrollBar::set_range(int start, int end) {
+  m_scroll_bar->setRange(start, end);
+}
+
+int ScrollBar::get_line_size() const {
+  return QApplication::wheelScrollLines();
```
kranar, line 57: The line size is not the same as the wheel scroll. The line size is the minimum scrollable unit and the wheel scroll is some constant multiple of that unit where the constant is usually 3.

Applications/Spire/Source/Ui/ScrollBar.cpp
```diff
@@ -0,0 +1,123 @@
...
+    QScrollBar:%1 {
+      padding: 0px;)").arg(orientation);
+  auto thumb_style = QString("QScrollBar::handle:%1 {").arg(orientation);
+  for(auto& property : scroll_bar_computed_style.get_properties()) {
+    property.visit(
+      [&] (const BackgroundColor& color) {
+        scroll_bar_style += "background: " +
+          color.get_expression().as<QColor>().name(QColor::HexArgb) + ";";
+      },
+      [] {});
```
kranar, line 100: Feel free to remove the ```[] {}```, it's not needed anymore.

Applications/Spire/Source/Ui/ScrollBar.cpp
```diff
@@ -0,0 +1,123 @@
...
+  }
+  scroll_bar_style += "}";
+  for(auto& property : thumb_computed_style.get_properties()) {
+    property.visit(
+      [&] (const BackgroundColor& color) {
+        thumb_style += "background: " +
+          color.get_expression().as<QColor>().name(QColor::HexArgb) + ";";
+      },
+      [] {});
```
kranar, line 109: Same here.

kranar (changes requested):
Looks good.

Applications/Spire/Source/Ui/ScrollBar.cpp
shaobinspire: I know that, so I asked you this question on the Asana. The line size is actually not used in our scroll bars, and the minimum unit for wheel scroll is needed in the specs. The line size can be used to calculate the unit of the wheel scroll, but it's not a straightforward way to set the unit of the wheel scroll.

kranar: ```get_line_size``` will need to return the line size, not the wheel scroll size.

shaobinspire: I see.

PR #596, opened 2021-03-30
UI viewer signals

spireken:


shaobinspire (approved):
Looks great.

Applications/Spire/Source/UiViewer/StandardUiProfiles.cpp
```diff
@@ -217,6 +217,11 @@ UiProfile Spire::make_decimal_box_profile() {
           }
         }
       });
+      decimal_box->get_model()->connect_current_signal(
```
kranar, line 220: This looks like a duplicate of line 227. Can you look into removing this chunk?

Applications/Spire/Source/UiViewer/StandardUiProfiles.cpp
```diff
@@ -217,12 +217,18 @@ UiProfile Spire::make_decimal_box_profile() {
...
```
kranar, line 220: Did you intend to delete this?

Applications/Spire/Source/UiViewer/StandardUiProfiles.cpp
```diff
@@ -217,12 +217,18 @@ UiProfile Spire::make_decimal_box_profile() {
...
+        [&] (const auto& value) {
+
+        });
       auto current_slot = profile.make_event_slot<QString>(
         QString::fromUtf8("Current"));
       decimal_box->get_model()->connect_current_signal(
-        [=] (const DecimalBox::Decimal& current) {
+        [=, &current] (const DecimalBox::Decimal& value) {
```
kranar, line 227: Can't capture by reference, causes a dangling pointer.

kranar (changes requested):

Applications/Spire/Source/UiViewer/StandardUiProfiles.cpp
spireken: Yes.

Applications/Spire/Source/UiViewer/StandardUiProfiles.cpp
spireken: In other test profiles we capture the properties that are used in a lambda slot as references because they can't be copied. If current is a reference and its captured as a reference, wouldn't it be valid as long as current was valid?

kranar: Yes you are right, I mixed up the name of this variable with another.

kranar (approved):

PR #600, opened 2021-04-13
Login window close button

spireken:


Applications/Spire/Include/Spire/Login/LoginWindow.hpp
```diff
@@ -82,7 +82,8 @@ namespace Spire {
       QLineEdit* m_username_line_edit;
       QLineEdit* m_password_line_edit;
       FlatButton* m_sign_in_button;
-      IconButton* m_exit_button;
+      Button* m_close_button;
+      boost::signals2::scoped_connection m_close_button_connection;
```
shaobinspire, line 86: Why is the connection needed here?

spireken: The scoped_connection is disconnected when it's destructed, but the connection returned by the connect_x_signal functions isn't disconnected when destructed. This can create problems when object that is connected to the signal is destroyed, the slot connected to that signal may no longer exist.

In this case it probably won't make a difference because the TextBox will be destroyed before the LoginWindow so there's no chance for the the slot to be null, but it's an informal standard to store the returned connection as a scoped_connection to prevent issues like this.

shaobinspire: In our project, some returned connections of the connect_x_signal function are stored, but others are not saved. If the issue that you said can be avoided, is it necessary to store it? Is there a guide for this?

kranar: Good question. If the signaller outlives the receiver, then the receiver must keep a ```scoped_connection``` to the signaller to avoid a dangling reference. A common example is a widget that connects to a model; models are shared among many widgets so it's very likely that a model outlives a widget and hence widgets often have ```scoped_connection```s to models. 

In this case, however, the ```LoginWindow``` outlives the ```m_close_button``` so no ```scoped_connection``` is needed.

spireken: I always store the connection, so maybe @kranar can expand on cases where we wouldn't store them.

The [boost documentation](https://www.boost.org/doc/libs/1_63_0/doc/html/signals/tutorial.html#idp607875008) has some info on connections, but not specifically on best practices.

If we passed a function pointer to the connect_x_signal functions, we might not have to store the connections because they would be disconnected when the object connecting to the signal is destroyed. I think our use of lambdas as slots prevents this automatic disconnection because the signal can't determine the state of the objects in the lambda slot.

shaobinspire: I see. Thanks @kranar @spireken

spireken: @kranar Should we say then, that in general for signals from a child widget to a parent widget slot don't bother storing the connection?

kranar: Yeah that works for most cases.

shaobinspire (approved):
Looks great.

Applications/Spire/Source/Login/LoginWindow.cpp
```diff
@@ -4,19 +4,35 @@
 #include <QVBoxLayout>
 #include "Spire/Login/ChromaHashWidget.hpp"
 #include "Spire/Spire/Dimensions.hpp"
+#include "Spire/Ui/Box.hpp"
```
kranar, line 7: What is ```Box.hpp``` used for?

Applications/Spire/Source/Login/LoginWindow.cpp
```diff
@@ -38,19 +54,17 @@ LoginWindow::LoginWindow(const std::string& version, QWidget* parent)
...
-  m_exit_button->installEventFilter(this);
-  connect(m_exit_button, &IconButton::clicked, [=] { window()->close(); });
-  title_bar_layout->addWidget(m_exit_button);
-  title_bar_layout->setStretchFactor(m_exit_button, 32);
+  m_close_button = make_icon_button(imageFromSvg(":/Icons/close.svg",
+    BUTTON_SIZE()), this);
+  set_style(*m_close_button, CLOSE_BUTTON_STYLE());
+  m_close_button->setFixedSize(BUTTON_SIZE());
+  m_close_button->setFocusPolicy(Qt::NoFocus);
+  m_close_button->installEventFilter(this);
```
kranar, line 62: What is the event filter used for?

Applications/Spire/Source/Login/LoginWindow.cpp
```diff
@@ -38,19 +54,17 @@ LoginWindow::LoginWindow(const std::string& version, QWidget* parent)
...
+  m_close_button->connect_clicked_signal([=] {
+    window()->close();
+  });
+  title_bar_layout->addWidget(m_close_button);
+  title_bar_layout->setStretchFactor(m_close_button, 32);
```
kranar, line 67: What is this ```32``` used for?

kranar (changes requested):

Applications/Spire/Source/Login/LoginWindow.cpp
spireken: The Box header defines the basic style properties, but neither the Button nor the Icon header include them.

Applications/Spire/Source/Login/LoginWindow.cpp
spireken: This comes from before I started using fixed sizes and spacing for widgets in layouts. By setting stretches the layout will automatically resize the widgets based on these stretch values, which is contained by the fixed size of the window. The layout won't have any effect on updating this window, but if you want I can update it to use fixed size widgets and spacing to lay out the window contents.

kranar: Okay keep it as is. As long as this scales properly based on resolution/DPI then it's fine.

kranar (approved):

PR #601, opened 2021-04-14
Login username input

spireken:


Applications/Spire/Source/Login/LoginWindow.cpp
```diff
@@ -33,6 +34,18 @@ namespace {
       set(Fill(QColor::fromRgb(0xE6, 0x3F, 0x45)));
     return style;
   }
+
+  auto INPUT_STYLE(StyleSheet style) {
```
shaobinspire, line 38: I think passing by reference is better here.

Applications/Spire/Source/Login/LoginWindow.cpp
```diff
@@ -101,22 +114,17 @@ LoginWindow::LoginWindow(const std::string& version, QWidget* parent)
   content_layout->addWidget(m_status_label);
   content_layout->setStretchFactor(m_status_label, 14);
   content_layout->addStretch(20);
-  m_username_line_edit = new QLineEdit(this);
-  connect(m_username_line_edit, &QLineEdit::textEdited,
-    [=] {on_input_updated();});
-  m_username_line_edit->setPlaceholderText(tr("Username"));
-  m_username_line_edit->setSizePolicy(QSizePolicy::Expanding,
+  m_username_text_box = new TextBox(this);
+  m_username_text_box->get_model()->connect_current_signal(
```
shaobinspire, line 118: Is the scoped connection needed here because of the model?

spireken: Correct.

Applications/Spire/Source/Login/LoginWindow.cpp
spireken: The get_style function returns a const reference, so I can't pass it to INPUT_STYLE as a reference, and StyleSheet's get function returns a reference, so I would have to make a copy of style anyway before I could modify it if it was a const reference.

I'm not sure if const reference is always the standard, but I in this case I do like the implicit copy when passing by value.

shaobinspire: Make sense. I didn't see that the get_style function returns a const reference.

Applications/Spire/Source/Login/LoginWindow.cpp
```diff
@@ -8,6 +8,7 @@
 #include "Spire/Ui/Button.hpp"
 #include "Spire/Ui/DropShadow.hpp"
 #include "Spire/Ui/FlatButton.hpp"
+#include "Spire/Ui/TextBox.hpp"
```
shaobinspire, line 11: include files alphabetically.

shaobinspire (approved):
Looks good now.

Applications/Spire/Source/Login/LoginWindow.cpp
```diff
@@ -101,22 +114,19 @@ LoginWindow::LoginWindow(const std::string& version, QWidget* parent)
   content_layout->addWidget(m_status_label);
   content_layout->setStretchFactor(m_status_label, 14);
   content_layout->addStretch(20);
-  m_username_line_edit = new QLineEdit(this);
-  connect(m_username_line_edit, &QLineEdit::textEdited,
-    [=] {on_input_updated();});
-  m_username_line_edit->setPlaceholderText(tr("Username"));
-  m_username_line_edit->setSizePolicy(QSizePolicy::Expanding,
+  m_username_text_box = new TextBox(this);
+  m_username_connection = m_username_text_box->get_model()->
```
kranar, line 118: Still don't need the connection here, the ```LoginWindow``` outlives the ```m_username_text_box```.

Applications/Spire/Source/Login/LoginWindow.cpp
```diff
@@ -33,6 +34,18 @@ namespace {
       set(Fill(QColor::fromRgb(0xE6, 0x3F, 0x45)));
     return style;
   }
+
+  auto INPUT_STYLE(StyleSheet style) {
+    style.get(Any()).set(border(scale_width(0), QColor::fromRgb(0, 0, 0, 0)));
+    auto font_style = Styles::find<TextStyle>(style.find(Any())->get_block());
+    if(font_style) {
```
kranar, line 41: Can you describe the intention here? Why not unconditionally set the font?

Applications/Spire/Source/Login/LoginWindow.cpp
```diff
@@ -101,22 +114,19 @@ LoginWindow::LoginWindow(const std::string& version, QWidget* parent)
...
-  connect(m_username_line_edit, &QLineEdit::textEdited,
-    [=] {on_input_updated();});
-  m_username_line_edit->setPlaceholderText(tr("Username"));
-  m_username_line_edit->setSizePolicy(QSizePolicy::Expanding,
+  m_username_text_box = new TextBox(this);
+  m_username_connection = m_username_text_box->get_model()->
+    connect_current_signal([=] (const auto& current) {
+      on_input_updated();
+    });
+  m_username_text_box->installEventFilter(this);
```
kranar, line 122: Looks like the event filter can be removed.

kranar (changes requested):
Looking good.

Applications/Spire/Source/Login/LoginWindow.cpp
spireken: I was thinking it's possible that the StyleSheet may not have a TextStyle set. That's not particularly likely because this is the TextBox's stylesheet, but it seemed like good practice to handle that case.

kranar: Oh I see what you mean. Go ahead and fully set the font, including ```"Roboto"``` etc...

kranar: Actually it's fine, leave it as is.

PR #602, opened 2021-04-14
Updated LoginWindow's build and status labels to use a read-only Text…

spireken:
I forgot the status label when listing the changes in the Asana case. I added that change in this PR because it's basically the same change as the build label in terms of styling.

There was an issue with the sizing of the sign in button so I applied a temporary fix to it. When I update the sign in button to use the label button I'll simplify the layout and widget sizing so this won't be an issue.


shaobinspire (approved):
Looks great!

kranar (approved):

PR #603, opened 2021-04-14
fix the infinite loop in DecimalBox 

shaobinspire:


kranar:
Fix incorporated into validate branch.

PR #605, opened 2021-04-15
Login password input

spireken:


Applications/Spire/Source/Login/LoginWindow.cpp
```diff
@@ -136,25 +142,19 @@ LoginWindow::LoginWindow(const std::string& version, QWidget* parent)
...
-  m_password_line_edit->setEchoMode(QLineEdit::Password);
-  m_password_line_edit->setPlaceholderText(tr("Password"));
-  m_password_line_edit->setSizePolicy(QSizePolicy::Expanding,
+  m_password_text_box = new TextBox(this);
+  set_style(*m_password_text_box, PASSWORD_INPUT_STYLE(get_style(
+    *m_password_text_box)));
+  m_password_text_box->get_model()->connect_current_signal(
+    [=] (const auto& current) {
+      on_input_updated();
+      on_password_updated();
```
shaobinspire, line 151: The current is already here, so I think the value can be passed into the on_password_updated function or the function can be replaced with m_chroma_hash_widget->set_text(current).

spireken: I think we can take it a step further and remove both of the on_x_update() functions. The password has no effect on the condition in the on_input_updated() function, and there doesn't appear to be a case where editing the password will update the username. The statement of the on_password_updated() function can be put into the lambda, and the condition in the on_input_update() function can be simplified and put into the lambda slot for the username change signal.

shaobinspire (approved):
Great.

Applications/Spire/Source/Login/LoginWindow.cpp
```diff
@@ -124,7 +130,10 @@ LoginWindow::LoginWindow(const std::string& version, QWidget* parent)
   content_layout->addStretch(20);
   m_username_text_box = new TextBox(this);
   m_username_text_box->get_model()->connect_current_signal(
-    [=] (const auto& current) { on_input_updated(); });
+    [=] (const auto& current) {
+      m_sign_in_button->setDisabled(
+        m_username_text_box->get_model()->get_current().isEmpty());
```
kranar, line 135: ```current.isEmpty()```

kranar (approved):

PR #607, opened 2021-04-22
Window size fix

spireken:
I made some API changes to simplify setting the window body and sizing, or at least make it less ambiguous. This also addresses another issue where the window's minimum size was set to the body size, and not the overall minimum window size.

Edit: demos are in the window_size folder.

### DIFF
--- a/Applications/Spire/Source/Ui/Window.cpp
+++ b/Applications/Spire/Source/Ui/Window.cpp
@@ -40,6 +40,7 @@ namespace {
 
 Window::Window(QWidget* parent)
     : QWidget(parent),
+      m_body(nullptr),
       m_is_resizable(true) {
   setWindowFlags(windowFlags() | Qt::Window | Qt::WindowSystemMenuHint);
   setObjectName("spire_window");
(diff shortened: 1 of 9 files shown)


Applications/Spire/Source/Ui/Window.cpp
```diff
@@ -185,17 +194,56 @@ bool Window::nativeEvent(const QByteArray& eventType, void* message,
...
+  resize_body(size);
+}
+
+void Window::set_fixed_body(QWidget* body, const QSize& size) {
+  if(m_body) {
+    return;
+  }
+  set_window_attributes(false);
+  body->setFixedSize(size);
+  layout()->addWidget(body);
```
shaobinspire, line 241: I don't understand how the Window is resized after setting fixed body because I don't see the resize function is called.

spireken: I think in this case the fixed size body in the layout is influencing the window size. I can't find where it would be documented, but a quick test in Qt Creator with a window that's initially 200x200 and a button that's 500x500 shows that the window will expand to fit the fixed size button, assuming its in the window's layout. In this same scenario, if the window has no layout then its shown at 200x200 pixels and most of the button overlaps (and is partly hidden by) the boundary of the window.

shaobinspire: I see. Thanks.

shaobinspire (approved):
Looks good.

spireken:
To the BookViewUiTester I added some temporary buttons to resize the window and body in fixed increments. This simulates the window and body (respectively) having resize called on them.

I didn't find a need for any flags because a resize is a no-op if the current and set sizes are the same.

PR #608, opened 2021-04-22
Updated ToolbarWindow layout to use fixed size widgets instead of str…

spireken:
The layout in the demo isn't correct yet, but it will be once the Window size fix is merged.


shaobinspire (approved):
Looks good.

kranar:
Resolve the merge conflict, determine if this branch is consistent with the fixes to Window's body, and then re-open the PR.

kranar (approved):
Very good Sir.

PR #612, opened 2021-04-28
ArrayTableModel.

shaobinspire:


Applications/Spire/Include/Spire/Ui/ArrayTableModel.hpp
```diff
@@ -71,10 +72,25 @@ namespace Spire {
...
+      mutable OperationSignal m_operation_signal;
+      std::vector<std::vector<std::any>> m_data;
+      Transaction m_transaction;
+      int m_transaction_level;
   };
 
   template<typename F>
   decltype(auto) ArrayTableModel::transact(F&& transaction) {
+    m_transaction_level++;
+    BOOST_SCOPE_EXIT(this_) {
```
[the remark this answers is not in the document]
shaobinspire: The transact function should return the result of transaction, and after calling transaction, something should be done, for example, whether to emit a signal of transaction when all transactions are over (when m_transaction_level == 0).  The code of the transact function is equivalent to the code snippet below:

`m_transaction_level++;
 auto result = transaction();
 m_transaction_level--;
 m_transaction_level == 0) {
        m_operation_signal(this_->m_transaction);
        m_transaction.m_operations.clear();
}
return result;`

The problem of this code snippet is that when the return type of the transaction is void, the statement "return result;" doesn't work because the void type can't be returned. 

I didn't find a way to eliminate the compiler error except to return transaction() directly. In order to execute subsequent code after returning transaction(), I chose BOOST_SCOPE_EXIT which can do something before the end of the transact function.

spireken (approved):

Applications/Spire/Include/Spire/Ui/ArrayTableModel.hpp
```diff
@@ -71,10 +72,25 @@ namespace Spire {
...
+    private:
+      mutable OperationSignal m_operation_signal;
+      std::vector<std::vector<std::any>> m_data;
+      Transaction m_transaction;
+      int m_transaction_level;
   };
 
   template<typename F>
   decltype(auto) ArrayTableModel::transact(F&& transaction) {
+    m_transaction_level++;
```
kranar, line 85: ```++x``` instead of ```x++```

Applications/Spire/Include/Spire/Ui/ArrayTableModel.hpp
```diff
@@ -71,10 +72,25 @@ namespace Spire {
...
+    BOOST_SCOPE_EXIT(this_) {
+      this_->m_transaction_level--;
+      if(this_->m_transaction_level == 0) {
+        this_->m_operation_signal(this_->m_transaction);
+        this_->m_transaction.m_operations.clear();
+      }
+    } BOOST_SCOPE_EXIT_END
+    return transaction();
```
kranar, line 93: Use perfect fowarding:

```return std::forward<F>(transaction)();```

File: Applications/Spire/Include/Spire/Ui/ArrayTableModel.hpp [hunk 4]
```diff
@@ -71,10 +72,25 @@ namespace Spire {
...
+      int m_transaction_level;
   };
 
   template<typename F>
   decltype(auto) ArrayTableModel::transact(F&& transaction) {
+    m_transaction_level++;
+    BOOST_SCOPE_EXIT(this_) {
+      this_->m_transaction_level--;
+      if(this_->m_transaction_level == 0) {
+        this_->m_operation_signal(this_->m_transaction);
```
kranar, line 89: Nice solution overall, this part isn't reentrant however. You will need to do something like:

    auto transaction = std::move(this_->m_transaction);
    this_->m_transaction.clear();
    this_->m_operation_signal(transaction);

Applications/Spire/Source/Ui/ArrayTableModel.cpp
```diff
@@ -1,38 +1,124 @@
 #include "Spire/Ui/ArrayTableModel.hpp"
+#include <algorithm>
+#include <stdexcept>
 
 using namespace boost;
 using namespace boost::signals2;
 using namespace Spire;
 
-ArrayTableModel::ArrayTableModel() {}
+ArrayTableModel::ArrayTableModel() : m_transaction_level(0) {}
```
kranar, line 9: ArrayTableModel::ArrayTableModel()
  : m_transaction_level(0) {}

Applications/Spire/Source/Ui/ArrayTableModel.cpp
```diff
@@ -1,38 +1,124 @@
...
 
-void ArrayTableModel::push(const std::vector<std::any>& row) {}
+void ArrayTableModel::push(const std::vector<std::any>& row) {
+  if(!m_data.empty() && row.size() != get_column_size()) {
+    throw std::out_of_range("row.size() != get_column_size()");
+  }
+  if(row.empty()) {
```
kranar, line 15: I don't think this check is needed.

Applications/Spire/Include/Spire/Ui/ArrayTableModel.hpp
kranar, line 89: If ```m_transaction.m_operations.size() == 1```, then signal it directly instead of signaling a transaction. Add a unit test for that scenario.

Applications/Spire/Source/Ui/ArrayTableModel.cpp
```diff
@@ -1,38 +1,124 @@
...
+    return;
+  }
+  m_data.push_back(row);
+  auto operation = AddOperation{static_cast<int>(m_data.size()) - 1};
+  if(m_transaction_level > 0) {
+    m_transaction.m_operations.push_back(std::move(operation));
+  } else {
+    m_operation_signal(std::move(operation));
```
kranar, line 23: Remove the ```std::move```, ```m_operation_signal``` takes an lvalue reference.

Applications/Spire/Source/Ui/ArrayTableModel.cpp
```diff
@@ -1,38 +1,124 @@
...
 
-void ArrayTableModel::insert(const TableModel& model, int index) {}
+void ArrayTableModel::insert(const std::vector<std::any>& row, int index) {
+  if(!row.empty() && !m_data.empty() && row.size() != get_column_size()) {
+    throw std::out_of_range("row.size() != get_column_size()");
+  }
+  if(index < 0 || index > get_row_size()) {
+    throw std::out_of_range("The index is out of range.");
+  }
+  if(row.empty()) {
```
kranar, line 34: See above.

Applications/Spire/Source/Ui/ArrayTableModel.cpp
```diff
@@ -1,38 +1,124 @@
...
+    return;
+  }
+  m_data.insert(std::next(m_data.begin(), index), row);
+  auto operation = AddOperation{index};
+  if(m_transaction_level > 0) {
+    m_transaction.m_operations.push_back(std::move(operation));
+  } else {
+    m_operation_signal(std::move(operation));
+  }
```
kranar, line 43: Go ahead and implement ```push``` in terms of ```insert```, ie. ```insert(row, row_count() - 1)```.

Applications/Spire/Source/Ui/ArrayTableModel.cpp
```diff
@@ -1,38 +1,124 @@
...
+  } else {
+    std::move_backward(std::next(m_data.begin(), destination),
+      std::next(m_data.begin(), source), std::next(m_data.begin(), source + 1));
+  }
+  m_data[destination] = std::move(source_row);
+  auto operation = MoveOperation{source, destination};
+  if(m_transaction_level > 0) {
+    m_transaction.m_operations.push_back(std::move(operation));
+  } else {
+    m_operation_signal(std::move(operation));
```
kranar, line 68: Remove ```std::move```

Applications/Spire/Source/Ui/ArrayTableModel.cpp
```diff
@@ -1,38 +1,124 @@
...
+  }
+}
 
 int ArrayTableModel::get_row_size() const {
-  return -1;
+  return m_data.size();
```
kranar, line 86: ```static_cast<int>(...)```

Applications/Spire/Source/Ui/ArrayTableModel.cpp
```diff
@@ -1,38 +1,124 @@
...
+void ArrayTableModel::remove(int index) {
+  if(index < 0 || index >= get_row_size()) {
+    throw std::out_of_range("The index is out of range.");
+  }
+  m_data.erase(std::next(m_data.begin(), index));
+  auto operation = RemoveOperation{index};
+  if(m_transaction_level > 0) {
+    m_transaction.m_operations.push_back(std::move(operation));
+  } else {
+    m_operation_signal(std::move(operation));
```
kranar, line 81: Remove ```std::move```

kranar (changes requested):
This is absolutely phenomenal work, really well done. Very impressed.

Applications/Spire/Include/Spire/Ui/ArrayTableModel.hpp
shaobinspire: I don't understand why we should do that. The basic idea of the solution is all operations during a transaction and its nested transactions are stored in m_transaction for one call of transact, and when all transactions including child transactions are complete ( m_transaction_level == 0), the transaction recording all operations will be signalled and the m_transaction will be cleared for the next call of the function transact.

kranar: Because this implementation isn't reentrant. Re-entrancy is a very tricky property to implement correctly and will introduce bugs. Add a unit test to ensure that the solution is re-entrant and then feel free to provide a solution for it. A re-entrancy test would be something along the lines of this pseudocode:

    table = ArrayTableModel();
    table.connect_signal([&] (const Transaction& transaction) {
      table.push({7, 8, 9});
      REQUIRE(transaction.m_operations == 2);
    }
    table.transact([&] {
      table.add({1, 2, 3});
      table.add({4, 5, 6});
    });

Applications/Spire/Source/Ui/ArrayTableModel.cpp
shaobinspire: The two-dimensional vector can insert an empty vector. We can imagine the scenario: when the model is empty and an empty vector is added into the model, then the size of the model is 1 and the column of the model is 0 at this time. Then when a non-empty vector is going to be inserted, the column of the model will be not same as the column of the vector.  What should we do? Allow the non-empty vector to be inserted and changed the model's column?

kranar: In this case line 13 will throw an exception ```row.size() != get_column_size()```.

shaobinspire: When an empty vector is added the model first, any non-empty vector will not be allowed to be added again.

kranar: That is mostly correct. The empty row can first be removed and then a non-empty row added. In general all rows must have the same number of columns.

shaobinspire: So should I add the process of removing a empty row if it exists or keep the check here to prevent an empty row from being added?

kranar: No, all that's need is to remove lines 15-17.

shaobinspire: Below is the unit test of insert, can you point out if it meets expectations? 

auto model = ArrayTableModel();
    auto signal_count = 0;
    auto index = 0;
    auto connection = model.connect_operation_signal(
      [&] (const TableModel::Operation& operation) {
        ++signal_count;
        auto add_operation = get<TableModel::AddOperation>(&operation);
        REQUIRE(add_operation != nullptr);
        REQUIRE(add_operation->m_index == index);
      });

    index = 0;
    REQUIRE_NOTHROW(model.insert({}, index));            // here an empty row is insert
    REQUIRE(model.get_row_size() == 1);                        // is the size correct?
    REQUIRE(signal_count == 1);

    index = 0;
    REQUIRE_NOTHROW(model.insert({1, 2, 3}, index));
    REQUIRE(model.get_row_size() == 1);            **// if the empty row above is removed, the row_size is still 1. It's not intuitive that the insert function deletes one row.**
    REQUIRE(signal_count == 2);
    REQUIRE(model.get<int>(0, 0) == 1);
    REQUIRE(model.get<int>(0, 1) == 2);

kranar: ```REQUIRE_NOTHROW(model.insert({1, 2, 3}, index));```

The above line should throw an exception ```std::out_of_range```

kranar (approved):

PR #613, opened 2021-04-30
Visitor of TableModel

shaobinspire:


spireken (approved):

Applications/Spire/Include/Spire/Ui/TableModel.hpp
```diff
@@ -113,12 +114,33 @@ namespace Details {
       virtual boost::signals2::connection connect_operation_signal(
         const typename OperationSignal::slot_type& slot) const = 0;
 
+      /**
+       * Applies a callable to the underlying operation.
+       * @param f The callable to apply.
+       */
+      template<typename F>
+      void visit(const Operation& operation, F&& f) const;
```
kranar, line 122: Prefer free functions over member functions:

https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c4-make-a-function-a-member-only-if-it-needs-direct-access-to-the-representation-of-a-class

Applications/Spire/Include/Spire/Ui/TableModel.hpp
```diff
@@ -127,6 +149,53 @@ namespace Details {
   const T& TableModel::get(int row, int column) const {
     return std::any_cast<const T&>(at(row, column));
   }
+
+  template<typename F>
+  void TableModel::visit(const TableModel::Operation& operation, F&& f) const {
+    if(operation.type() == typeid(Transaction)) {
```
kranar, line 155: Use the declaration form of the if statement to avoid performing two tests:

```if(auto transaction = boost::get<Transaction>(&operation)) {```

https://en.cppreference.com/w/cpp/language/if

File: Applications/Spire/Include/Spire/Ui/TableModel.hpp [hunk 3]
```diff
@@ -127,6 +149,53 @@ namespace Details {
...
+      auto transaction = boost::get<Transaction>(&operation);
+      for(const auto& transaction_operation : transaction->m_operations) {
```
kranar, line 157: Remove ```const```

Applications/Spire/Include/Spire/Ui/TableModel.hpp
```diff
@@ -127,6 +149,53 @@ namespace Details {
...
+        visit(transaction_operation, std::forward<F>(f));
+      }
+    } else {
+      if constexpr(std::is_invocable_v<std::decay_t<F>>) {
+        std::forward<F>(f)();
```
kranar, line 162: Remove this clause.

Applications/Spire/Include/Spire/Ui/TableModel.hpp
```diff
@@ -127,6 +149,53 @@ namespace Details {
...
+      } else if constexpr(std::is_invocable_r_v<void, F, const Operation&>) {
+        std::forward<F>(f)(operation);
```
kranar, line 164: Remove this clause as well.

kranar (changes requested):
Looks great.

Applications/Spire/Include/Spire/Ui/TableModel.hpp
shaobinspire: This part is used to deal with the default hanlder whose parameter is auto.

Applications/Spire/Include/Spire/Ui/TableModel.hpp
shaobinspire: If we will define several visit functions for the different models in the future, will there be a risk of naming confict?

Applications/Spire/Source/SpireTester/TableModelTester.cpp
```diff
@@ -0,0 +1,250 @@
...
+#include <doctest/doctest.h>
+#include "Spire/Ui/ArrayTableModel.hpp"
+
+using namespace boost;
+using namespace boost::signals2;
+using namespace Spire;
+
+TEST_SUITE("TableModel") {
+  TEST_CASE("visit_all_operations") {
+    auto model = ArrayTableModel();
```
kranar, line 11: For the unit tests, make use of the minimum number of dependencies needed to ensure correctness, this helps ensure that if something goes wrong, what is wrong is directly related to the specific functionality being tested.

For the visitor, that means removing the use of the ```ArrayTableModel``` and working directly with ```TableModel::Operation```.

Applications/Spire/Include/Spire/Ui/TableModel.hpp
kranar: No, C++ uses ADL to resolve naming conflicts.

shaobinspire: Oh, I got it. When the visit function template is instantiated, the function will has different parameters for the different models.

Applications/Spire/Include/Spire/Ui/TableModel.hpp
shaobinspire: I don't understand why const should be removed. According to the express intent in the core guidelines, here const expresses the intent that the operation in the loop can't be modified.

https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#p3-express-intent

kranar: For our codebase we don't use ```const``` other than when it's needed to satisfy an API. So for example we don't declare variables to be ```const```.

Applications/Spire/Include/Spire/Ui/TableModel.hpp
```diff
@@ -150,17 +129,32 @@ namespace Details {
...
+  template<typename T>
+  struct TypeExtractor {};
+
+  template<typename T>
+  struct TypeExtractor<Beam::TypeSequence<T>> {
+    using type = std::decay_t<T>;
+  };
+
+  template<typename T, typename U>
+  struct TypeExtractor<Beam::TypeSequence<T, U>> {
```
kranar, line 141: What is this used for?

Applications/Spire/Include/Spire/Ui/TableModel.hpp
```diff
@@ -150,17 +129,32 @@ namespace Details {
...
+  void visit(const TableModel::Operation& operation, F&& f){
+    if(auto transaction = boost::get<TableModel::Transaction>(&operation)) {
+      for(auto& transaction_operation : transaction->m_operations) {
         visit(transaction_operation, std::forward<F>(f));
       }
     } else {
-      if constexpr(std::is_invocable_v<std::decay_t<F>>) {
-        std::forward<F>(f)();
-      } else if constexpr(std::is_invocable_r_v<void, F, const Operation&>) {
+      if constexpr(std::is_invocable_r_v<
```
kranar, line 156: Prefer ```is_invocable_v``` over ```is_invocable_r_v```.

Applications/Spire/Source/SpireTester/TableModelTester.cpp
```diff
@@ -1,250 +1,159 @@
...
+        [&] (const TableModel::RemoveOperation& remove_operation) {
+          REQUIRE(remove_operation.m_index == 1);
+        },
+        [&] (const TableModel::UpdateOperation& update_operation) {
+          REQUIRE(update_operation.m_row == 0);
+          REQUIRE(update_operation.m_column == 0);
+        });
+    };
+    auto add_operation = TableModel::AddOperation{0};
+    visitor(add_operation);
```
kranar, line 28: These tests don't guarantee correctness. For example this unit test would pass if ```visit``` had the following implementation:

    template<typename... F>
    void visit(const TableModel::Operation&, F&&...) {}

One strategy to ensure unit tests work properly is to write a few unit tests before working on the implementation and verifying that they all fail.

kranar (changes requested):
Looks better, some minor improvements could be made.

Applications/Spire/Source/SpireTester/TableModelTester.cpp
shaobinspire: I think the implementation that you mentioned should be included in the following implementation.
  template<typename F, typename... G>
  void visit(const TableModel::Operation& operation, F&& f, G&&... g) {}

In the unit tests, visiting all operations, visiting partial operations and visiting the default handler have been included, so the two implementations below have been tested:
 template<typename F>
  void visit(const TableModel::Operation& operation, F&& f){}
and
template<typename F, typename... G>
  void visit(const TableModel::Operation& operation, F&& f, G&&... g) {}

kranar: My point is that your unit tests are not testing anything. If you submitted a pull request with the following implementation:

    template<typename... F>
    void visit(const TableModel::Operation&, F&&...) {}

Almost all of your unit tests would pass. In order for a unit test to be valid, they must all fail given the above implementation.

PR #615, opened 2021-05-05
Translated table model

shaobinspire:


spireken (approved):
Looks good.

Applications/Spire/Include/Spire/Ui/TranslatedTableModel.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+  /** Implements a TableModel performing operations without modifying the original model. */
+  class TranslatedTableModel : public TableModel {
+    public:
+
+      /** 
+       * Constructs a TranslatedTableModel from a TableModel.
+       * @param source The original model.
+      */
+      TranslatedTableModel(std::shared_ptr<TableModel> source);
```
kranar, line 16: ```explicit```

Applications/Spire/Include/Spire/Ui/TranslatedTableModel.hpp
```diff
@@ -0,0 +1,81 @@
...
+      boost::signals2::scoped_connection m_source_operation_connection;
+      Transaction m_transaction;
+      int m_transaction_level;
+
+      void push(Operation&& operation);
+  };
+
+  template<typename F>
+  decltype(auto) TranslatedTableModel::transact(F&& transaction) {
+    ++m_transaction_level;
```
kranar, line 62: This is an exact copy/paste of ```ArrayTableModel::transact```, avoid copying and pasting code, identify a suitable abstraction and encapsulate the functionality so it can be reused without copy/paste.

Applications/Spire/Include/Spire/Ui/TranslatedTableModel.hpp
```diff
@@ -0,0 +1,81 @@
...
+      void move(int source, int destination);
+
+      int get_row_size() const override;
+
+      int get_column_size() const override;
+
+      const std::any& at(int row, int column) const override;
+
+      boost::signals2::connection connect_operation_signal(
+        const typename OperationSignal::slot_type& slot) const override;
```
kranar, line 42: Remove ```typename```.

Applications/Spire/Source/Ui/TranslatedTableModel.cpp
```diff
@@ -0,0 +1,132 @@
...
+  m_f();
+}
+
+TranslatedTableModel::TranslatedTableModel(std::shared_ptr<TableModel> source)
+    : m_source(source),
+      m_transaction_level(0) {
+  m_index_mapping.resize(source->get_row_size());
+  std::iota(m_index_mapping.begin(), m_index_mapping.end(), 0);
+  m_source_operation_connection = m_source->connect_operation_signal(
+    [=] (const Operation& operation) {
```
kranar, line 20: Only code involving construction should go into the constructor, things like handling signals or general "business logic" should be moved out into its own function.

Applications/Spire/Source/Ui/TranslatedTableModel.cpp
```diff
@@ -0,0 +1,132 @@
...
+      visit(operation,
+        [=] (const AddOperation& add_operation) {
```
kranar, line 22: Local lambdas should capture by reference.

Applications/Spire/Source/Ui/TranslatedTableModel.cpp
```diff
@@ -0,0 +1,132 @@
...
+          std::transform(m_index_mapping.begin(), m_index_mapping.end(),
```
kranar, line 23: Optimize the common case, for example when a row is added to the end of the table this can be implemented in ```O(1)```.

Applications/Spire/Source/Ui/TranslatedTableModel.cpp
```diff
@@ -0,0 +1,132 @@
...
+              if(index >= add_operation.m_index) {
+                return index + 1;
+              }
+              return index;
+            });
+          m_index_mapping.push_back(add_operation.m_index);
+          push(AddOperation{get_row_size() - 1});
+        },
+        [=] (const MoveOperation& move_operation) {
+          if(move_operation.m_source < move_operation.m_destination) {
```
kranar, line 35: This looks good!

Applications/Spire/Source/Ui/TranslatedTableModel.cpp
```diff
@@ -0,0 +1,132 @@
...
+            [&] (auto index) {
+              if(index > remove_operation.m_index) {
+                return index - 1;
+              }
+              return index;
+            });
+          push(RemoveOperation{removed_index});
+        },
+        [=] (const UpdateOperation& update_operation) {
+          push(UpdateOperation{std::distance(m_index_mapping.begin(),
```
kranar, line 76: This is too slow unfortunately, as a general matter the API for this table should not be slower than the API for the source. In other words if the source can implement this in ```O(1)``` then the translation also needs to implement this in ```O(1)``` as well. This implementation, however, is ```O(N)```.

Moves, adds, deletions to the source... those are all ```O(N)``` so it's perfectly fine if this translation is also implements them in ```O(N)```, but updates are universally an ```O(1)``` operation.

Applications/Spire/Source/Ui/TranslatedTableModel.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+void TranslatedTableModel::move(int source, int destination) {
+  if(source < 0 || source >= get_row_size() || destination < 0 ||
+      destination >= get_row_size()) {
+    throw std::out_of_range("The source or destination is out of range.");
+  }
+  if(source == destination) {
+    return;
+  }
+  auto source_row = std::move(m_index_mapping[source]);
```
kranar, line 91: Remove ```std::move``` and don't bother moving ```int```s.

Applications/Spire/Source/Ui/TranslatedTableModel.cpp
kranar, line 20: This does not handle transactions properly. If a transaction is performed on source then its translation must also be a transaction. Add a unit test to handle transactions and then implement a fix for it.

kranar (changes requested):
Excellent first attempt. A lot of changes to make but it's definitely on the right track, keep it going!

Applications/Spire/Source/Ui/TranslatedTableModel.cpp
shaobinspire: There are some unit tests about transaction of source. When a transaction is performed on source, the transaction will be passed into visit function inside the slot and it will be broken down to basic operations.

kranar: A transaction in the source must be treated as a transaction in the translation.

shaobinspire: How to handle it? Does the translation deal with all operations in a transaction of the source but not signal them?

kranar: translaction.connect([&] (const auto& operation) {
      auto transaction = get<Transaction>(&operation);
      REQUIRE(transaction != nullptr);
      REQUIRE(get<AddOperation>(&transaction->m_operations[0]) != nullptr);
      REQUIRE(get<AddOperation>(&transaction->m_operations[1]) != nullptr);
    });
    source.transact([&] {
      source.push({5});
      source.push({6});
    });

shaobinspire: oh, I see.

kranar (approved):

PR #617, opened 2021-05-13
SortedTableModel

shaobinspire:


Applications/Spire/Source/Ui/SortedTableModel.cpp
```diff
@@ -0,0 +1,216 @@
...
+            } else if(i->m_index == operation.m_column) {
+              return true;
+            }
+          }
+          return false;
+        }();
+        if(!is_sorted_column) {
+          return;
+        }
+        sort();
```
kranar, line 197: Resorting on every update is way to inefficient, the complexity of a sort operation is ```Ω(n * log(n))``` which means every time a value is modified a full blown quasi-linear operation is performed. Consider a financial table that supports sorting like the one on this website:

https://www.tradingview.com/screener/

Now imagine if anytime any one of those numbers was updated, a full blown ```Ω(n * log(n))``` had to be performed. That table would effectively be unusable.

Applications/Spire/Source/SpireTester/SortedTableModelTester.cpp
```diff
@@ -0,0 +1,1010 @@
...
+    auto order = sorted_model.get_column_order();
+    REQUIRE(sorted_model.get<int>(0, 0) == 1);
+    REQUIRE(sorted_model.get<int>(1, 0) == 2);
+    REQUIRE(sorted_model.get<int>(2, 0) == 4);
+    REQUIRE(sorted_model.get<int>(3, 0) == 9);
+    REQUIRE(source->get<int>(0, 0) == 4);
+    REQUIRE(source->get<int>(1, 0) == 2);
+    REQUIRE(source->get<int>(2, 0) == 9);
+    REQUIRE(source->get<int>(3, 0) == 1);
+    *order.begin() = order.front().cycle();
```
kranar, line 30: ```*order.begin()``` -> ```order.front()```

kranar (changes requested):
The unit tests are very well done, very thorough and well thought out. The implementation has a great deal of room for improvement both in terms of performance and abstraction.

Applications/Spire/Source/Ui/SortedTableModel.cpp
```diff
@@ -0,0 +1,216 @@
...
+int SortedTableModel::get_row_size() const {
+  return m_source->get_row_size();
+}
+
+int SortedTableModel::get_column_size() const {
+  return m_source->get_column_size();
+}
+
+const std::any& SortedTableModel::at(int row, int column) const {
+  return m_source->at(m_mapping[row], column);
```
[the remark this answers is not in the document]
shaobinspire: The size of m_mapping is the same as the size of m_source, and m_source->at has done the same checking.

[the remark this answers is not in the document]
kranar: The following will result in undefined behavior, you can run it in Scratch or add a unit test for it, if you run it in Debug mode on MSVC you should get an assertion:

    auto source = std::make_shared<ArrayTableModel>();
    auto sorted = SortedTableModel(source);
    sorted.at(234892, 0);

Applications/Spire/Source/Ui/SortedTableModel.cpp
```diff
@@ -0,0 +1,216 @@
...
+  } 
+  return ColumnOrder{m_index, Ordering::NONE};
+}
+
+SortedTableModel::SortedTableModel(std::shared_ptr<TableModel> source)
+  : SortedTableModel(source, m_order) {}
+
+SortedTableModel::SortedTableModel(std::shared_ptr<TableModel> source,
+  std::vector<ColumnOrder> order)
+    : SortedTableModel(source, order, get_default_comparator()) {}
```
spireken, line 23: Line up the colon with the parameter on line 22 (single indent).

Applications/Spire/Source/Ui/SortedTableModel.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+SortedTableModel::SortedTableModel(std::shared_ptr<TableModel> source,
+  Comparator comparator)
+    : SortedTableModel(source, m_order, comparator) {}
+
+SortedTableModel::SortedTableModel(std::shared_ptr<TableModel> source,
+  std::vector<ColumnOrder> order, Comparator comparator)
```
spireken, line 30: When a constructor has a body, double indent the parameter and colon, and triple indent additional members in the initializer list.

shaobinspire:
I improved the time complexity of the update operation to O(log(n)) + O(n). The basic idea is to search the insert range in the upper part and the lower part of the table divided by the update row using std::lower_bound and std::upper_bound and the insert index can be got through the range. The search process costs O(log(n)). The next step is to swap between the inserted index and the updated index, which is O(n). This approach is better than std::sort ( n * O(log(n) ).
I don’t have time to abstract the duplicate code today, and I plan to do it tomorrow.

Applications/Spire/Source/Ui/SortedTableModel.cpp
```diff
@@ -0,0 +1,283 @@
...
+  if(m_order == Ordering::NONE) {
+    return ColumnOrder{m_index, Ordering::ASCENDING};
+  } else if(m_order == Ordering::ASCENDING) {
+    return ColumnOrder{m_index, Ordering::DESCENDING};
+  } 
+  return ColumnOrder{m_index, Ordering::NONE};
+}
+
+SortedTableModel::SortedTableModel(std::shared_ptr<TableModel> source)
+  : SortedTableModel(source, m_order) {}
```
kranar, line 40: The use of ```m_order``` here is undefined behavior.

Forward arguments using ```std::move```.

Applications/Spire/Source/Ui/SortedTableModel.cpp
```diff
@@ -0,0 +1,216 @@
...
+    if(m_reverse_mapping[i] >= row_index) {
+      m_reverse_mapping[i] += direction;
+    }
+  }
+}
+
+void SortedTableModel::on_operation(const Operation& operation) {
+  m_transaction.transact([&] {
+    visit(operation,
+      [&] (const AddOperation& operation) {
```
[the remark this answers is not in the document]
kranar: This has not been resolved.

Applications/Spire/Source/Ui/SortedTableModel.cpp
```diff
@@ -0,0 +1,283 @@
...
+      [&] (const RemoveOperation& operation) {
+        auto row_index = m_reverse_mapping[operation.m_index];
+        translate(-1, operation.m_index);
+        m_mapping.erase(m_mapping.begin() + row_index);
+        m_reverse_mapping.erase(m_reverse_mapping.begin() + operation.m_index);
+        m_transaction.push(RemoveOperation{row_index});
+      },
+      [&] (const UpdateOperation& operation) {
+        auto row_index = m_reverse_mapping[operation.m_row];
+        m_transaction.push(UpdateOperation{row_index, operation.m_column});
```
kranar, line 222: This violates the requirement that this table is always in sorted order. The update must be performed at the sorted position, not before.

kranar (commented):
I don't think this will manage to get done tonight, so feel free to work on one of the other tasks assigned to you.

Applications/Spire/Source/Ui/SortedTableModel.cpp
shaobinspire: The row_index is the index of the sorted table, not the index of the source.

PR #619, opened 2021-05-18
Remove the nativeEvent filter from DecimalBox

shaobinspire:
The updated UiViewer was put under demos/decimal_box_removing_native_event.


spireken (approved):
Much simpler, looks good.

kranar (approved):

PR #628, opened 2021-05-31
Overlay Panel

shaobinspire:


Applications/Spire/Source/Ui/OverlayPanel.cpp
```diff
@@ -0,0 +1,219 @@
...
+    : QWidget(parent, Qt::Tool | Qt::FramelessWindowHint |
+        Qt::NoDropShadowWindowHint),
+      m_body(body),
+      m_top_level_window(nullptr),
+      m_is_closed_on_blur(true),
+      m_is_closed(false),
+      m_positioning(Positioning::PARENT) {
+  setAttribute(Qt::WA_TranslucentBackground);
+  setAttribute(Qt::WA_QuitOnClose);
+  auto box = new Box(m_body);
```
spireken, line 70: I think Box needs a parent here.

spireken (changes requested):
Looks very good.

Applications/Spire/Source/Ui/OverlayPanel.cpp
shaobinspire: The box will be added into a layout, and the layout will automatically reparent the box.

spireken (approved):

Applications/Spire/Source/Ui/OverlayPanel.cpp
```diff
@@ -0,0 +1,219 @@
...
+OverlayPanel::OverlayPanel(QWidget* body, QWidget* parent)
+    : QWidget(parent, Qt::Tool | Qt::FramelessWindowHint |
+        Qt::NoDropShadowWindowHint),
+      m_body(body),
+      m_top_level_window(nullptr),
+      m_is_closed_on_blur(true),
+      m_is_closed(false),
+      m_positioning(Positioning::PARENT) {
+  setAttribute(Qt::WA_TranslucentBackground);
+  setAttribute(Qt::WA_QuitOnClose);
```
kranar, line 69: Why is this attribute set?

Applications/Spire/Source/Ui/OverlayPanel.cpp
```diff
@@ -0,0 +1,219 @@
...
+  layout->addWidget(box);
+  proxy_style(*this, *box);
+  set_style(*box, DEFAULT_STYLE());
+  auto shadow = new QGraphicsDropShadowEffect();
+  shadow->setColor(DROP_SHADOW_COLOR);
+  shadow->setOffset(translate(DROP_SHADOW_OFFSET));
+  shadow->setBlurRadius(scale_width(DROP_SHADOW_RADIUS));
+  box->setGraphicsEffect(shadow);
+  m_body->installEventFilter(this);
+  if(m_top_level_window = find_top_level_window()) {
```
kranar, line 82: This doesn't look right.

Applications/Spire/Source/Ui/OverlayPanel.cpp
```diff
@@ -0,0 +1,219 @@
...
+}
+
+void OverlayPanel::position() {
+  if(m_positioning == Positioning::PARENT) {
+    auto parent_geometry = parentWidget()->rect();
+    auto parent_bottom_left = parentWidget()->mapToGlobal(
+      parent_geometry.bottomLeft());
+    auto screen_geometry =
+      get_current_screen(parent_bottom_left)->availableGeometry();
+    auto get_x = [&] {
```
kranar, line 197: Why is this a function?

kranar (changes requested):

Applications/Spire/Source/Ui/OverlayPanel.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+  auto DEFAULT_STYLE() {
+    auto style = StyleSheet();
+    style.get(Any()).
+      set(BackgroundColor(QColor::fromRgb(0xFF, 0xFF, 0xFF))).
+      set(border(scale_width(1), QColor::fromRgb(0xA0, 0xA0, 0xA0)));
+    return style;
+  }
+
+  QWidget* find_top_level_window() {
```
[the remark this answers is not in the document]
shaobinspire: The function is used to find the top level window. The OverlayPanel moves along with the top level window by tracking the move event. Child widgets don't send the move event, and only top level window can send it, so the OverlayPanel can't receive the move event from its parent but it can get the event from the top level window.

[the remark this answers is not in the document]
kranar: Okay, go ahead and remove this function, in general there is no such thing as "the top level window". There are many top level windows and just picking one arbitrarily is incorrect.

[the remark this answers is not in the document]
shaobinspire: If the OverlayPanel doesn't get a top-level window, it can't stick on its parent. 

Logically, the top-level window should be the top-level window where the OverlayPanel's parent is located. The top level window can be got by parent->window(). However, if a widget doesn't have a parent, the window() function will return the widget. For example, in the UiViewer, the OverlayPanel's parent is a button, but the button don't specify a parent because the button's parent can't be got from the factory function of the UiProfile. If find_top_level_window is removed, the OverlayPanel can't get move events to move along with its parent.

In general every widget we use should have a parent. If in order to have the "stick" effect and remove this function, we need to add a parent to each widget in the UiViewer.

Applications/Spire/Source/Ui/OverlayPanel.cpp
shaobinspire: It makes Qt quit the application when the last widget with the attribute set has accepted closeEvent(). If the attribute is not set and the application is closed, the GUI will disappear but the application will be alive to be a zombie process.

Applications/Spire/Source/Ui/OverlayPanel.cpp
shaobinspire: Yep, it should be removed.

Applications/Spire/Source/Ui/OverlayPanel.cpp
kranar: Sounds good.

Applications/Spire/Source/Ui/OverlayPanel.cpp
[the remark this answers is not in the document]
kranar: Okay, look at fixing this and then resubmit the PR. Everything else looks good.

kranar (changes requested):
Looks good, there is a warning as follows:

QLayout: Attempting to add QLayout "" to QWidget "", which already has a layout


Look into what's causing that warning and then update the PR, that's all that's left for this.

shaobinspire:
> Attempting to add QLayout "" to QWidget "", which already has a layout

I rebuilt the project, but I didn't find any build warning.

kranar (approved):

PR #629, opened 2021-06-01
Updated TextBox to support automatic resizing to its contents when re…

spireken:
Because the UiViewer sets fixed sizes when changed using the width and height properties, modifying the height and width results in this auto resize effect being disabled.

### DIFF
--- a/Applications/Spire/Source/Ui/TextBox.cpp
+++ b/Applications/Spire/Source/Ui/TextBox.cpp
@@ -111,7 +111,6 @@ TextBox::TextBox(std::shared_ptr<TextModel> model, QWidget* parent)
   m_layers = new LayeredWidget(this);
   m_line_edit = new QLineEdit(m_model->get_current());
   m_line_edit->setFrame(false);
-  m_line_edit->setTextMargins(-2, 0, 0, 0);
   m_line_edit->setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Expanding);
   m_text_validator = new TextValidator(m_model, this);
   m_line_edit->setValidator(m_text_validator);
@@ -185,7 +184,7 @@ connection
 
 QSize TextBox::sizeHint() const {
   auto size_hint = QSize(m_line_edit->fontMetrics().horizontalAdvance(
-      m_model->get_current()) + 3, m_line_edit->font().pixelSize());
+      m_model->get_current()) + 5, m_line_edit->font().pixelSize());
   return size_hint.grownBy(m_padding).grownBy(m_border_sizes);
 }
 
--- a/Applications/Spire/Source/Ui/TextBox.cpp
+++ b/Applications/Spire/Source/Ui/TextBox.cpp
@@ -27,7 +27,8 @@ namespace {
       set(border(scale_width(1), QColor::fromRgb(0xC8, 0xC8, 0xC8))).
       set(text_style(font, QColor::fromRgb(0, 0, 0))).
       set(TextAlign(Qt::Alignment(Qt::AlignLeft) | Qt::AlignVCenter)).
-      set(horizontal_padding(scale_width(8)));
+      set(horizontal_padding(scale_width(8))).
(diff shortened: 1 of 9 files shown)


kranar (commented):
This is okay as a proof of concept but still needs a lot of work. There is nothing special about being read-only, remove anything to do with read-only. Also this implementation conflicts with how Qt manages layouts and is implemented without any regard to the layout system. We can't have one of the most fundamental building blocks in the UI kit disregard the layout system.

There's also a lot of quirkiness which doesn't compose well, like the line edit having a fixed size.

Overall this solution is kind of a brute force approach that technically passes the UI kit, but as soon as this gets integrated with other components will fall apart.

spireken:
As a more natural test of the TextBox I added a second demo that shows the TextBox in a layout with a Maximum size hint. The demo's relevant code in scratch is: https://gist.github.com/spireken/2c373f78f413d0b94effed0ef62d0774

Regarding the calls to adjustSize in the TextBox's test profile, since the test TextBox is in the UiViewerWindow's scroll area the TextBox's calls to updateGeometry are effectively ignored because the scroll area's handling of the TextBox's LayoutRequest doesn't result in a resize, but calling adjustSize manually does cause the TextBox to resize itself to its size hint.

Applications/Spire/Source/UiViewer/StandardUiProfiles.cpp
```diff
@@ -685,20 +685,24 @@ UiProfile Spire::make_text_box_profile() {
...
       });
       auto& placeholder = get<QString>("placeholder", profile.get_properties());
       placeholder.connect_changed_signal([text_box] (const auto& text) {
         text_box->set_placeholder(text);
       });
-      text_box->get_model()->connect_current_signal([&] (const auto& value) {
-        current.set(value);
-      });
+      text_box->get_model()->connect_current_signal(
+        [&, text_box = text_box] (const auto& value) {
```
shaobinspire, line 702: Why is text_box = text_box used here?

Applications/Spire/Source/Ui/TextBox.cpp
```diff
@@ -184,7 +183,9 @@ connection
 }
 
 QSize TextBox::sizeHint() const {
-  return scale(160, 30);
+  auto size_hint = QSize(m_line_edit->fontMetrics().horizontalAdvance(
+      m_model->get_current()) + 5, m_line_edit->fontMetrics().height());
```
shaobinspire, line 187: Why add 5 to the width?

Applications/Spire/Source/UiViewer/StandardUiProfiles.cpp
spireken: 'current' is captured by reference as the default capture , but if I capture text_box by reference it's referencing a pointer that is going out of scope. So in this case I copy the text_box pointer into another pointer that's also called text_box that is used within the lambda.

Applications/Spire/Source/Ui/TextBox.cpp
spireken: Qt adds a fixed horizontalMargin of 2 to both sides of the text, and adds another pixel to the right side of the text for the cursor. These 5 extra pixels are always there regardless of the padding of the QLineEdit, so I add it back so that when the TextBox's size hint is calculated it doesn't result in the text 'scrolling' behind these extra margins. Qt doesn't scale these values so I simply add '+ 5' to the size hint's width.

It would be preferable to be able to remove these margins but aside from implementing a custom QStyle or painting the line edit, it doesn't appear like there's a simple solution so it was decided to maintain these margins.

Applications/Spire/Source/UiViewer/StandardUiProfiles.cpp
shaobinspire: I think [&, text_box] or [=, &current] is better.

shaobinspire (approved):

kranar (approved):

PR #633, opened 2021-06-09
OverlayPanel can be closed on escape key press.

shaobinspire:
The TextBox can deal with ESC event, so when the TextBox gets focus, the panel can't be closed.

### DIFF
--- a/Applications/Spire/Source/Ui/OverlayPanel.cpp
+++ b/Applications/Spire/Source/Ui/OverlayPanel.cpp
@@ -142,9 +142,9 @@ bool OverlayPanel::event(QEvent* event) {
 void OverlayPanel::keyPressEvent(QKeyEvent* event) {
   if(event->key() == Qt::Key_Escape && !m_is_closed) {
     fade(true);
-  } else {
-    QWidget::keyPressEvent(event);
+    return;
   }
+  QWidget::keyPressEvent(event);
 }
 
 void OverlayPanel::fade(bool reverse) {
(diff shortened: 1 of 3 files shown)


Applications/Spire/Source/Ui/OverlayPanel.cpp
```diff
@@ -139,6 +139,16 @@ bool OverlayPanel::event(QEvent* event) {
   return QWidget::event(event);
 }
 
+void OverlayPanel::keyPressEvent(QKeyEvent* event) {
+  if(event->key() == Qt::Key_Escape) {
+    if(!m_is_closed) {
+      fade(true);
+    }
```
spireken, line 146: In cases like this I think we usually do an 'early return', where after the second if statement the method would return, then the call to `QWidget::keyPressEvent(event)` wouldn't have to be an else statement.

spireken (changes requested):

Applications/Spire/Source/Ui/OverlayPanel.cpp
```diff
@@ -140,10 +140,8 @@ bool OverlayPanel::event(QEvent* event) {
 }
 
 void OverlayPanel::keyPressEvent(QKeyEvent* event) {
-  if(event->key() == Qt::Key_Escape) {
-    if(!m_is_closed) {
-      fade(true);
-    }
+  if(event->key() == Qt::Key_Escape && !m_is_closed) {
```
spireken, line 143: This is even shorter than what I had in mind, but for the early return I meant something like this:

```
if(event->key() == Qt::Key_Escape && !m_is_closed) {
  fade(true);
  return;
}
```

spireken (approved):

kranar (approved):
Short and sweet. Perfect.

PR #634, opened 2021-06-09
Numeric filter panel

shaobinspire:


kranar:
This implementation deviates way too much from the spec.

Reimplement this in a way that follows the spec exactly and then resubmit the PR.

shaobinspire:
I removed something that is not listed in the specs from the implementation.

kranar:
Typing in 3.14 in the Min field doesn't work, the initial values are incorrect, and the UI Viewer should let me modify the min and max values.

PR #636, opened 2021-06-11
Numeric filter panel

shaobinspire:


Applications/Spire/Include/Spire/Ui/NumericFilterPanel.hpp
```diff
@@ -0,0 +1,274 @@
...
+
+      /** A ScalarValueModel over optional Type. */
+      using OptionalNumericModel = ScalarValueModel<boost::optional<Type>>;
+
+      /** A LocalScalarValueModel over optional Type. */
+      using LocalOptionalNumericModel =
+        LocalScalarValueModel<boost::optional<Type>>;
+
+      /**
+       * Signals that the minimum value and the maximum value for the filter.
```
spireken, line 44: Is this supposed to say "...for the filter was submitted."?

Applications/Spire/Include/Spire/Ui/NumericFilterPanel.hpp
```diff
@@ -0,0 +1,274 @@
...
+      boost::optional<Type> default_max_value,
+      QString title, QWidget* parent)
+      : QWidget(parent),
+        m_min_model(std::move(min_model)),
+        m_max_model(std::move(max_model)),
+        m_default_min_value(std::move(default_min_value)),
+        m_default_max_value(std::move(default_max_value)) {
+    if(m_default_min_value && m_default_max_value &&
+        *m_default_min_value > *m_default_max_value) {
+      std::swap(m_default_min_value, m_default_max_value);
```
spireken, line 139: Is it always the case that these should be swapped, instead of an error being raised? It seems that if this condition is true then there's an error somewhere.

spireken (commented):
Looks very good.

Applications/Spire/Include/Spire/Ui/NumericFilterPanel.hpp
shaobinspire: I didn't see any errors from here. The VS output window may display an exception which is raised by the parse_decimal function in UiViewer which tries to convert a string to a decimal when the default min/max value is empty, but this output is not caused by the swap.

kranar: As a follow up, do not mask or cover up for input or programming errors. If a mistake is made try to crash immediately. This promotes a type of local reasoning where errors in the program can be identified within a constant amount of time instead of having to reason globally about an application to track down a problem.

Applications/Spire/Include/Spire/Ui/NumericFilterPanel.hpp
```diff
@@ -0,0 +1,274 @@
...
+    layout->addSpacing(scale_height(10));
+    auto max_input = make_input_field(m_max_model);
+    auto max_layout = make_row_layout(tr("Max"), max_input);
+    layout->addLayout(max_layout);
+    m_filter_panel = new FilterPanel(std::move(title), this, parent);
+    m_filter_panel->connect_reset_signal([=] {
+      reset();
+    });
+    min_input->connect_submit_signal([=] (const auto& value) {
+      if(m_max_value && value && *value > *m_max_value) {
```
kranar, line 156: The constructor shall only be used to initialize an object, do not implement anything inside of it other than what is needed to fully initialize an object. It is very hard to reason about code that mixes in several different responsibilities and layers together.

Applications/Spire/Include/Spire/Ui/NumericFilterPanel.hpp
```diff
@@ -0,0 +1,274 @@
...
+      m_filter_panel->show();
+    }
+    return QWidget::event(event);
+  }
+
+  template<typename T>
+  typename NumericFilterPanel<T>::NumericBox*
+      NumericFilterPanel<T>::make_input_field(
+        std::shared_ptr<OptionalNumericModel> model) {
+    auto field = [=] {
```
kranar, line 227: Only use ```[=]``` when the closure outlives its scope.

Applications/Spire/Include/Spire/Ui/NumericFilterPanel.hpp
```diff
@@ -0,0 +1,274 @@
...
+  bool NumericFilterPanel<T>::event(QEvent* event) {
+    if(event->type() == QEvent::ShowToParent) {
+      reset();
+      m_filter_panel->show();
+    }
+    return QWidget::event(event);
+  }
+
+  template<typename T>
+  typename NumericFilterPanel<T>::NumericBox*
```
kranar, line 224: This is declared to be an instance method, but it does not make use of ```this```. Use ```static``` for methods that are not tied to any instance.

Applications/Spire/Include/Spire/Ui/NumericFilterPanel.hpp
```diff
@@ -0,0 +1,274 @@
...
+      NumericFilterPanel<T>::make_input_field(
+        std::shared_ptr<OptionalNumericModel> model) {
+    auto field = [=] {
+      if constexpr(std::is_same_v<NumericBox, DurationBox>) {
+        return new NumericBox(model);
```
kranar, line 229: Use move semantics.

Applications/Spire/Include/Spire/Ui/NumericFilterPanel.hpp
```diff
@@ -0,0 +1,274 @@
...
+      return new NumericBox(model, std::move(modifiers));
+    }();
+    field->setFixedSize(scale(120, 26));
+    return field;
+  }
+
+  template<typename T>
+  QHBoxLayout* NumericFilterPanel<T>::make_row_layout(QString label_name,
+      NumericBox* input_field) {
+    auto layout = new QHBoxLayout();
```
kranar, line 245: See above about the use of ```static```.

kranar (changes requested):
This looks good but it still doesn't adhere to the spec. Ensure that the implementation follows the spec exactly and then resubmit the PR.

kranar:
Good progress on this but it is absolutely critical that the spec is followed, do not substitute your own interpretation for what's in the spec. For a simple component like this we can get away with it, but as things get more complicated, deviating from the spec will create a snowball effect where components will not work well together.

shaobinspire:
I tried my best to understand the user flow and reimplement it. Sebastian helped me to clear something up and review the updated NumericFilterPanel. I hope that I can understand correctly this time.

kranar:
Still doesn't satisfy the spec, give it one last shot and if it still doesn't work then that's fine, I'll go ahead and implement it.

shaobinspire:
I think the only part that may need to change is when the default values are reset and the min and max values changes, the "Filter" signal will be triggered twice, but it conflicts with what Sebastian explained to me.

kranar:
Sure, as long as it follows the specification that's all that matters. Sebastian might forget, or he might make a mistake, or he might have a different opinion than another designer. Ultimately the implementation must match the spec exactly. If you have issues with the spec, you find a bug or a problem, feel free to point it out in Asana or Discord.

shaobinspire:
I took it another try. Hope it works.

kranar:
Still not valid. The initialization is incorrect, it keeps initializing to default values which is not what the spec says and is also not the intended behavior, plus there are other issues with the min/max updating. I'll go ahead and implement it at this point.

I would look into things like FSMs to get a solid understanding of how to translate a state machine description into raw code. The implementation as I see it is very ad-hoc, whereas the user flow can be modelled directly as a finite state machine.

PR #638, opened 2021-06-29
Delete icon button

spireken:


shaobinspire (changes requested):
There is a border when the button gets focus, but I don't see it in the specs of the DeleteIconButton.

Applications/Spire/Source/Ui/Button.cpp
```diff
@@ -101,6 +101,20 @@ Button* Spire::make_icon_button(QImage icon, QWidget* parent) {
...
 }
 
+Button* Spire::make_delete_icon_button(QWidget* parent) {
+  auto button = make_icon_button(imageFromSvg(":/Icons/delete.svg",
+    scale(16, 16)), parent);
+  auto style = get_style(*button);
+  style.get(Body()).
+    set(BackgroundColor(QColor::fromRgb(0, 0, 0, 0))).
+    set(Fill(QColor::fromRgb(0xC8, 0xC8, 0xC8)));
+  style.get(Focus() / Body() || Hover() / Body()).
```
kranar, line 111: Factor this out ```(Focus() || Hover()) / Body()```

spireken:
> There is a border when the button gets focus, but I don't see it in the specs of the DeleteIconButton.

The DeleteIconButton is an IconButton so I believe it inherits the IconButton style.

shaobinspire (approved):

kranar:
I believe work still needs to be done on this after some changes to icon, resubmit it later.

PR #639, opened 2021-06-30
The selected state of the ListItem.

shaobinspire:


spireken (approved):
Looks good.

Applications/Spire/Source/Ui/ListItem.cpp
```diff
@@ -9,7 +9,8 @@ using namespace Spire;
 using namespace Spire::Styles;
 
 ListItem::ListItem(QWidget* component, QWidget* parent)
-    : QWidget(parent) {
+    : QWidget(parent),
+      m_is_selected(false) {
   setFocusPolicy(Qt::StrongFocus);
   auto layout = new QHBoxLayout(this);
   layout->setContentsMargins({});
```
kranar, line 16: The specs were updated to place a ```Box``` inside of the ```Button``` but this implementation isn't doing that.

Applications/Spire/Source/Ui/ListItem.cpp
```diff
@@ -18,15 +19,32 @@ ListItem::ListItem(QWidget* component, QWidget* parent)
...
-    BackgroundColor(QColor::fromRgb(0xFF, 0xFF, 0xFF)));
+  style.get(Body()).
+    set(BackgroundColor(QColor::fromRgb(0xFF, 0xFF, 0xFF))).
+    set(border(scale_width(1), QColor::fromRgb(0, 0, 0, 0))).
+    set(horizontal_padding(scale_width(8)));
   style.get(Hover() / Body()).set(
     BackgroundColor(QColor::fromRgb(0xF2, 0xF2, 0xFF)));
   style.get(Focus() / Body()).set(
-    BackgroundColor(QColor::fromRgb(0x68, 0x4B, 0xC7)));
+    border(scale_width(1), QColor::fromRgb(0x4B, 0x23, 0xA0)));
```
kranar, line 29: This is styling both the size and the color, but the spec only styles the color.

Applications/Spire/Source/Ui/ListItem.cpp
```diff
@@ -18,15 +19,32 @@ ListItem::ListItem(QWidget* component, QWidget* parent)
...
 }
 
+bool ListItem::is_selected() const {
+  return m_is_selected;
+}
+
+void ListItem::set_selected(bool is_selected) {
+  m_is_selected = is_selected;
+  if(m_is_selected) {
+    match(*m_button, Selected());
```
kranar, line 42: Is is the ```ListItem``` that matches the selector, not the ```Button```.

kranar (changes requested):

Applications/Spire/Source/Ui/ListItem.cpp
```diff
@@ -8,23 +8,50 @@ using namespace boost::signals2;
...
 ListItem::ListItem(QWidget* component, QWidget* parent)
-    : QWidget(parent) {
+    : QWidget(parent),
+      m_is_selected(false) {
   setFocusPolicy(Qt::StrongFocus);
   auto layout = new QHBoxLayout(this);
   layout->setContentsMargins({});
-  m_button = new Button(component, this);
+  auto box = new Box(component);
+  box->setDisabled(true);
```
kranar, line 35: Where in the spec is this specified?

kranar (changes requested):

Applications/Spire/Source/Ui/ListItem.cpp
shaobinspire: If the box is not disabled, the sub-component in the box will get focus even though the component is disabled. For example, when the component is a disabled and read-only TextBox, the TextBox still can get focus by pressing Tab key for some reason.

Applications/Spire/Source/Ui/ListItem.cpp
```diff
@@ -8,23 +8,50 @@ using namespace boost::signals2;
...
-  auto style = get_style(*m_button);
-  style.get(Body()).set(
-    BackgroundColor(QColor::fromRgb(0xFF, 0xFF, 0xFF)));
-  style.get(Hover() / Body()).set(
-    BackgroundColor(QColor::fromRgb(0xF2, 0xF2, 0xFF)));
-  style.get(Focus() / Body()).set(
-    BackgroundColor(QColor::fromRgb(0x68, 0x4B, 0xC7)));
-  set_style(*m_button, std::move(style));
+  proxy_style(*this, *box);
+  set_style(*this, DEFAULT_STYLE());
```
kranar, line 41: This is styling the entire component, but the spec is styling only the ```Box```.

Applications/Spire/Source/Ui/ListItem.cpp
kranar: Okay, that sounds good, look at fixing the issue with the ```Box```'s focus proxy.

shaobinspire: I wrote a simple demo using Qt Creator to simulate the case: The Button class with the focus policy Qt::StrongFocus, and the Box class with the focus proxy on a disabled QLineEdit. The disabled QLineEdit still can get focus if the Button contains the Box and the Box holds the QLineEdit, just like the ListItem. I believe that using the Qt::StrongFocus policy in the Button and the focus proxy in the Box together could cause the problem. When either not using the Button (no Qt::StrongFocus policy) or using the Button but removing setFocusProxy from the Box (no focus proxy), the disabled QLineEdit can't get focus.

Applications/Spire/Source/Ui/ListItem.cpp
shaobinspire: The style proxy of the ListItem is the Box, so the styling is actually applying on the Box.

kranar: A proxy is one-way, so styling the ```ListItem``` styles the ```Box``` but styling the ```Box``` does not style the ```ListItem```. This is an important detail since a parent of ```ListItem``` can override ```ListItem```'s style using ```set_stylesheet``` which will completely wipe out all of the intended styling for the ```Box```.

Applications/Spire/Source/Ui/ListItem.cpp
kranar: Okay this is a good catch, is there a proposal for how to modify ```Box``` or ```Button``` or some other modification to fix this issue?

shaobinspire: I don't prefer to modify Box or Button. I try to fix the issue by changing the focus proxy of the ListItem. When the sub-component is disabled, the ListItem's focus proxy is button and the box's focus proxy is none to prevent the disabled component from getting focus; when the sub-component is enabled, it will be set to the focus proxy of the ListItem, for example, there are two sub-components in the CheckBpx (a Icon and a TextBox), if the Checkbox is not the focus proxy, the focus will be on the button first and then on the Icon in turn by pressing Tab continually.

kranar:
Looks good, well done.

PR #641, opened 2021-07-02
Button press

spireken:
The demo is in the 'label_button_updates' folder.


Applications/Spire/Source/UiViewer/StandardUiProfiles.cpp
```diff
@@ -748,6 +731,33 @@ UiProfile Spire::make_key_tag_profile() {
...
+      auto& label = get<QString>("label", profile.get_properties());
+      auto button = make_label_button(label.get());
+      apply_widget_properties(button, profile.get_properties());
+      auto& pressed_color = get<QColor>("pressed-color",
+        profile.get_properties());
+      pressed_color.connect_changed_signal([=] (const auto& color) {
+        auto style = get_style(*button);
+        style.get(Press() / Body()).
+          set(BackgroundColor(color));
+        set_style(*button, style);
```
shaobinspire, line 752: Use std::move for style.

shaobinspire (changes requested):

shaobinspire (commented):
It looks like there is a different behavior between the domo under the demos\button_press folder and the one which I build from code.  The demo still has the hover styling when I press the left button and hold it on the button, at the same time I press Alt+Tab twice to make UiViewer to be not active and to be active agian, but the one built from code loses the hover styling after these operations.

spireken:
@shaobinspire I updated the demo, the timestamp is 12:18PM.

shaobinspire (approved):

Applications/Spire/Source/Ui/Button.cpp
```diff
@@ -37,6 +37,8 @@ void Button::focusOutEvent(QFocusEvent* event) {
   if(event->reason() != Qt::PopupFocusReason && m_is_down) {
     m_is_down = false;
   }
+  unmatch(*this, Press());
+  unmatch(*this, Hover());
```
kranar, line 41: What does line 41 correspond to in the spec?

Applications/Spire/Source/Ui/Button.cpp
```diff
@@ -71,18 +75,33 @@ void Button::keyReleaseEvent(QKeyEvent* event) {
   }
 }
 
+void Button::mouseMoveEvent(QMouseEvent* event) {
+  if(!rect().contains(event->pos())) {
```
kranar, line 79: What does this correspond to in the spec?

Applications/Spire/Source/Ui/Button.cpp
```diff
@@ -71,18 +75,33 @@ void Button::keyReleaseEvent(QKeyEvent* event) {
...
 
 void Button::mouseReleaseEvent(QMouseEvent* event) {
-  if(m_is_down && event->button() == Qt::LeftButton &&
-      rect().contains(event->pos())) {
-    m_is_down = false;
-    m_clicked_signal();
+  if(event->button() == Qt::LeftButton) {
+    unmatch(*this, Press());
+    if(rect().contains(event->pos())) {
+      match(*this, Hover());
```
kranar, line 99: I don't see anything in the spec referencing ```Hover()```.

Applications/Spire/Source/Ui/Button.cpp
```diff
@@ -116,6 +135,8 @@ Button* Spire::make_label_button(const QString& label, QWidget* parent) {
   style.get(Hover() / Body()).
```
kranar, line 135: The spec styles ```hover or Button.press``` but the implementation is styling ```Hover``` separately from ```Press```.

kranar (changes requested):

Applications/Spire/Source/Ui/Button.cpp
spireken: It doesn't; this should be moved to the Box.

Applications/Spire/Source/Ui/Box.cpp
```diff
@@ -94,6 +95,35 @@ QSize Box::sizeHint() const {
   return *m_size_hint;
 }
 
+void Box::focusOutEvent(QFocusEvent* event) {
+  m_is_mouse_down = false;
```
kranar, line 99: What part of the spec does this correspond to?

Applications/Spire/Source/Ui/Button.cpp
```diff
@@ -113,7 +121,7 @@ Button* Spire::make_label_button(const QString& label, QWidget* parent) {
     set(TextColor(QColor::fromRgb(0, 0, 0))).
```
kranar, line 121: This text color is not in the specs, remove it from this.

Applications/Spire/Source/Ui/Button.cpp
```diff
@@ -113,7 +121,7 @@ Button* Spire::make_label_button(const QString& label, QWidget* parent) {
...
     set(border(scale_width(1), QColor::fromRgb(0, 0, 0, 0))).
     set(horizontal_padding(scale_width(8)));
-  style.get(Hover() / Body()).
+  style.get((Hover() || Press()) / Body()).
     set(BackgroundColor(QColor::fromRgb(0x4B, 0x23, 0xA0))).
     set(TextColor(QColor::fromRgb(0xFF, 0xFF, 0xFF)));
   style.get(Focus() / Body()).set(
```
kranar, line 127: The disabled styling only changes the text_color, not the background color, remove that from the implementation.

kranar (changes requested):

Applications/Spire/Source/Ui/Button.cpp
spireken: I can remove it for now but the proper text color will depend on the inclusion of the new Label component.

Applications/Spire/Source/Ui/Box.cpp
```diff
@@ -94,6 +95,34 @@ QSize Box::sizeHint() const {
...
```
kranar, line 99: What part of the spec involves anything to do with focus out/blur?

kranar:
Do what you can to implement this in a way that satisfies the spec and then submit it once and only once more for review. Unfortunately this is not something that can be reviewed over and over and over again, if it's too challenging to implement this properly then this task will be closed.

spireken:
I removed the stored mouse state because it had no use per the spec. Qt ties the mouse position relative to the last widget that was clicked (regardless of which button), so left clicking on a different component then dragging onto the Box/Button won't trigger the hover effect. This technically goes against the Box hover flow, but I think this issue goes beyond the scope of this case.

Another issue is the default text color which will require the new Label component.

kranar (approved):

Applications/Spire/Source/Ui/Box.cpp
```diff
@@ -94,6 +95,34 @@ QSize Box::sizeHint() const {
...
+    unmatch(*this, Hover());
+  } else if(m_is_mouse_down) {
+    match(*this, Hover());
+  }
+  QWidget::mouseMoveEvent(event);
+}
+
+void Box::mousePressEvent(QMouseEvent* event) {
+  if(event->button() == Qt::LeftButton) {
+    m_is_mouse_down = true;
```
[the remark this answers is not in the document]
spireken: This comment is about an old version, right?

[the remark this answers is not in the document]
kranar: Yes, the PR has been approved and merged.

PR #642, opened 2021-07-02
Label

shaobinspire:


spireken:
@shaobinspire Do you have a demo?

shaobinspire:
> @shaobinspire Do you have a demo?

Yes, the demo is in the demos\label.

spireken (approved):
I think my client is just slow so I didn't see the demo at first.

Applications/Spire/Source/Ui/TextBox.cpp
```diff
@@ -480,3 +480,14 @@ void TextBox::on_style() {
...
   });
 }
+
+TextBox* Spire::make_label(QString label, QWidget* parent) {
+  auto text_box = new TextBox(std::move(label), parent);
+  text_box->setDisabled(true);
+  text_box->set_read_only(true);
+  auto style = get_style(*text_box);
+  style.get(ReadOnly() && Disabled()).
+    set(TextColor(QColor::fromRgb(0, 0, 0)));
```
kranar, line 490: The spec does not use ReadOnly && Disabled as a selector.

kranar (changes requested):

Applications/Spire/Source/Ui/TextBox.cpp
shaobinspire: The default text color of the disabled TextBox is #C8C8C8, so I think the text_color in the spec is for the disabled TextBox.

kranar: Don't ever substitute your own opinion for what's in the spec. If you have a disagreement about the spec or need clarification then feel free to let the designer know during review, discuss it with them, share ideas and work as a team. But don't ever, under any circumstance, override the intent and work of the designers with your own opinion.

shaobinspire: I discussed it with Sebastian, he said that the text within Label should always be black by default. The text_color is the property of the TextBox, and text_color could have the different colors in different states. In the Label, the TextBox is disabled and read-only, so the text_color represents the text color in the disabled and read-only state. Even though the spec doesn't use the ReadOnly && Disabled, the state of the TextBox implicitly tells us the text_color is the color in the ReadOnly & Disabled state.

kranar: There is nothing implicit, if the spec says that you are to style a component in a certain way, then you MUST style it that way without exception. If you want to propose a change to the spec you are welcome to do so, but your PR will always get rejected until it satisfies the spec.

shaobinspire: The implementation meets the specs now, since the new spec has been approved.

kranar (approved):

PR #643, opened 2021-07-03
Radio button

spireken:


kranar:
The styling on the checkbox is not correct.

Applications/Spire/Source/Ui/Icon.cpp
```diff
@@ -60,6 +60,12 @@ void Icon::on_style() {
           update();
         });
       },
+      [&] (const IconImage& image) {
+        stylist.evaluate(image, [=] (auto image) {
+          m_icon = image;
```
kranar, line 65: Use a move.

kranar (changes requested):

PR #646, opened 2021-07-06
Icon button updates

spireken:


Applications/Spire/Source/Ui/Box.cpp
```diff
@@ -62,18 +62,6 @@ QSize Box::sizeHint() const {
     if(m_size_hint->isValid()) {
```
kranar, line 62: Undo all changes to ```Box```.

Applications/Spire/Source/Ui/Tooltip.cpp
```diff
@@ -10,7 +10,7 @@ namespace {
     auto style = StyleSheet();
     style.get(Any()).
       set(BackgroundColor(QColor::fromRgb(0, 0, 0))).
-      set(border_size(0));
+      set(border_color(QColor::fromRgb(0, 0, 0)));
```
kranar, line 13: What part of the spec does this correspond to? I don't see any border color styling and the border size is supposed to be 0.

kranar (changes requested):
Overall good.

Applications/Spire/Source/Ui/Box.cpp
spireken: I can't have the IconButton conform to the spec without these changes.

kranar: Okay, I will remove these changes in a local branch and see what potential issues this causes.

kranar: I don't see any issue, what property is violated when Box is reverted to spire2018?

spireken: The issue is the one I outlined in the case: "...if the icon is 26x26, the IconButton is expected to have a default size of 26x26. With the implementation of the Box, the size hint of the IconButton for a 26x26 Icon would be 28x28."

The specs use inset borders, so even aside from this issue I'm not sure what the intent is by expanding the Box by the size of its borders.

kranar: There is no such thing as a default size, the icon is however big its parent layout is. If the parent layout is 10000x10000, then the icon is 10000x10000.

spireken: Isn't the green sizing in the layouts considered a sort of 'default' size, which would be the size of the widget with no limitations and no explicitly specified size?

kranar: A layout labeled ```<X>``` with two parallel green borders means the size of the region between the two parallel borders is equal to the size of the ```X```'s layout across the same dimension. So if a layout labelled ```<X>``` has green horizontal borders, then that means the width of ```<X>``` is the width of ```X```'s layout.

spireken: I see, I interpreted the last component (the Button) in IconButton's layout file to mean that the IconButton had some intrinsic size, determined by the contents of the Button. But, that size is for the Button *in* the IconButton, but not the IconButton itself.

kranar: Oh gotcha. We are very close to fully formalizing layouts, perhaps it would be best to stamp out all remaining details in this Thursday's meeting to address remaining corner cases and really nail everything down.

kranar (approved):

PR #661, opened 2021-07-24
Added the selection model to the ListView.

shaobinspire:


spireken (approved):
Looks good.

Applications/Spire/Source/Ui/ListView.cpp
```diff
@@ -108,8 +124,9 @@ ListView::SelectionMode ListView::get_selection_mode() const {
 
 void ListView::set_selection_mode(SelectionMode selection_mode) {
   m_selection_mode = selection_mode;
-  if(m_selection_mode == SelectionMode::NONE && m_selected.has_value()) {
-    select_item(false);
+  if(m_selection_mode == SelectionMode::NONE &&
```
kranar, line 127: This isn't in the spec, remove it.

Applications/Spire/Source/Ui/ListView.cpp
```diff
@@ -260,11 +273,13 @@ QLayoutItem* ListView::get_column_or_row(int index) {
   return get_layout()->itemAt(index);
 }
 
-void ListView::select_item(bool is_selected) {
+void ListView::select_item(const boost::optional<std::any>& selection) {
   for(auto i = 0; i < m_list_model->get_size(); ++i) {
-    if(m_list_model->at(i).type() == m_selected.type() &&
-        displayTextAny(m_list_model->at(i)) == displayTextAny(m_selected)) {
-      m_items[i].m_item->set_selected(is_selected);
+    if(selection && m_list_model->at(i).type() == (*selection).type() &&
```
kranar, line 278: ```selection->type()```

Applications/Spire/Source/Ui/ListView.cpp
shaobinspire: The purpose is when the selection mode becomes NONE from SINGLE, the selected item should clear the selected state. Otherwise, there may be a selected item even though the selection mode is NONE.

kranar: What part of the spec says that the selected item should clear?

shaobinspire: Specs don't say anything about switching selection mode. I will remove it.

kranar: In the future if you see something you think is wrong, you are always welcome to bring it up to the UX designer or to me. There could definitely be issues with the spec, bugs, or you are welcome to express your opinion and provide insight for us to learn from, but without bringing the issue up I will usually insist that things be implemented according to the spec.

In this particular case, selection mode only determines how the user interacts with the list, it doesn't determine what is selected or not selected. So for example it's possible to have a read-only selection by setting the selection mode to NONE. There could still be selected items but the user will have no way to modify what is or isn't selected.

kranar (approved):

kranar:
Resolve merge conflicts.

PR #662, opened 2021-07-26
Set the default minimum value of the QuantityBox to 0.

shaobinspire:


Applications/Spire/Source/Ui/QuantityBox.cpp
```diff
@@ -8,6 +9,10 @@ namespace {
   struct QuantityToDecimalModel : ToDecimalModel<Quantity> {
     using ToDecimalModel<Quantity>::ToDecimalModel;
 
+    optional<Decimal> get_minimum() const override {
```
kranar, line 12: This is too narrow of a change. A ```QuantityBox``` that needs a non-negative value (aka unsigned) needs to use an unsigned model.

shaobinspire: I am little bit confused. Do you mean replace the Nexus::Quantity with unsigned int? Just like IntegerBox, the only difference is that the template parameter of the DecimalBoxAdaptor is unsigned int.

kranar: Any restrictions on QuantityBox's value shall be implemented in the model that is passed into the QuantityBox's constructor. A QuantityBox by itself does not restrict values, only display values, it is a view class. The data is part of the model and any properties of that data shall be implemented in the model.

kranar:
This is way off and once again, making changes to the API without proposing them is taking a HUGE risk and will almost certainly result in wasted time and having those changes get rejected.

PR #665, opened 2021-07-27
The model for the unsigned Quantity value in the QuantityBox

shaobinspire:


Applications/Spire/Include/Spire/Ui/QuantityBox.hpp
```diff
@@ -41,6 +41,9 @@ namespace Spire {
       QuantityBox(std::shared_ptr<OptionalQuantityModel> model,
         QHash<Qt::KeyboardModifier, Nexus::Quantity> modifiers,
         QWidget* parent = nullptr);
+
+    private:
+      struct UnsignedQuantityModel;
```
spireken, line 46: If this is only used by the QuantityBox source file, can it just be declared and defined in an anonymous namespace in the QuantityBox.cpp file?

Applications/Spire/Source/Ui/QuantityBox.cpp
```diff
@@ -14,14 +16,65 @@ namespace {
...
+    }
+    return m_model->set_current(value);
+  }
+
+  connection connect_current_signal(
+      const CurrentSignal::slot_type& slot) const override {
+    return m_current_signal.connect(slot);
+  }
+
+  void on_current(const optional<Quantity>& current) {
```
spireken, line 63: Is there a reason this method and the other members (connection, model, etc.) aren't private?

Applications/Spire/Include/Spire/Ui/QuantityBox.hpp
shaobinspire: Because it is only used by the QuantityBox, it is declared the nested class of the QuantityBox, which works like DecimalBox::DecimalToTextModel.

Applications/Spire/Source/Ui/QuantityBox.cpp
shaobinspire: I think the class is only used by the QuantityBox and its definition is in the source file, so other classes can't access it. In this case, there is no difference between private and public.

Applications/Spire/Include/Spire/Ui/QuantityBox.hpp
spireken: Okay, I see. I didn't notice there was a precedent.

spireken (approved):

Applications/Spire/Include/Spire/Ui/QuantityBox.hpp
kranar, line 46: Put the definition in its own header file.

Applications/Spire/Source/Ui/QuantityBox.cpp
```diff
@@ -1,5 +1,7 @@
 #include "Spire/Ui/QuantityBox.hpp"
 
+using namespace boost;
```
kranar, line 3: Implement the model in its own source file.

Applications/Spire/Source/Ui/QuantityBox.cpp
```diff
@@ -14,14 +16,65 @@ namespace {
...
+
+  UnsignedQuantityModel(std::shared_ptr<OptionalQuantityModel> model)
+    : m_model(std::move(model)),
+      m_current_connection(m_model->connect_current_signal(
+        [=] (const auto& current) {
+          on_current(current);
+        })) {}
+
+  optional<Quantity> get_minimum() const override {
+    return 0;
```
kranar, line 32: This is too permissive of a condition. When implementing models that add constraints, you must respect the constraints of the original model in addition to any new constraints.

Applications/Spire/Source/Ui/QuantityBox.cpp
```diff
@@ -14,14 +16,65 @@ namespace {
...
+  QValidator::State get_state() const override {
+    return m_model->get_state();
+  }
+
+  const optional<Quantity>& get_current() const override {
+    return m_model->get_current();
+  }
+
+  QValidator::State set_current(const optional<Quantity>& value) override {
+    if(value && *value < 0) {
```
kranar, line 52: Never suppress errors, all this does is hide an issue and makes it incredibly hard to debug and identify the cause of a problem. When a precondition is violated, use the most appropriate mechanism available to report the issue as soon as possible so it can be handled locally.

Applications/Spire/Source/Ui/QuantityBox.cpp
```diff
@@ -14,14 +16,65 @@ namespace {
...
+
 QuantityBox::QuantityBox(QHash<Qt::KeyboardModifier, Quantity> modifiers,
   QWidget* parent)
   : QuantityBox(std::make_shared<LocalOptionalQuantityModel>(),
       std::move(modifiers), parent) {}
 
 QuantityBox::QuantityBox(std::shared_ptr<OptionalQuantityModel> model,
     QHash<Qt::KeyboardModifier, Quantity> modifiers, QWidget* parent)
-    : DecimalBoxAdaptor(model, std::make_shared<QuantityToDecimalModel>(model),
+    : DecimalBoxAdaptor(std::make_shared<UnsignedQuantityModel>(model),
```
kranar, line 75: The unsigned model is only used by default, other users of the ```QuantityBox``` are more than welcome to provide a model of their own.

kranar (changes requested):
Looks very good, big improvement over the previous proposals.

Applications/Spire/Include/Spire/Ui/QuantityBox.hpp
```diff
@@ -3,6 +3,7 @@
 #include "Nexus/Definitions/Quantity.hpp"
 #include "Spire/Ui/DecimalBoxAdaptor.hpp"
 #include "Spire/Ui/Ui.hpp"
+#include "Spire/Ui/UnsignedQuantityModel.hpp"
```
kranar, line 6: Only include what you use.

Applications/Spire/Source/Ui/UnsignedQuantityModel.cpp
```diff
@@ -0,0 +1,55 @@
...
+  }
+  return m_model->set_current(value);
+}
+
+connection UnsignedQuantityModel::connect_current_signal(
+    const CurrentSignal::slot_type& slot) const {
+  return m_current_signal.connect(slot);
+}
+
+void UnsignedQuantityModel::on_current(const optional<Quantity>& current) {
```
kranar, line 53: Factor this method out entirely.

Applications/Spire/Source/Ui/UnsignedQuantityModel.cpp
```diff
@@ -0,0 +1,55 @@
...
+optional<Quantity> UnsignedQuantityModel::get_minimum() const {
+  auto minimum = m_model->get_minimum();
+  if(!minimum || (minimum && *minimum < 0)) {
+    return 0;
+  }
+  return minimum;
+}
+
+optional<Quantity> UnsignedQuantityModel::get_maximum() const {
+  return m_model->get_maximum();
```
kranar, line 25: Handle the case where the maximum is negative (in which case this is models an empty quantity).

kranar (changes requested):
Much better, minor improvements remaining.

Applications/Spire/Source/Ui/UnsignedQuantityModel.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+optional<Quantity> UnsignedQuantityModel::get_minimum() const {
+  auto minimum = m_model->get_minimum();
+  if(!minimum || (minimum && *minimum < 0)) {
+    return 0;
+  }
+  return minimum;
+}
+
+optional<Quantity> UnsignedQuantityModel::get_maximum() const {
```
kranar, line 20: This is a bug.

kranar (changes requested):

PR #667, opened 2021-07-30
List view disabled items

shaobinspire:


Applications/Spire/Include/Spire/Ui/ListView.hpp
```diff
@@ -227,10 +227,14 @@ namespace Styles {
       int get_index_by_value(const std::any& value) const;
       QLayout* get_layout();
       QLayoutItem* get_column_or_row(int index);
+      bool is_column_or_row_disabled(QLayout* target, int begin, int end);
       void select_item(const boost::optional<std::any>& selection);
-      void cross_move(bool is_next);
-      int move_next();
-      int move_previous();
+      void cross_move(int position, int index, bool is_next, bool track_move);
```
spireken, line 232: Can track_move be removed? I see it passed to multiple functions but I can't see where it's used.

shaobinspire: You are right. It is redundant and should be removed.

spireken:
Can you remove the other instances of track_move? If I search for all instances of track_move in the ListView I see they're passed around but not evaluated or reassigned anywhere.

shaobinspire:
> Can you remove the other instances of track_move? If I search for all instances of track_move in the ListView I see they're passed around but not evaluated or reassigned anywhere.

No, it can’t be removed. The track_move is used to determine whether the x and y should be updated. Based on the specs, the cross move (The move passes through different columns or rows) doesn’t update x and y, because x and y are used to tracked the path to get the ability to go back to the starting point with the same path,  while the normal move (A move doesn’t cross different columns or rows) should update x and y for each move. You can find the function update_current in uses the parameter on line 435.

I can remove track_move from the jump_move function, because it can’t be used for the cross move and track_move is always true.

spireken (approved):

kranar:
Aspects of this PR have been merged into spire2018.

PR #675, opened 2021-08-10
Region list item

shaobinspire:


Applications/Spire/Source/Ui/RegionListItem.cpp
```diff
@@ -0,0 +1,128 @@
...
+  layout->addWidget(name_label);
+}
+
+const Region& RegionListItem::get_region() const {
+  return m_region;
+}
+
+TextBox* RegionListItem::make_value_label() {
+  if(!m_region.GetSecurities().empty()) {
+    m_type = Type::SECURITY;
```
spireken, line 91: Semantically, I wouldn't expect a function called make_value_label to also assign the region type. It might be preferred to have a method to update/set the region type, then these if/else if statements can just check the value of m_type.

spireken (changes requested):
Looks good.

Applications/Spire/Source/Ui/RegionListItem.cpp
shaobinspire: Good point.

spireken (approved):

Applications/Spire/Include/Spire/Ui/RegionListItem.hpp
```diff
@@ -0,0 +1,42 @@
...
+       * @param region The region which can be a security, a market or a country.
+       * @param parent The parent widget.
+       */
+      explicit RegionListItem(Nexus::Region region, QWidget* parent = nullptr);
+
+      /** Returns the region. */
+      const Nexus::Region& get_region() const;
+
+    private:
+      enum class Type {
```
kranar, line 27: By default an ```enum class``` uses an ```int``` as its representation. When performance of reads/writes is insignificant, use a ```std::uint8_t``` to save on space.

Applications/Spire/Source/Ui/RegionListItem.cpp
```diff
@@ -0,0 +1,135 @@
...
+    : QWidget(parent),
+      m_region(std::move(region)),
+      m_type(get_type()) {
+  auto value_label = make_value_label();
+  value_label->setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Preferred);
+  set_style(*value_label, VALUE_LABEL_STYLE(get_style(*value_label)));
+  auto name_label = make_label(QString::fromStdString(m_region.GetName()));
+  name_label->setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Preferred);
+  set_style(*name_label, NAME_LABEL_STYLE(get_style(*name_label)));
+  auto type_icon = make_type_icon();
```
kranar, line 62: Declare things at the point of usage, keep code local.

Applications/Spire/Source/Ui/RegionListItem.cpp
```diff
@@ -0,0 +1,135 @@
...
+  auto value_container_layout = new QHBoxLayout();
+  value_container_layout->setContentsMargins({});
+  value_container_layout->setSpacing(0);
+  value_container_layout->addWidget(value_label);
```
kranar, line 66: Even here, there is a jump from value to name back to value. Keep code local. Local reasoning is a lot easier than global reasoning.

Applications/Spire/Source/Ui/RegionListItem.cpp
```diff
@@ -0,0 +1,135 @@
...
+  layout->setSpacing(0);
+  layout->addLayout(value_container_layout);
+  layout->addWidget(name_label);
+}
+
+const Region& RegionListItem::get_region() const {
+  return m_region;
+}
+
+RegionListItem::Type RegionListItem::get_type() {
```
kranar, line 89: A method that performs only ```const``` operations on ```this``` should be marked as ```const``` to indicate such.

It would be even better to remove this member function altogether and make it a free function.

Applications/Spire/Source/UiViewer/StandardUiProfiles.cpp
```diff
@@ -1154,6 +1156,45 @@ UiProfile Spire::make_radio_button_profile() {
     });
 }
 
+UiProfile Spire::make_region_list_item_profile() {
+  auto properties = std::vector<std::shared_ptr<UiProperty>>();
+  populate_widget_properties(properties);
+  auto type_property = define_enum<int>(
```
kranar, line 1162: This is a very neat use of ```define_enum```.

Applications/Spire/Source/UiViewer/StandardUiProfiles.cpp
```diff
@@ -1154,6 +1156,45 @@ UiProfile Spire::make_radio_button_profile() {
...
+          auto country_code = DefaultCountries::US();
+          auto region = Region(country_code);
+          region.SetName(
+            GetDefaultCountryDatabase().FromCode(country_code).m_name);
+          return region;
+        }
+      }();
+      auto item = new RegionListItem(region);
+      apply_widget_properties(item, profile.get_properties());
+      item->setMinimumSize(0, 0);
```
kranar, line 1191: Why is the minimum/maximum size set?

kranar (changes requested):

Applications/Spire/Source/UiViewer/StandardUiProfiles.cpp
shaobinspire: After rebuilding, the new component keeps the minimum/maximum size of the previous component. If the size of the rebuilt component is different from the old one, the component will not be resized to fit its content unless removing its minimum/maximum size.

Applications/Spire/Source/Ui/RegionListItem.cpp
shaobinspire: The RegionListItem::Type which is the return type of the function is inaccessible, so I think the free function is not good alternative.

Applications/Spire/Source/Ui/RegionListItem.cpp
```diff
@@ -53,40 +53,40 @@ RegionListItem::RegionListItem(Region region, QWidget* parent)
...
   value_container_layout->setSpacing(0);
+  auto value_label = make_value_label();
+  value_label->setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Preferred);
+  set_style(*value_label, VALUE_LABEL_STYLE(get_style(*value_label)));
   value_container_layout->addWidget(value_label);
+  auto type_icon_layout = new QVBoxLayout();
+  type_icon_layout->setContentsMargins({});
+  type_icon_layout->setSpacing(0);
+  type_icon_layout->addStretch();
+  auto type_icon = make_type_icon();
```
kranar, line 70: Even this can be made local:

    if(auto type_icon = make_type_icon()) {

Applications/Spire/Include/Spire/Ui/RegionListItem.hpp
```diff
@@ -33,7 +33,7 @@ namespace Spire {
       Nexus::Region m_region;
       Type m_type;
 
-      Type get_type();
+      Type get_type() const;
       TextBox* make_value_label();
```
kranar, line 37: When a conceptual issue is pointed out in a review, look to identify other instances of that issue as well and correct all of them. It's not always feasible in a review to point out every single instance of an issue verbatim.

kranar (changes requested):

PR #678, opened 2021-08-11
Text area box

spireken:


Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,634 @@
...
+    &TextAreaBox::on_text_changed);
+  m_placeholder = new ElidedLabel(this);
+  m_placeholder->setFixedSize(0, 0);
+  m_stacked_widget->addWidget(m_placeholder);
+  m_placeholder->setCursor(m_text_edit->cursor());
+  m_placeholder->setAttribute(Qt::WA_TransparentForMouseEvents);
+  m_scroll_box = new ScrollBox(m_stacked_widget, this);
+  m_scroll_box->set(ScrollBox::DisplayPolicy::NEVER,
+    ScrollBox::DisplayPolicy::ON_OVERFLOW);
+  m_scroll_box->get_vertical_scroll_bar().adjustSize();
```
kranar, line 202: What is the purpose of this ```adjustSize``` call?

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,634 @@
...
+  m_scroll_box->setSizePolicy(QSizePolicy::Minimum, QSizePolicy::Minimum);
+  m_scroll_box->installEventFilter(this);
+  m_scroll_box->findChild<ScrollableLayer*>()->installEventFilter(this);
```
kranar, line 205: This is almost certainly not a good idea, what is the reason for this?

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,634 @@
...
+  }
+  QWidget::changeEvent(event);
+}
+
+bool TextAreaBox::eventFilter(QObject* watched, QEvent* event) {
+  if(watched == m_text_edit && event->type() == QEvent::FocusOut) {
+    m_submission = m_model->get_current();
+    m_submit_signal(m_submission);
+  } else if(event->type() == QEvent::Wheel) {
+    if(m_text_edit->isReadOnly()) {
```
kranar, line 274: Why is wheeling disabled on read-only?

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,634 @@
...
+      return true;
+    }
+  }
+  return QWidget::eventFilter(watched, event);
+}
+
+void TextAreaBox::mousePressEvent(QMouseEvent* event) {
+  m_text_edit->setFocus();
```
kranar, line 282: Why is this needed? What happens if it's removed? Clicking on a widget should always give it focus by default.

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,634 @@
...
+void TextAreaBox::commit_placeholder_style() {
+  auto alignment = m_placeholder_styles.m_alignment.value_or(
+    Qt::Alignment(Qt::AlignmentFlag::AlignLeft));
+  m_placeholder->set_alignment(alignment);
+  auto font = m_placeholder_styles.m_font.value_or(QFont());
+  if(m_placeholder_styles.m_size) {
+    font.setPixelSize(*m_placeholder_styles.m_size);
+  }
+  m_placeholder->setFont(font);
+  m_placeholder->set_text_color(m_placeholder_styles.m_color.value_or(
```
kranar, line 300: What is the purpose of using an optional for color if the default color is ```QColor```? Better off using ```QColor``` and remove the use of optional.

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,634 @@
...
+void TextAreaBox::resizeEvent(QResizeEvent* event) {
+  update_layout();
+  QWidget::resizeEvent(event);
+}
+
+void TextAreaBox::commit_placeholder_style() {
+  auto alignment = m_placeholder_styles.m_alignment.value_or(
+    Qt::Alignment(Qt::AlignmentFlag::AlignLeft));
+  m_placeholder->set_alignment(alignment);
+  auto font = m_placeholder_styles.m_font.value_or(QFont());
```
kranar, line 295: See below about color and ```optional```.

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,634 @@
...
+  update_placeholder_text();
+}
+
+void TextAreaBox::commit_style() {
+  auto stylesheet = QString(
+    R"(QTextEdit {
+      background: transparent;
+      border-width: 0px;)");
+  m_text_edit_styles.m_styles.write(stylesheet);
+  auto alignment = m_text_edit_styles.m_alignment.value_or(
```
kranar, line 311: See above about placeholder color. If the default is ```AlignLeft``` then use that as ```m_text_edit_styles.m_alignment``` instead of an empty ```optional```.

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,634 @@
...
+}
+
+void TextAreaBox::update_display_text() {
+  if(is_read_only()) {
+    if(m_text_edit->toPlainText() != m_model->get_current()) {
+      m_text_edit->blockSignals(true);
+      m_text_edit->setText(m_model->get_current());
+      m_text_edit->blockSignals(false);
+    }
+    auto line_count = std::floor(static_cast<double>((height() -
```
kranar, line 355: Break up at higher AST depth.

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,634 @@
...
+      get_padding_size().height())) /
+      static_cast<double>(m_computed_line_height));
```
kranar, line 357: Only cast one value to a double, not both.

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,634 @@
...
+
+void TextAreaBox::update_display_text() {
+  if(is_read_only()) {
+    if(m_text_edit->toPlainText() != m_model->get_current()) {
+      m_text_edit->blockSignals(true);
+      m_text_edit->setText(m_model->get_current());
+      m_text_edit->blockSignals(false);
+    }
+    auto line_count = std::floor(static_cast<double>((height() -
+      get_padding_size().height())) /
```
kranar, line 356: Remove excessive parenthesis.

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,634 @@
...
+      static_cast<double>(m_computed_line_height));
+    if(line_count > 0) {
+      auto lines = [&] {
```
kranar, line 359: Remove the IIFE, it's not serving any purpose, line 360 completely invalidates the use of an IIFE.

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,634 @@
...
+        auto lines = QStringList();
+        for(auto i = 0; i < m_text_edit->document()->blockCount(); ++ i) {
```
kranar, line 361: No space after ```++```.

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,634 @@
...
+          auto block = m_text_edit->document()->findBlockByNumber(i);
+          if(block.isValid()) {
+            auto block_text = block.text();
+            for(int i = 0; i != block.layout()->lineCount(); ++i) {
+              auto line = block.layout()->lineAt(i);
+              lines.append(block_text.mid(line.textStart(), line.textLength()));
```
kranar, line 367: 80 column max.

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,634 @@
...
+            }
+            if(lines.count() > line_count) {
+              break;
+            }
+          }
+        }
+        return lines;
+      }();
+      if(lines.count() > line_count) {
+        auto is_elided = lines.count() > line_count;
```
kranar, line 380: This is fishy business, line 379 checks ```lines.count() > line_count```, and then line 380 computes this again, and then line 381 yet again. Fix this whole section up.

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,634 @@
...
+    get_padding_size());
+  m_stacked_widget->adjustSize();
+  update_display_text();
+  update_placeholder_text();
+}
+
+void TextAreaBox::update_line_height() {
+  m_computed_line_height =
+    static_cast<int>(static_cast<double>(m_text_edit->font().pixelSize()) *
+    *m_text_edit_styles.m_line_height);
```
kranar, line 439: If ```m_line_height``` is already a ```double```, then don't cast the ```pixelSize()``` expression to a ```double``` as well.

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,634 @@
...
+  }
+}
+
+void TextAreaBox::update_text_alignment() {
+  auto alignment = m_text_edit_styles.m_alignment.value_or(Qt::AlignLeft);
+  auto cursor_pos = m_text_edit->textCursor().position();
+  for(auto i = 0; i < m_text_edit->document()->blockCount(); ++i) {
+    auto block = m_text_edit->document()->findBlockByNumber(i);
+    if(block.isValid()) {
+      auto cursor = m_text_edit->textCursor();
```
kranar, line 464: This looks very similar to some code written above. See if you can extract the common code.

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,634 @@
...
+  cursor.setPosition(cursor_pos);
+  m_text_edit->setTextCursor(cursor);
+}
+
+void TextAreaBox::update_text_edit_width() {
+  auto border_size = get_border_size();
+  auto padding_size = get_padding_size();
+  if(m_text_edit->document()->size().toSize().height() > height() -
+      padding_size.height() - border_size.height() &&
+      !m_text_edit->isReadOnly()) {
```
kranar, line 480: When you have an ```A && B``` that share no dependencies, put the cheaper operation on the left-hand side so that the expression can short circuit without doing unnecessary work. ```foo.is_read_only``` is trivial and cheap and if it's not true then you can avoid all this calculation as opposed to the way it's structured now where a bunch of calculations are performed, and then when foo.is_read_only is evaluated, all of those calculations go to waste.

kranar (changes requested):
A lot of excellent work done, but still quite a bit more work needed. Keep it up.

Applications/Spire/Include/Spire/Ui/ScrollBox.hpp
```diff
@@ -66,6 +66,12 @@ namespace Spire {
       /** Sets the horizontal and vertical display policies. */
       void set(DisplayPolicy horizontal_policy, DisplayPolicy vertical_policy);
 
+      /** Returns the vertical ScrollBar. */
+      ScrollBar& get_vertical_scroll_bar();
```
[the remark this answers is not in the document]
spireken: How can I scroll the ScrollBox to show the cursor without an accessor for the ScrollBars?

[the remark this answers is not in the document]
kranar: Okay fine, work on everything else and then we can figure this one out.

Applications/Spire/Source/Ui/TextAreaBox.cpp
spireken: In update_text_edit_width(), it's possible that the scroll bar will be shown after the text edit's width is updated, but it's not currently visible. This adjustSize call makes sure that when width() is called that it returns the size the scroll bar would be if it was shown.

On reviewing it, I can move this call to right before the width is queried.

Applications/Spire/Source/Ui/TextAreaBox.cpp
spireken: I don't adjust the size of the text edit when it's read-only so it's possible to scroll it. I can probably clean this up and remove the call to findChild in the above comment by resizing the text edit when it's read-only.

kranar: Good, clean it up.

Applications/Spire/Source/Ui/TextAreaBox.cpp
spireken: Even with the focus proxy set the text edit isn't receiving focus when the TextAreabox is clicked. I'll dig into it a little more.

Applications/Spire/Source/Ui/TextAreaBox.cpp
kranar: If you want what the width would or "should" be, use the component's size hint. ```adjustSize``` is a very expensive call that produces size effects, ```sizeHint``` produces no side effects and is generally an ```O(1)``` operation.

Applications/Spire/Source/Ui/TextAreaBox.cpp
kranar: Sure, if there's a bug then create a task for it. Don't work around bugs without first reporting them, it results in everyone coming up with hacky ways of working around issues instead of solving it at the root.

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,634 @@
...
+          m_text_edit->fontMetrics().horizontalAdvance(ELLIPSES_CHAR));
+        if(is_elided && !last_line.endsWith(ELLIPSES_CHAR)) {
+          last_line.append(ELLIPSES_CHAR);
+        }
+        m_text_edit->blockSignals(true);
+        m_text_edit->setText(lines.join(""));
+        m_text_edit->blockSignals(false);
+      }
+    }
+    update_text_alignment();
```
shaobinspire, line 397: Line 397- 399 could be removed, and Line 403 - 404 can do it.

shaobinspire (changes requested):
When“abcd” is input into the TextAreaBox, the cursor can’t be put between ‘c’ and ‘d’ by the mouse, but it can be done using the arrow key.

spireken:
These commits don't address the ScrollBox's ScrollBar getters, nor the issue @shaobinspire pointed out which I think is related to the ScrollableLayer.

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,630 @@
...
+  update_layout();
+}
+
+connection TextAreaBox::connect_submit_signal(
+    const SubmitSignal::slot_type& slot) const {
+  return m_submit_signal.connect(slot);
+}
+
+QSize TextAreaBox::sizeHint() const {
+  return m_scroll_box->sizeHint();
```
shaobinspire, line 275: I think the sizeHint function is not necessary, because the m_scroll_box is placed in the layout of the TextAreaBox and Qt can handle it automatically.

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,630 @@
...
+      static_cast<double>(m_computed_line_height));
+    auto is_elided = false;
+    if(line_count > 0) {
+      auto lines = QStringList();
+      for(auto i = 0; i < m_text_edit->document()->blockCount() && !is_elided;
+          ++i) {
+        auto block = m_text_edit->document()->findBlockByNumber(i);
+        if(block.isValid()) {
+          auto block_text = block.text();
+          for(int i = 0; i != block.layout()->lineCount(); ++i) {
```
shaobinspire, line 381: The inner loop and the outer loop have the same iterator name "i", even though it is allowed by the standard, I don't think it's a good practice.

shaobinspire (changes requested):

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,630 @@
...
+#include <QTextBlock>
+#include <QTextDocument>
+#include "Spire/Spire/Dimensions.hpp"
+#include "Spire/Spire/LocalValueModel.hpp"
+#include "Spire/Ui/ScrollableLayer.hpp"
+#include "Spire/Ui/ScrollBar.hpp"
+#include "Spire/Ui/ScrollBox.hpp"
+
+using namespace boost;
+using namespace boost::posix_time;
```
shaobinspire, line 14: I don't see where the namespace is used.

spireken:
@shaobinspire There was an update to the ScrollableLayer in spire2018 that fixed the issue you pointing out with selecting the right side text. It's merged into this branch now.

shaobinspire (approved):
Looks good.

Applications/Spire/Include/Spire/Ui/TextAreaBox.hpp
```diff
@@ -0,0 +1,124 @@
...
+      ElidedLabel* m_placeholder;
+      StyleProperties m_placeholder_styles;
+      ScrollBox* m_scroll_box;
+      boost::signals2::scoped_connection m_current_connection;
+      QString m_submission;
+      QString m_placeholder_text;
+      int m_computed_line_height;
+
+      void apply_block_formatting(
+        const std::function<void(const QTextBlock& block)> formatting);
```
kranar, line 103: Missing ```&```.

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,627 @@
...
+    style.get(Placeholder()).set(TextColor(QColor::fromRgb(0xA0, 0xA0, 0xA0)));
+    style.get(Disabled() / Placeholder()).
+      set(TextColor(QColor::fromRgb(0xC8, 0xC8, 0xC8)));
+    return style;
+  }
+}
+
+class TextAreaBox::ContentSizedTextEdit : public QTextEdit {
+  public:
+
```
kranar, line 57: Remove vertical space.

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,627 @@
...
+      set(BackgroundColor(QColor::fromRgb(0, 0, 0, 0))).
+      set(border_color(QColor::fromRgb(0, 0, 0, 0)));
+    style.get(Placeholder()).set(TextColor(QColor::fromRgb(0xA0, 0xA0, 0xA0)));
+    style.get(Disabled() / Placeholder()).
+      set(TextColor(QColor::fromRgb(0xC8, 0xC8, 0xC8)));
+    return style;
+  }
+}
+
+class TextAreaBox::ContentSizedTextEdit : public QTextEdit {
```
kranar, line 55: Why is a ```QTextEdit``` used instead of a ```QPlainTextEdit```?

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,627 @@
...
+    QSize minimumSizeHint() const override {
+      return QSize();
+    }
+
+  private:
+    int m_longest_line_width;
+
+    void on_text_changed() {
+      m_longest_line_width = get_longest_line_width();
+      updateGeometry();
```
kranar, line 94: What if nothing has changed? Calling ```updateGeometry``` is not cheap.

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,627 @@
...
+            std::max(longest, fontMetrics().horizontalAdvance(block.text()));
+        }
+      }
+      return longest;
+    }
+};
+
+class TextAreaBox::ElidedLabel : public QWidget {
+  public:
+
```
kranar, line 112: Vertical white space.

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,627 @@
...
+    explicit ElidedLabel(QWidget *parent = nullptr)
```
kranar, line 113: Use an inheriting constructor.

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,627 @@
...
+      update();
+    }
+
+    void set_line_height(int line_height) {
+      m_line_height = line_height;
+      update();
+    }
+
+    QSize sizeHint() const override {
+      return {width(), height()};
```
kranar, line 137: This looks unusual, can you confirm it's needed?

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,627 @@
...
+    }
+
+  protected:
+    void paintEvent(QPaintEvent *event) override {
+      auto painter = QPainter(this);
+      painter.setPen(m_text_color);
+      painter.setFont(font());
+      auto fontMetrics = painter.fontMetrics();
```
kranar, line 145: Why is a copy of this being made?

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,627 @@
...
+      auto y = m_line_height - fontMetrics.height();
+      auto textLayout = QTextLayout(m_text, painter.font());
```
kranar, line 147: Why is a ```QTextLayout``` constantly being constructed on every paint operation?

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,627 @@
...
+      auto opt = textLayout.textOption();
+      opt.setAlignment(m_alignment);
+      textLayout.setTextOption(opt);
+      textLayout.beginLayout();
+      auto line = textLayout.createLine();
+      while(line.isValid()) {
```
kranar, line 153: This appears like a for loop:

    for(auto line = foo(); line.isValid(); line = foo()) { ... }

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,627 @@
...
+  update_placeholder_text();
+}
+
+bool TextAreaBox::is_read_only() const {
+  return m_text_edit->isReadOnly();
+}
+
+void TextAreaBox::set_read_only(bool read_only) {
+  m_text_edit->setReadOnly(read_only);
+  m_scroll_box->get_vertical_scroll_bar().set_position(0);
```
kranar, line 257: Why does setting read only always reposition the scrollbar?

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,627 @@
...
+  m_placeholder_text = value;
+  update_placeholder_text();
+}
+
+bool TextAreaBox::is_read_only() const {
+  return m_text_edit->isReadOnly();
+}
+
+void TextAreaBox::set_read_only(bool read_only) {
+  m_text_edit->setReadOnly(read_only);
```
kranar, line 256: This should be a no-op when no change is made.

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,627 @@
...
+  }
+  update_text_alignment();
+  update_document_line_height();
+}
+
+void TextAreaBox::update_document_line_height() {
+  apply_block_formatting([=] (const auto& block) {
+    auto cursor = m_text_edit->textCursor();
+    cursor.setPosition(block.position());
+    m_text_edit->setTextCursor(cursor);
```
kranar, line 419: Why is ```setTextCursor``` called twice without any intermediate side effects on ```m_text_edit```?

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,627 @@
...
+    m_stacked_widget->adjustSize();
+    m_placeholder->show();
+  } else {
+    m_placeholder->hide();
+  }
+}
+
+void TextAreaBox::update_text_alignment() {
+  apply_block_formatting(
+    [=, alignment = m_text_edit_styles.m_alignment] (const auto& block) {
```
kranar, line 460: ```alignment``` is never used.

Applications/Spire/Source/Ui/TextAreaBox.cpp
```diff
@@ -0,0 +1,627 @@
...
+      m_scroll_box->get_vertical_scroll_bar().sizeHint().width() -
+      border_size.width() - padding_size.width());
+  } else {
+    m_text_edit->setFixedWidth(width() - border_size.width() -
+      padding_size.width());
+  }
+}
+
+void TextAreaBox::on_current(const QString& current) {
+  if(m_text_edit->toPlainText() != current) {
```
kranar, line 484: What does this check do? It's not cheap comparing two strings, by itself it's an ```O(N)``` and when you consider that typing happens character by character this makes typing an ```O(N^2)``` operation.

kranar (changes requested):
There's definitely a lot of good work here, but needs a lot more attention to details.

Applications/Spire/Source/Ui/TextAreaBox.cpp
spireken: Do you mean all whitespace between the method definitions?

kranar: Just line 57.

Applications/Spire/Source/Ui/TextAreaBox.cpp
spireken: The QPlaintTextEdit doesn't support alignments, so I used the QTextEdit so I could reuse setAlignment.

kranar (changes requested):
Looks better, there's a lot of stuff that is potentially uninitialized and the code isn't structured in a way that makes it clear that all fields are guaranteed to be initialized before being read. Can you guarantee this or in cases where no such guarantee exists, initialize fields to a default value?

PR #688, opened 2021-08-25
List item disabled state

shaobinspire:


spireken (approved):
Looks good.

Applications/Spire/Source/Ui/ListItem.cpp
```diff
@@ -49,6 +50,9 @@ bool ListItem::is_selected() const {
 }
 
 void ListItem::set_selected(bool is_selected) {
+  if(!isEnabled()) {
```
kranar, line 53: This violates both the API and the model. ```ListItem``` is a view class, it has absolutely no authority to override the model about whether the value it represents is selected or not.

Applications/Spire/Source/Ui/ListItem.cpp
```diff
@@ -1,5 +1,6 @@
 #include "Spire/Ui/ListItem.hpp"
 #include <QEvent>
+#include <QFocusEvent>
```
kranar, line 3: When including one header file ```A``` that is required to include header file ```B```, remove all includes of ```B``` to avoid implicit duplicate and redundant includes.

kranar (changes requested):

kranar:
The implementation has been revised to use the FocusObserver.

PR #693, opened 2021-08-31
The DropDownBoxAdaptor and the SideBox

shaobinspire:


spireken (changes requested):
Looks great, but are you sure you can use a template widget class: https://doc.qt.io/qt-5/moc.html#limitations

We don't use the Q_OBJECT macro but the base QWidget does.

shaobinspire:
> Looks great, but are you sure you can use a template widget class: https://doc.qt.io/qt-5/moc.html#limitations
> 
> We don't use the Q_OBJECT macro but the base QWidget does.

I think moc just can't handle the template class instead of its base class. There are some similar class templates in our code like DecimalBoxAdaptor, ScalarFilterPanel.

spireken (approved):

Applications/Spire/Include/Spire/Ui/DropDownBoxAdaptor.hpp
```diff
@@ -0,0 +1,149 @@
...
+      /** Returns the model. */
+      const std::shared_ptr<ValueModel<Type>>& get_model() const;
+
+      /** Returns <code>true</code> iff this DropDownBoxAdaptor is read-only. */
+      bool is_read_only() const;
+
+      /**
+       * Sets the read-only state.
+       * @param is_read_only True iff the DropDownBoxAdaptor should be
+       *                     read-only.
```
kranar, line 60: Fix indentation, the continuation goes below ```is_read_only```.

Applications/Spire/Include/Spire/Ui/DropDownBoxAdaptor.hpp
```diff
@@ -0,0 +1,149 @@
...
+        DropDownBox* m_drop_down_box;
+        boost::signals2::scoped_connection m_current_connection;
+
+        void populate_data();
+        void on_current(const Type& current);
+  };
+
+  template<typename T>
+  DropDownBoxAdaptor<T>::DropDownBoxAdaptor(QWidget* parent)
+    : DropDownBoxAdaptor(Type::NONE, parent) {}
```
kranar, line 83: What is ```Type::NONE```?

Applications/Spire/Include/Spire/Ui/SideBox.hpp
```diff
@@ -0,0 +1,18 @@
+#ifndef SPIRE_SIDE_BOX_HPP
+#define SPIRE_SIDE_BOX_HPP
+#include "Nexus/Definitions/Side.hpp"
+#include "Spire/Ui/DropDownBoxAdaptor.hpp"
+#include "Spire/Ui/Ui.hpp"
+
+namespace Spire {
+
+  template<>
+  void DropDownBoxAdaptor<Nexus::Side>::populate_data() {
```
kranar, line 10: This is effectively hard coding the model, have ```DropDownBoxAdaptor``` take a model in its constructor. It's not true that every ```DropDownBox``` for a given type always displays the same hardcoded values.

kranar (changes requested):
Good as an initial implementation but still has a lot of room for improvement. Look at simple ways of eliminating unneeded code and go over it in fine detail to remove or simplify the implementation.

Applications/Spire/Include/Spire/Ui/DropDownBoxAdaptor.hpp
```diff
@@ -0,0 +1,140 @@
...
+       * @param submission The submitted value.
+       */
+      using SubmitSignal = Signal<void (const Type& submission)>;
+
+      /**
+       * Constructs a DropDownBoxAdaptor using a LocalValueModel.
+       * @param list_model The model of values to display.
+       * @param parent The parent widget.
+       */
+      DropDownBoxAdaptor(std::shared_ptr<ListModel> list_model,
```
kranar, line 33: ```explicit```

Applications/Spire/Include/Spire/Ui/DropDownBoxAdaptor.hpp
```diff
@@ -0,0 +1,140 @@
...
+        QWidget* parent = nullptr);
+
+      /**
```
kranar, line 36: Fix indentation.

Applications/Spire/Include/Spire/Ui/DropDownBoxAdaptor.hpp
```diff
@@ -0,0 +1,140 @@
...
+        boost::signals2::scoped_connection m_current_connection;
+
+        void on_current(const boost::optional<int>& current);
+  };
+
+  template<typename T>
+  DropDownBoxAdaptor<T>::DropDownBoxAdaptor(
+    std::shared_ptr<ListModel> list_model, QWidget* parent)
+    : DropDownBoxAdaptor(std::move(list_model),
+        std::make_shared<LocalValueModel<optional<int>>>(), parent) {}
```
kranar, line 82: You must fully qualify all names residing outside of the current ```namespace```.

Applications/Spire/Include/Spire/Ui/DropDownBoxAdaptor.hpp
```diff
@@ -0,0 +1,140 @@
...
+       */
+      void set_read_only(bool is_read_only);
+
+      /** Connects a slot to the submit signal. */
+      boost::signals2::connection connect_submit_signal(
+        const typename SubmitSignal::slot_type& slot) const;
+
+      private:
+        mutable SubmitSignal m_submit_signal;
+        std::shared_ptr<ListModel> m_list_model;
```
kranar, line 69: This suggests that the ```ListModel```'s ownership is shared by the ```DropDownBoxAdaptor``` and the ```ListView```, why is this shared with both instead of having the ```ListView``` own it?

Applications/Spire/Include/Spire/Ui/DropDownBoxAdaptor.hpp
```diff
@@ -0,0 +1,140 @@
...
+        std::shared_ptr<ValueModel<boost::optional<int>>> m_current_model;
```
kranar, line 70: Same thing here, why is the ownership of the model being shared?

Applications/Spire/Include/Spire/Ui/DropDownBoxAdaptor.hpp
```diff
@@ -0,0 +1,140 @@
...
+        ListView* m_list_view;
```
kranar, line 71: The more state that ```DropDownBoxAdaptor``` maintains, the more ways for things to fail. Looking over this it seems like almost all of these fields can be eliminated and only increase the surface area for bugs.

Applications/Spire/Source/UiViewer/StandardUiProfiles.cpp
```diff
@@ -137,6 +140,26 @@ namespace {
     return box;
   }
 
+  template<typename B>
+  auto setup_drop_down_box_adaptor_profile(UiProfile& profile) {
+    using Type = std::decay_t<B>::Type;
```
kranar, line 145: What does this do?

kranar (changes requested):
Looks better but still needs a lot more work. You can remove probably 40%-50% of the code and in doing so reduce the risk of bugs and inconsistencies.

kranar:
This is definitely going in the wrong direction. As a general matter, it is taking a very big risk to implement or change an API without first proposing it and getting it approved. The API is the most important and highest risk aspect of a component. An implementation can always change, it can be made faster, it can have bugs fixed, it can be refactored, etc etc... but a broken API will be broken forever. An API can't really be changed easily, once an API is submitted it stays around for a loooong time and any issues with the API will accumulate into the future.

No API was proposed for this; the initial API was reasonable but had some minor issues that could be improved upon. The current API in this PR is inherently broken.

For the time being, work on another task. I'll find some time to work on this and when it's ready I'll let you know and you can use it to implement the remaining drop down boxes.

PR #697, opened 2021-09-03
Added the free function of make_label with a model.

shaobinspire:


Applications/Spire/Source/Ui/TextBox.cpp
```diff
@@ -489,7 +489,11 @@ void TextBox::on_style() {
 }
 
 TextBox* Spire::make_label(QString label, QWidget* parent) {
-  auto text_box = new TextBox(std::move(label), parent);
+  return make_label(std::make_shared<LocalTextModel>(std::move(label)), parent);
```
kranar, line 492: The ```TextBox``` has a special constructor for handling a constant values and this change inhibits the ```TextBox``` from special casing that scenario. Instead take the code that applies a "label" styling to a ```TextBox``` and factor that out into its own function, and have both ```make_label``` overloads use that function.

PR #698, opened 2021-09-08
Incorrect times on profile page

spiresheep:
This should solve the bug provided the duration values from the server are correct.

### DIFF
--- a/Applications/WebPortal/WebApp/library/source/pages/account_page/profile_page/profile_page.tsx
+++ b/Applications/WebPortal/WebApp/library/source/pages/account_page/profile_page/profile_page.tsx
@@ -176,7 +176,10 @@ export class ProfilePage extends React.Component<Properties, State> {
         const localizedDate = new Date(Date.UTC(
           serverDateTime.date.year,
           serverDateTime.date.month - 1,
-          serverDateTime.date.day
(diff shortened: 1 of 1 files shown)


Applications/WebPortal/WebApp/library/source/pages/account_page/profile_page/profile_page.tsx
```diff
@@ -166,6 +166,36 @@ export class ProfilePage extends React.Component<Properties, State> {
         return null;
       }
     })();
+    const lastLogin = (() => {
+      const serverDateTime = this.props.identity.lastLoginTime;
+      if(serverDateTime.equals(Beam.DateTime.NEG_INFIN) ||
```
kranar, line 171: Pull this out into its own function.

spiresheep: Just to be clear, you mean this:
 ```
          serverDateTime.equals(Beam.DateTime.NEG_INFIN) ||
          serverDateTime.equals(Beam.DateTime.POS_INFIN) ||
          serverDateTime.equals(Beam.DateTime.NOT_A_DATE_TIME)
```

PR #699, opened 2021-09-09
FocusObserver

shaobinspire:


Applications/Spire/Include/Spire/Ui/FocusObserver.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+        /**
+         * The state indicates a widget gets focus via keyboard or other
+         * non-pointing device.
+         */
+        FOCUS_VISIBLE = 0x04
+      };
+
+      /**
+       * Signals that the state has changed.
```
kranar, line 36: Document the parameter.

Applications/Spire/Include/Spire/Ui/FocusObserver.hpp
```diff
@@ -0,0 +1,87 @@
...
+        NONE = 0,
+
+        /** The state indicates a widget gets focus. */
+        FOCUS = 0x01,
+
+        /**
+         * The state indicates a widget or one of this widget's children
+         * gets focus.
+         */
+        FOCUS_IN = 0x02,
```
kranar, line 26: These enums model mutually exclusive states, but they have dependencies with one another meaning the definition has the risk of introducing inconsistencies. For example it is never possible to have ```FOCUS``` without ```FOCUS_IN```, hence ```FOCUS``` must include ```FOCUS_IN```, a similar property holds for ```FOCUS_VISIBLE``` something along the lines of:

   FOCUS_IN = 0x01,
   FOCUS = FOCUS_IN | 0x02,
   FOCUS_VISIBLE = FOCUS | 0x04

This way it's never possible to be in the ```FOCUS``` state without being in the ```FOCUS_IN``` state as well.

Clarify whether ```FOCUS_VISIBLE``` requires ```FOCUS``` or just ```FOCUS_IN```.

Applications/Spire/Source/Ui/FocusObserver.cpp
```diff
@@ -0,0 +1,129 @@
...
+      m_observer->m_state_signal(m_observer->m_state);
+    }
+  }
+};
+
+FocusObserver::FocusObserver(const QWidget& widget)
+    : m_widget(&widget),
+      m_state(State::NONE) {
+  if(m_widget->hasFocus()) {
+    m_state = State::FOCUS | State::FOCUS_IN;
```
kranar, line 94: Once you adopt the fix above, you won't need to do ```FOCUS | FOCUS_IN```, it will never be possible to have an inconsistent state like that.

Applications/Spire/Source/Ui/FocusObserver.cpp
```diff
@@ -0,0 +1,129 @@
...
+#include <QApplication>
+#include <QFocusEvent>
+
+using namespace boost::signals2;
+using namespace Spire;
+
+namespace {
+  QObject* find_focus_event_filter_from_parent(QWidget& widget) {
+    auto parent = &widget;
+    while (parent) {
```
kranar, line 11: This is too brittle, parents can change, windows can change.

Applications/Spire/Source/Ui/FocusObserver.cpp
```diff
@@ -0,0 +1,129 @@
...
+    : m_widget(&widget),
+      m_state(State::NONE) {
+  if(m_widget->hasFocus()) {
+    m_state = State::FOCUS | State::FOCUS_IN;
+  } else if(m_widget->isAncestorOf(QApplication::focusWidget())) {
+    m_state = State::FOCUS_IN;
+  }
+  m_old_state = m_state;
+  m_focus_event_filter = std::make_unique<FocusEventFilter>(*this);
+  QObject::connect(m_widget, &QObject::destroyed, [=] (QObject*) {
```
kranar, line 100: Remove line 100-102. It assumes that a ```FocusObserver``` outlives ```widget``` but that's not a requirement of the observer and in fact it's not a desirable property.

Applications/Spire/Source/Ui/FocusObserver.cpp
```diff
@@ -0,0 +1,129 @@
...
+    return nullptr;
+  }
+}
+
+struct FocusObserver::FocusEventFilter : QObject {
+  FocusObserver* m_observer;
+  Qt::FocusReason m_focus_reason;
+
+  FocusEventFilter(FocusObserver& observer)
+      : QObject(const_cast<QWidget*>(observer.m_widget)),
```
kranar, line 31: Remove the ```const_cast```.

Applications/Spire/Source/Ui/FocusObserver.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+FocusObserver::State Spire::operator |(FocusObserver::State left,
+    FocusObserver::State right) {
+  return static_cast<FocusObserver::State>(static_cast<std::uint8_t>(left) |
+    static_cast<std::uint8_t>(right));
+}
+
+FocusObserver::State Spire::operator &(FocusObserver::State left,
+    FocusObserver::State right) {
+  return static_cast<FocusObserver::State>(static_cast<std::uint8_t>(left) &
```
kranar, line 122: Replace ```std::uint8_t``` with ```std::underlying_type_t<FocusObserver::State>```.

kranar (changes requested):
Some good work here but could be improved quite a bit. One idea is to associate a ```QWidget``` with a ```std::shared_ptr<FocusObserver>``` and reuse a single object shared by multiple observers. Right now an observer is associated with the top level window which doesn't make much sense and looks like it would break if multiple widgets have observers belonging to the same window.

At any rate, there are good ideas but the implementation will need to be improved.

Applications/Spire/Include/Spire/Ui/FocusObserver.hpp
shaobinspire: I think `FOCUS_VISIBLE` should be independent of `FOCUS` and `FOCUS_IN`. `FOCUS_VISIBLE` is not only a focus state but also indicates the widget gets focus from a keyboard event, so `FOCUS_VISIBLE` could happen upon `FOCUS` or `FOCUS_IN`. When the widget itself gets focus by keyboard, the focus state will be `FOUCS |  FOCUS_VISIBLE`; When a child widget gets focus by keyboard, the state will be `FOCUS_IN | FOCUS_VISIBLE`.

kranar: You will need to be very careful about the dependencies between these states. Think through it carefully based on the specs and any details you may need from Jon and then revise the API accordingly.

Applications/Spire/Source/Ui/FocusObserver.cpp
shaobinspire: `observer.m_widget` is a `const` variable, but the copy constructor of `QObject` doesn't accept a `const` object.

Applications/Spire/Include/Spire/Ui/FocusObserver.hpp
shaobinspire: According to the definition of these states, focus_in indicates the widget or any of its descendants has focus, whereas focus indicates only the widget gets focus. The overlap of focus_in and focus is the widget itself gets focus. I think Focus_visible is not mutually exclusive to focus/focus_in because it represents the focus getting from the keyboard but it can’t tell whether this widget gets focus or one of its children gets focus.

I think the enums should be

NONE = 0,
FOCUS = 0x01,
FOCUS_IN = FOCUS | 0x02,
FOCUS_VISIBLE = 0x04

Applications/Spire/Source/Ui/FocusObserver.cpp
shaobinspire: The event filter doesn't need a parent, and it can outlive the widget, so it can be removed.

Applications/Spire/Include/Spire/Ui/FocusObserver.hpp
kranar: Okay, this still has issues, it says that FOCUS and FOCUS_VISIBLE are independent states. It also says that having FOCUS_IN requires having FOCUS.

Go over this very carefully again and then update your proposal.

Applications/Spire/Source/Ui/FocusObserver.cpp
kranar: ```QObject``` does not have a copy constructor. The fact that ```QObject```'s constructor is rejecting this should be an indication that the entire operation is incorrect.

Applications/Spire/Include/Spire/Ui/FocusObserver.hpp
shaobinspire: Updated the enums:

NONE = 0,
FOCUS_IN = 0x01,
FOCUS = FOCUS_IN | 0x02,
FOCUS_VISIBLE = FOCUS | 0x04

kranar: This says FOCUS_IN is an independent state.
FOCUS is a new state that depends on being in the FOCUS_IN state.
FOCUS_VISIBLE is a new state that requires being in the FOCUS state (which implies being in the FOCUS_IN state).

If this is correct then update the API accordingly.

Applications/Spire/Source/Ui/FocusObserver.cpp
```diff
@@ -0,0 +1,97 @@
...
+struct FocusObserver::FocusEventFilter : QObject {
+  FocusObserver* m_observer;
+  Qt::FocusReason m_focus_reason;
+
+  FocusEventFilter(FocusObserver& observer)
+      : QObject(),
+        m_observer(&observer),
+        m_focus_reason(Qt::MouseFocusReason) {
+    qApp->installEventFilter(this);
+    QObject::connect(qApp,
```
kranar, line 17: Remove the ```QObject::```.

Applications/Spire/Source/Ui/FocusObserver.cpp
```diff
@@ -0,0 +1,97 @@
...
+    }
+    auto state = m_observer->m_state;
+    if(m_observer->m_widget == now) {
+      m_observer->m_state = State::FOCUS;
+    } else if(m_observer->m_widget->isAncestorOf(now)) {
+      m_observer->m_state = State::FOCUS_IN;
+    } else {
+      m_observer->m_state = State::NONE;
+    }
+    if(m_observer->m_state != State::NONE) {
```
kranar, line 44: I believe the check should be ```state == State::FOCUS```. FOCUS_VISIBLE depends on FOCUS.

kranar (changes requested):

Applications/Spire/Source/Ui/FocusObserver.cpp
shaobinspire: I don't think so. For example, DurationBox contains multiple sub-components, when it gets focus by pressing Tab, the focus state is FOCUS_IN, then because the focus reason is Qt::TabFocusReason, the focus state should be updated FOCUS_VISIBLE. FOCUS_VISIBLE also implies being in the FOCUS_IN state

kranar: Give me an example where ```m_observer->m_widget != now``` and ```m_observer->m_state == State::FOCUS_VISIBLE```

shaobinspire: The FocusObserver monitors the focus state of the DurationBox. The DurationBox contains three DecimalBoxs whose focus proxies are three TextBoxs. When pressing Tab, the first TextBox (The TextBox is the focus proxy of the first DruationBox) in the DurationBox gets focus, so now is TextBox and m_observer->m_widget is DurationBox (m_observer->m_widget != now). Because the DurationBox gets focus from keyboard, its focus state should be `State::FOCUS_VISIBLE`.

shaobinspire: It seems `focus_visible` is mutually exclusive to `focus_in` and `focus`. `foucs_in` and `focus` indicates a component or one of its descendants gets focus, whereas the `focus_visible` represents the component getting focus by keyboard. I think it would be easy to understand these states to me if the definition of `focus_in` is updated to “any of the box’s descendants has focus” by removing “the box has focus”, so `focus` doesn’t include `focus_in` again.

kranar: The does not satisfy the definition of focus_visible given in the model. The model states that focus_visible is true when focus is true and either focus was received from a non-pointing device or focus was received from a component that itself had focus_visible.

kranar (changes requested):

Applications/Spire/Source/Ui/FocusObserver.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+FocusObserver::FocusObserver(const QWidget& widget) {
+  static auto widget_workers = std::unordered_map<const QWidget*,
+    std::weak_ptr<FocusEventFilter>>();
+  auto worker = widget_workers.find(&widget);
+  if(worker != widget_workers.end()) {
+    m_worker = worker->second.lock();
+  } else {
+    m_worker = std::shared_ptr<FocusEventFilter>(
+      new FocusEventFilter(widget), [] (auto* p) {
```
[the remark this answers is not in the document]
shaobinspire: Looks like the VS compiler doesn’t support the deduction.

```
error C2227: left of '->m_widget' must point to class/struct/union/generic type
FocusObserver.cpp(101,15): message : type is '_T1'
1>        with
1>        [
1>            _T1=std::nullptr_t
1>        ]
FocusObserver.cpp(106): message : see reference to function template instantiation 'auto Spire::FocusObserver::{ctor}::<lambda_1>::operator ()<std::nullptr_t>(_T1) const' being compiled
1>        with
1>        [
1>            _T1=std::nullptr_t
1>        ]
FocusObserver.cpp(105): message : see reference to variable template 'const bool conjunction_v<std::is_move_constructible<`Spire::FocusObserver::FocusObserver'::`7'::<lambda_1> >,std::_Can_call_function_object<`Spire::FocusObserver::FocusObserver'::`7'::<lambda_1> &,std::nullptr_t &,void> >' being compiled
```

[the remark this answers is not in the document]
kranar: This is an MSVC bug, at any rate the issue is not handling a ```nullptr```. You can continue to use ```auto*``` to workaround the MSVC issue but you will need to handle the case where ```p == nullptr```.

kranar (approved):

PR #702, opened 2021-09-09
Title bar updates

spireken:


Applications/Spire/Include/Spire/Ui/TitleBar.hpp
```diff
@@ -24,19 +23,19 @@ namespace Spire {
        */
       void set_icon(const QImage& icon);
 
-      /** Returns the title text QLabel. */
-      QLabel* get_title_label() const;
+      /** Returns the title text label. */
+      TextBox* get_title_label() const;
```
kranar, line 27: If this can't be null, then return a reference. Consider whether the state of the ```TextBox``` is part of the substate of the ```TitleBar``` as a whole. If it is part of the substate then you must return a ```const TextBox&```, if it isn't then you can keep it as a ```TextBox&```.

kranar (changes requested):

Applications/Spire/Source/Ui/TitleBar.cpp
```diff
@@ -45,65 +62,66 @@ namespace {
...
-  m_icon_button->setFixedSize(scale(26, 26));
-  m_icon_button->setFocusPolicy(Qt::NoFocus);
-  set_style(*m_icon_button, std::move(icon_button_style));
-  m_layout->insertWidget(0, m_icon_button);
+  delete m_window_button;
+  m_window_button = make_icon_button(icon, this);
+  m_window_button->setFixedSize(scale(26, 26));
+  m_window_button->setFocusPolicy(Qt::NoFocus);
+  auto icon_button_style = WINDOW_BUTTON_STYLE();
+  set_style(*m_window_button, std::move(icon_button_style));
```
shaobinspire, line 120: It can be replaced by set_style(*m_window_button, WINDOW_BUTTON_STYLE());

Applications/Spire/Include/Spire/Ui/TitleBar.hpp
spireken: Can you explain what you mean by substate? I would look at this from an API perspective, where I would return a const reference because nothing else should need to modify the TitleBar.

Applications/Spire/Include/Spire/Ui/TitleBar.hpp
```diff
@@ -24,19 +23,19 @@ namespace Spire {
...
 
     protected:
       void changeEvent(QEvent* event) override;
       bool eventFilter(QObject* watched, QEvent* event) override;
       void resizeEvent(QResizeEvent* event) override;
 
     private:
-      QHBoxLayout* m_layout;
+      QHBoxLayout* m_container_layout;
       QImage m_icon;
```
shaobinspire, line 36: I think it can be removed, because I see it is only used in constructor.

shaobinspire (changes requested):

Applications/Spire/Include/Spire/Ui/TitleBar.hpp
kranar: The ```TitleBar``` class has a set of state that makes up a title bar. Is the state of the ```TextBox``` returned by ```get_title_label``` a subset of the state of the ```TitleBar```? For example consider a ```shared_ptr<int>``` which is a shared pointer to an integer. The state of the integer is not a part of the state of the ```shared_ptr<int>```, and hence ```shared_ptr<int>::get() const``` returns an ```int*``` instead of an ```const int*```. Changing the ```int``` does not change the state of the ```shared_ptr<int>```, the states are non-overlapping. If the state of the ```TextBox``` is non-overlapping with the state of the ```TitleBar```, then return a ```TextBox&```, if there is overlap then return a ```const TextBox&```.

kranar (approved):

PR #704, opened 2021-09-13
Overlay panel updates

spireken:


spireken:
@shaobinspire There's a separate case to handle the activation of the OverlayPanel.

Applications/Spire/Source/Ui/OverlayPanel.cpp
```diff
@@ -97,12 +102,30 @@ void OverlayPanel::set_positioning(Positioning positioning) {
...
+}
+
 bool OverlayPanel::eventFilter(QObject* watched, QEvent* event) {
   if(watched == m_body) {
     if(event->type() == QEvent::MouseButtonPress) {
       auto mouse_event = static_cast<QMouseEvent*>(event);
       m_mouse_pressed_position = mouse_event->pos();
-    } else if(event->type() == QEvent::MouseMove) {
+    } else if(event->type() == QEvent::MouseMove && m_is_draggable &&
+        m_positioning != Positioning::PARENT) {
```
shaobinspire, line 128: m_positioning != Positioning::PARENT is used on line 131 again.

Applications/Spire/Source/Ui/OverlayPanel.cpp
```diff
@@ -112,6 +135,12 @@ bool OverlayPanel::eventFilter(QObject* watched, QEvent* event) {
   } else if(watched == parentWidget()->window()) {
     if(event->type() == QEvent::Move) {
       position();
+    } else if(event->type() == QEvent::WindowDeactivate &&
+        m_is_closed_on_blur && !isActiveWindow() && m_was_activated) {
```
shaobinspire, line 139: When does the situation of '!isActiveWindow() && m_was_activated' happen?

shaobinspire (changes requested):

Applications/Spire/Source/Ui/OverlayPanel.cpp
spireken: This event is when the parent widget's window is deactivated, so this covers the cases where:

1. This instance of the OverlayPanel was activated. In this case m_was_activated will be true, but since this is the active window it shouldn't close.
2. Some unrelated window (or a sibling) was activated, but since this instance of the OverlayPanel was never activated, then it shouldn't close.

There's typically no situation where the parent widget's window is deactivated and this instance of the OverlayPanel was already activated previously, but since it's shorter to include both statements in the same condition and they don't interfere with each other I left it that way. I say typically because we don't currently change m_is_closed_on_blur at runtime, but a requirement could come up in the future where that happens and this condition would still satisfy the spec, which could be an argument for keeping it this way.

shaobinspire: I think that m_was_activated and isActiveWindow() should be equivalent, so m_was_activated and !isActiveWindow() always are opposite.

spireken: The difference is that m_was_activated is never updated after it's set to true, but isActiveWindow() is conditional on the state of the window. m_was_activated is false initially, but as soon as the OverlayPanel is activated m_was_activated is set to true and it stays true for the lifetime of the OverlayPanel.

shaobinspire: I see. Thank you for your clarification.

Applications/Spire/Source/Ui/OverlayPanel.cpp
```diff
@@ -97,21 +102,44 @@ void OverlayPanel::set_positioning(Positioning positioning) {
...
+bool OverlayPanel::event(QEvent* event) {
+  if(m_is_closed_on_blur) {
+    if(event->type() == QEvent::WindowDeactivate) {
+      if(!is_ancestor(qApp->activeWindow())) {
+        close();
+      }
+    } else if(event->type() == QEvent::WindowActivate) {
+      m_was_activated = true;
+    }
+  } else if(event->type() == QEvent::ParentAboutToChange) {
```
kranar, line 114: Have you tested this branch at any point?

Applications/Spire/Source/Ui/OverlayPanel.cpp
```diff
@@ -97,21 +102,44 @@ void OverlayPanel::set_positioning(Positioning positioning) {
   m_positioning = positioning;
 }
 
+bool OverlayPanel::event(QEvent* event) {
+  if(m_is_closed_on_blur) {
+    if(event->type() == QEvent::WindowDeactivate) {
```
kranar, line 107: What part of the spec does this implement?

Applications/Spire/Source/Ui/OverlayPanel.cpp
```diff
@@ -130,6 +158,17 @@ void OverlayPanel::keyPressEvent(QKeyEvent* event) {
   QWidget::keyPressEvent(event);
 }
 
+bool OverlayPanel::is_ancestor(QWidget* widget) {
```
kranar, line 161: https://doc.qt.io/qt-5/qwidget.html#isAncestorOf

kranar (changes requested):

Applications/Spire/Source/Ui/OverlayPanel.cpp
spireken: I tried that originally but it only checks up until the window containing the given widget, while my implementation of it goes  up to the the highest parent regardless of what window it's in.

Applications/Spire/Source/Ui/OverlayPanel.cpp
spireken: This was the main update in this case for the concept of a 'FocusOut'. If a child window of the OverlayPanel is activated then it's only considered a blur but not a FocusOut, so the OverlayPanel should stay open.

kranar: Okay, use the recently merged ```FocusObserver``` to monitor the focus state of the ```OverlayPanel```. If you identify bugs in the ```FocusObserver```, report them.

Applications/Spire/Source/Ui/OverlayPanel.cpp
spireken: No. I'll add a feature into the UiViewer to test re-parenting of the OverlayPanel.

Applications/Spire/Include/Spire/Spire/Utility.hpp
```diff
@@ -5,9 +5,9 @@
 
 namespace Spire {
 
-  //! Deletes an object when control returns to the thread's event loop.
-  /*
-    \param object The object to schedule for deletion.
+  /**
+   * Deletes an object when control returns to the thread's event loop.
+   * @param object The object to schedule for deletion.
```
kranar, line 10: Fix indentation on line 11.

Applications/Spire/Include/Spire/Spire/Utility.hpp
```diff
@@ -17,15 +17,25 @@ namespace Spire {
...
+
+  /**
+   * Calculates a value from a second interval as a relative equivalent of a
+   * first interval, where the intervals are both closed.
+   * @param value The value to map.
+   * @param a The minimum value of the first interval.
+   * @param b The maximum value of the first interval.
+   * @param c The minimum value of the second interval.
+   * @param d The maximum value of the second interval.
+   * @returns The value mapped from the first interval to the second interval.
```
kranar, line 38: Fix indentation on line 39.

Applications/Spire/Include/Spire/Spire/Utility.hpp
```diff
@@ -35,33 +45,33 @@ namespace Spire {
...
+  /**
+   * Calculates a value from a second interval as a relative equivalent of a
+   * first interval, where the intervals are both closed. Prevents truncation
+   * of integers while mapping the value.
+   * @param value The value to map.
+   * @param a The minimum value of the first interval.
+   * @param b The maximum value of the first interval.
+   * @param c The minimum value of the second interval.
+   * @param d The maximum value of the second interval.
+   * @returns The value mapped from the first interval to the second interval.
```
kranar, line 57: See above.

Applications/Spire/Include/Spire/Spire/Utility.hpp
```diff
@@ -35,33 +45,33 @@ namespace Spire {
...
```
kranar, line 74: Same here.

Applications/Spire/Include/Spire/Ui/OverlayPanel.hpp
```diff
@@ -41,11 +42,21 @@ namespace Spire {
 
       /**
        * Sets whether the panel closes on blur event.
-       * @param is_closed_on_blur Sets the panel to be colsed on blur event iff
-       *                          is_closed_on_blur is true.
+       * @param is_closed_on_blur Sets the panel to be closed on blur event iff
```
kranar, line 45: Very nice.

Applications/Spire/Source/Ui/OverlayPanel.cpp
```diff
@@ -43,11 +44,15 @@ namespace {
...
+    : QWidget(parent,
+        Qt::Tool | Qt::FramelessWindowHint | Qt::NoDropShadowWindowHint),
       m_body(body),
       m_is_closed_on_blur(true),
-      m_positioning(Positioning::PARENT) {
+      m_is_draggable(true),
+      m_was_activated(false),
+      m_positioning(Positioning::PARENT),
+      m_focus_observer(*this),
+      m_parent_focus_observer(*parent) {
```
kranar, line 55: Don't dereference a value without a guarantee about its correctness.

Applications/Spire/Source/Ui/OverlayPanel.cpp
```diff
@@ -78,15 +88,15 @@ bool OverlayPanel::is_closed_on_blur() const {
 }
 
 void OverlayPanel::set_closed_on_blur(bool is_closed_on_blur) {
```
kranar, line 90: The spec has been updated to reflect that it's close on focus out rather than on blur.

kranar (changes requested):
This looks good overall.

Applications/Spire/Source/Ui/OverlayPanel.cpp
spireken: In that case, I think OverlayPanel, DropDownList, and FilterPanel should take a parent by reference in their constructors instead of as a pointer.

kranar: If they are all required to have a parent, then absolutely. Good call on that.

kranar (approved):
This was well done, substantial work with very few issues.

PR #706, opened 2021-09-14
Set focus on DropDownList

shaobinspire:
It relates to the task of "ListItem can’t get focus after ListView is updated"

### DIFF
--- a/Applications/Spire/Source/Ui/DropDownList.cpp
+++ b/Applications/Spire/Source/Ui/DropDownList.cpp
@@ -1,6 +1,7 @@
 #include "Spire/Ui/DropDownList.hpp"
 #include <QEvent>
 #include <QHBoxLayout>
+#include "Spire/Spire/ValueModel.hpp"
 #include "Spire/Ui/ArrayListModel.hpp"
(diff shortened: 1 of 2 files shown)


Applications/Spire/Source/Ui/DropDownList.cpp
```diff
@@ -33,6 +34,7 @@ DropDownList::DropDownList(ListView& list_view, QWidget* parent)
   m_panel = new OverlayPanel(this, parent);
   m_panel->set_closed_on_blur(true);
   on_panel_style();
+  setFocusProxy(m_list_view);
```
kranar, line 37: What's the reasoning for setting the focus proxy on line 37, where it splits up the work done on styling the panel?

shaobinspire: The purpose is not for styling. The OverlayPanel sets focus proxy to its body (DropDownList) and DropDownList sets focus proxy to the ListView. When setting focus on OverlayPanel, the ListView finally gets focus.

spireken (approved):

PR #711, opened 2021-09-20
ListValueModel

shaobinspire:


Applications/Spire/Source/Ui/ListValueModel.cpp
```diff
@@ -0,0 +1,61 @@
...
+    return QValidator::State::Invalid;
+  } else {
+    return QValidator::State::Acceptable;
+  }
+}
+
+const ListValueModel::Type& ListValueModel::get_current() const {
+  if(get_state() == QValidator::State::Acceptable) {
+    return m_source->at(m_index);
+  } else {
```
spireken, line 24: This else statement and the else statement in get_state() can be removed, because when the if statement is false then the body of the else statement will always be executed.

Applications/Spire/Source/Ui/ListValueModel.cpp
```diff
@@ -0,0 +1,61 @@
...
+connection ListValueModel::connect_current_signal(
+    const CurrentSignal::slot_type& slot) const {
+  return m_current_signal.connect(slot);
+}
+
+void ListValueModel::on_operation(const ListModel::Operation& operation) {
+  visit(operation,
+    [&] (const ListModel::AddOperation& operation) {
+      if(m_index >= operation.m_index && m_index < m_source->get_size()) {
+        m_current_signal(get_current());
```
spireken, line 42: What's the purpose of signaling the current value during these operations?

spireken (changes requested):

Applications/Spire/Source/Ui/ListValueModel.cpp
shaobinspire: From the description of the task, "As changes to the ListModel's value are made, the ListValueModel will reflect those changes and signal them."

spireken: I see. I guess I expected that the index would be updated if it was moved, but in this case it appears that it's fixed to one specific index relative to the view.

shaobinspire: The ListValueModel takes a ListModel and an index, and models the value in the ListModel at the specified index, so the index is fixed for the ListValueModel.

spireken (approved):

Applications/Spire/Include/Spire/Ui/ListValueModel.hpp
```diff
@@ -0,0 +1,44 @@
...
+      QValidator::State set_current(const Type& value) override;
+
+      boost::signals2::connection connect_current_signal(
+        const CurrentSignal::slot_type& slot) const override;
+
+    private:
+      mutable CurrentSignal m_current_signal;
+      std::shared_ptr<ListModel> m_source;
+      int m_index;
+      std::any m_empty_value;
```
kranar, line 37: Remove this.

Applications/Spire/Source/Ui/ListValueModel.cpp
```diff
@@ -0,0 +1,59 @@
+#include "Spire/Ui/ListValueModel.hpp"
+
+using namespace boost::signals2;
+using namespace Spire;
+
+ListValueModel::ListValueModel(std::shared_ptr<ListModel> source, int index)
+    : m_source(std::move(source)),
+      m_index(index) {
+  m_operation_connection = m_source->connect_operation_signal(
```
kranar, line 9: Use initializer lists to construct non-trivial types.

Applications/Spire/Source/Ui/ListValueModel.cpp
```diff
@@ -0,0 +1,59 @@
...
+      }
+    },
+    [&] (const ListModel::RemoveOperation& operation) {
+      if(m_index >= operation.m_index && m_index <= m_source->get_size()) {
+        m_current_signal(get_current());
+      }
+    },
+    [&] (const ListModel::MoveOperation& operation) {
+      auto bounds = std::minmax(operation.m_source, operation.m_destination);
+      if(m_index >= bounds.first && m_index <= bounds.second) {
```
kranar, line 50: After such a move, the ```m_index``` no longer corresponds to the index in the ```ListModel```. Just like with the row based view over a TableModel, the correspondence between rows is preserved across mutable operations.

Review the class RowViewTableModel to see how to implement something that preserves the index.

kranar: Okay keep this implementation for now, in the future you'll implement one that keeps the reference stable.

kranar (approved):

PR #714, opened 2021-09-23
Time in force box

shaobinspire:


Applications/Spire/Source/Ui/DropDownBox.cpp
```diff
@@ -155,6 +154,11 @@ bool DropDownBox::eventFilter(QObject* watched, QEvent* event) {
       match(*this, PopUp());
     } else if(event->type() == QEvent::Hide) {
       unmatch(*this, PopUp());
+    } else if(event->type() == QEvent::MouseButtonPress) {
+      auto& mouse_event = *static_cast<QMouseEvent*>(event);
+      if(rect().contains(mapFromGlobal(mouse_event.globalPos()))) {
+        m_drop_down_list->window()->setAttribute(Qt::WA_NoMouseReplay);
```
spireken, line 160: The documentation for Qt::WA_NoMouseReplay says:

"Used for pop-up widgets. Indicates that the most recent mouse press event should not be replayed when the pop-up widget closes. The flag is set by the widget's author and cleared by the Qt kernel every time the widget receives a new mouse event."

What does it mean for mouse events to be 'replayed'?

Applications/Spire/Source/Ui/OverlayPanel.cpp
```diff
@@ -131,18 +132,21 @@ void OverlayPanel::keyPressEvent(QKeyEvent* event) {
...
       parent_geometry.bottomLeft());
     auto screen_geometry = parentWidget()->screen()->availableGeometry();
-    auto panel_size = [&] {
-      if(layout()->contentsMargins() == DROP_SHADOW_MARGINS()) {
-        return size() - QSize(0, DROP_SHADOW_HEIGHT());
-      }
-      return size();
-    }();
+    auto panel_size = size();
+    if(auto win = windowHandle()) {
```
spireken, line 147: In which cases does the OverlayPanel not have a windowHandle?

spireken (commented):
Looks good.

Applications/Spire/Source/Ui/DropDownBox.cpp
shaobinspire: It means the mouse press event which is received by a popup widget before the widget is closed will be replayed on the next focused widget after the popup widget is closed. 

Take the DropDownbox as an example, the OverlayPanel in the DropDownBox is a popup widget. When the panel is opened, you click on the Button of the DropDownBox. The panel receives the mouse press event, and after the it is closed, the Button will get the “replayed” mouse event again if the attribute is not set. It will cause an issue: The OverlayPanel which should be closed will be reopened when clicking on the button as well as the panel is open. The reason is that the panel has been invisible when the “replayed” mouse event is received by the button and it will be shown again based on `DropDownBox::on_click()`.

Without Qt::WA_NoMouseReplay:
  ```a mouse press event ->  OverlayPanel  (before the panel is close) ->  Button (after the panel is close)```
With Qt::WA_NoMouseReplay: 
  ```a mouse press event ->  OverlayPanel```

Applications/Spire/Source/Ui/OverlayPanel.cpp
shaobinspire: I am not sure. Just in case it may be null.

shaobinspire:
The OverlayPanel has been update, so I close it to update DropDownBox.

shaobinspire:
Based on the updated OverlayPanel, two code snippets that @spireken mentioned has been removed because they are used to deal with the case of Qt::Popup which has been removed from the OverlayPanel.

spireken (approved):

kranar (approved):
Very nice.

PR #716, opened 2021-09-24
Hover observer

spireken:


kranar:
Definitely not happy with this implementation. I could be missing some major details or making a big oversight but this implementation looks very crude and hacky, performing multiple ```O(N)``` operations every 50 milliseconds seems like a recipe for disaster.

I will have to think it over very carefully. Like I said maybe I'm missing something here and the approach you took is sensible given some details I have not considered, but the implementation just looks overly complex, brute, and doesn't take advantage or leverage any very elegant properties that can be used to cut down on the complexity.

Give me some time to consider this and I will provide feedback.

PR #723, opened 2021-10-02
List view query

spireken:


Applications/Spire/Source/Ui/ListView.cpp
```diff
@@ -207,11 +207,19 @@ void ListView::append_query(const QString& query) {
   if(!m_items.empty()) {
     auto start = m_current_model->get_current().get_value_or(-1);
     auto i = (start + 1) % static_cast<int>(m_items.size());
+    auto is_repeated_query = m_query.count(m_query.at(0)) == m_query.count();
+    auto short_match = boost::optional<int>();
```
kranar, line 211: Remove ```boost::```. It's already being used via ```using namespace boost;```. Only ```std::``` goes unused because it results in a lot of name clashes and because ```std``` is fairly short.

Applications/Spire/Source/Ui/ListView.cpp
```diff
@@ -207,11 +207,19 @@ void ListView::append_query(const QString& query) {
...
-        set_current(i);
-        break;
+      if(m_items[i]->m_item->isEnabled()) {
+        if(displayTextAny(
+            m_list_model->at(i)).toLower().startsWith(m_query.toLower())) {
+          short_match = boost::none;
+          set_current(i);
+          break;
+        } else if(is_repeated_query && !short_match && displayTextAny(
+            m_list_model->at(i)).toLower().startsWith(m_query[0])) {
```
kranar, line 220: This is computing ```displayTextAny(...)``` twice instead of reusing the calculation that must have been performed on line 214. Reuse as much of this as possible since these are not trivial operations.

kranar (changes requested):
Looks good overall.

kranar (approved):
This is the one.

PR #724, opened 2021-10-02
Duration box submit

spireken:


kranar (approved):
Kept it to the point, that's good to see.

PR #726, opened 2021-10-09
Calendar day label

spireken:
The previous CalendarDatePicker implementation used CalendarDayLabels that took a model so that all I had to do was update the models for each label to update the calendar, but this created an issue where mouse pressing on a date that changes the displayed month would trigger another date change in the new month when the mouse was released. The updated implementation will just destroy and insert new CalendarDayLabels into the calendar ListView so this won't be an issue.

This PR adds the weekday header labels and adds the CalendarDayLabel for review.

### DIFF
--- a/Applications/Spire/Source/Ui/CalendarDatePicker.cpp
+++ b/Applications/Spire/Source/Ui/CalendarDatePicker.cpp
@@ -16,12 +16,12 @@ using namespace Spire::Styles;
 
 namespace {
   auto make_header_label(QString text, QWidget* parent) {
-    auto label = make_label(std::move(text), parent);
-    label->setFixedSize(scale(24, 24));
-    auto style = get_style(*label);
     auto font = QFont("Roboto");
     font.setWeight(60);
     font.setPixelSize(scale_width(12));
+    auto label = make_label(std::move(text), parent);
+    label->setFixedSize(scale(24, 24));
+    auto style = get_style(*label);
     style.get(Disabled() && ReadOnly()).
       set(Font(font)).
       set(TextAlign(Qt::AlignCenter)).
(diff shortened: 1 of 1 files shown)


Applications/Spire/Source/Ui/CalendarDatePicker.cpp
```diff
@@ -14,6 +14,38 @@ using namespace boost::signals2;
 using namespace Spire;
 using namespace Spire::Styles;
 
+namespace {
+  auto make_header_label(QString text, QWidget* parent) {
+    auto label = make_label(std::move(text), parent);
+    label->setFixedSize(scale(24, 24));
+    auto style = get_style(*label);
```
kranar, line 21: Declare at point of use. This jumps from ```style``` to ```font``` and then back to ```style```. Keep things local. Same above, you could do ```font``` then ```label``` then ```style```.

Applications/Spire/Source/Ui/CalendarDatePicker.cpp
```diff
@@ -14,6 +14,38 @@ using namespace boost::signals2;
...
+    header->setFixedSize(scale(168, 26));
+    auto layout = new QHBoxLayout(header);
+    layout->setContentsMargins({});
+    layout->setSpacing(0);
+    auto locale = QLocale();
+    layout->addWidget(make_header_label(locale.dayName(7).at(0), header));
+    for(auto i = 1; i < 7; ++i) {
+      layout->addWidget(make_header_label(locale.dayName(i).at(0), header));
+    }
+    return header;
```
kranar, line 45: This is good, everything is local here.

Applications/Spire/Source/Ui/CalendarDatePicker.cpp
```diff
@@ -101,6 +133,49 @@ class MonthSpinner : public QWidget {
     }
 };
 
+class CalendarDayLabel : public QWidget {
+  public:
+    CalendarDayLabel(
+        date day, date::month_type month, QWidget* parent = nullptr)
+        : QWidget(parent) {
+      setFixedSize(scale(24, 24));
+      auto layout = new QHBoxLayout(this);
```
kranar, line 142: Even here, jump from ```layout``` to ```label``` back to ```layout```, but ```layout``` can be worked with  after line 169 without any dependencies/issues.

PR #727, opened 2021-10-13
Calendar list model

spireken:


spireken:
There will be one final PR after this to tie everything together.

Applications/Spire/Source/Ui/CalendarDatePicker.cpp
```diff
@@ -46,6 +50,53 @@ namespace {
...
+    }
+
+  private:
+    std::shared_ptr<DateModel> m_model;
+    scoped_connection m_current_connection;
+    std::array<std::any, DAY_COUNT> m_dates;
+    ListModelTransactionLog m_transaction;
+
+    void on_current(date current) {
+      auto day = date(current.year(), current.month(), 1);
```
kranar, line 88: Think about the meeting we had two weeks ago and incorporate those concepts into the implementation.

Applications/Spire/Source/Ui/CalendarDatePicker.cpp
```diff
@@ -46,6 +50,82 @@ namespace {
...
+          day += days(-day.day_of_week());
+        }
+        for(auto i = 0; i < DAY_COUNT; ++i) {
+          m_dates[i] = day;
+          day += days(1);
+        }
+      }
+    }
+
+    const Dates& get_dates() {
```
shaobinspire, line 77: The method should be const.

Applications/Spire/Source/Ui/CalendarDatePicker.cpp
```diff
@@ -164,15 +250,30 @@ class CalendarDayLabel : public QWidget {
...
+    }
+
+  private:
+    std::shared_ptr<ListValueModel> m_model;
+    scoped_connection m_current_connection;
+    std::shared_ptr<DateModel> m_month_model;
+    TextBox* m_label;
+
+    void on_current(date day) {
+      m_label->get_model()->set_current(QString("%12").arg(day.day()));
```
shaobinspire, line 266: Why is the place marker "%12"?

Applications/Spire/Source/Ui/CalendarDatePicker.cpp
```diff
@@ -198,6 +299,33 @@ CalendarDatePicker::CalendarDatePicker(
...
+    set(Overflow(Overflow::WRAP));
+  calendar_style.get(Any() >> is_a<ListItem>()).
+    set(border_size(0)).
+    set(padding(0));
+  calendar_style.get(Any() >> (is_a<ListItem>() && Hover())).
+    set(BackgroundColor(QColor::fromRgb(0xFFFFFF)));
+  calendar_style.get(
+      Any() >> (is_a<ListItem>() && Selected()) >> is_a<CalendarDayLabel>()).
+    set(BackgroundColor(QColor::fromRgb(0x4B23A0))).
+    set(border(0, QColor::fromRgb(0, 0, 0, 0))).
```
shaobinspire, line 325: Replace QColor::fromRgb(0,0,0,0) with QColor(Qt::transparent).

Applications/Spire/Source/Ui/CalendarDatePicker.cpp
```diff
@@ -198,6 +299,33 @@ CalendarDatePicker::CalendarDatePicker(
...
+  auto calendar_style = StyleSheet();
+  calendar_style.get(Any()).
+    set(Qt::Horizontal).
+    set(EdgeNavigation(EdgeNavigation::CONTAIN)).
+    set(Overflow(Overflow::WRAP));
+  calendar_style.get(Any() >> is_a<ListItem>()).
+    set(border_size(0)).
+    set(padding(0));
+  calendar_style.get(Any() >> (is_a<ListItem>() && Hover())).
+    set(BackgroundColor(QColor::fromRgb(0xFFFFFF)));
```
shaobinspire, line 321: Replace QColor::fromRgb(0xFFFFFF) with QColor(0xFFFFFF). I think the use of fromRgb will be discouraged.

Applications/Spire/Source/Ui/CalendarDatePicker.cpp
spireken: It looks like QString will match any placeholder up to the number of arg(...) calls, so even with this typo it still functions correctly.

Applications/Spire/Source/Ui/CalendarDatePicker.cpp
spireken: I used Qt::black for black because there's ambiguity between the QColor(const char *name) and QColor(QRgb) constructors when passing 0x000000.

Applications/Spire/Source/Ui/CalendarDatePicker.cpp
shaobinspire: Yeah, it doesn't affect the function, but I think it would be better to assign the place marker starting from 1.

Applications/Spire/Source/Ui/CalendarDatePicker.cpp
```diff
@@ -46,6 +50,82 @@ namespace {
...
+        const OperationSignal::slot_type& slot) const override {
+      return m_transaction.connect_operation_signal(slot);
+    }
+
+  private:
+    std::shared_ptr<DateModel> m_model;
+    scoped_connection m_current_connection;
+    CalendarModel m_calendar_model;
+    std::array<std::any, CalendarModel::DAY_COUNT> m_dates;
+    ListModelTransactionLog m_transaction;
```
kranar, line 118: Why is this transaction log used?

kranar: Okay that's good then.

spireken: Well, now I'm confused. As a private implementation of the CalendarDatePicker I know an update in the CalendarListModel won't result in additional transactions that cause reentrancy issues, but do I address reentrancy here or trust that the CalendarDatePicker won't abuse its own models?

I'm leaning back towards keeping the transaction log on the chance that the CalendarDatePicker is updated and these assumptions on safety break.

spireken: I believe the update is considered a transaction.

kranar: Excellent, that makes a lot more sense now. You want a transaction to update all of the dates in one shot without interruption. Look over the TransactionLog's API and unit tests to get a sense of how it works and if you have questions, ask them on Asana.

Applications/Spire/Source/Ui/CalendarDatePicker.cpp
```diff
@@ -46,6 +50,86 @@ namespace {
   }
 }
 
+class CalendarModel {
```
kranar, line 53: What does this class do?

PR #728, opened 2021-10-14
Added a container to the InfoTip with margins that offsets the body b…

spireken:
I made some temporary styling changes in the demo so that the fix was obvious. The default demo uses a label with a transparent background, so the issue wasn't apparent when the InfoTip was originally implemented.


shaobinspire (approved):
Looks good.

PR #735, opened 2021-10-21
The ListView crashes with an exception when its current is the last item and one of the items is removed.

shaobinspire:


kranar (approved):
Well done.

PR #738, opened 2021-11-04
ListValueModel

shaobinspire:
Reimplemented the ListValueModel to update the index when the source ListView is operated.


kranar (approved):
Flawless, well done.

PR #739, opened 2021-11-09
Time in force filter panel

shaobinspire:


Applications/Spire/Source/Ui/TimeInForceFilterPanel.cpp
```diff
@@ -0,0 +1,49 @@
...
+    std::make_shared<TimeInForceListModel>(std::make_shared<ArrayListModel>()),
+      parent);
+}
+
+TimeInForceFilterPanel* Spire::make_time_in_force_filter_panel(
+    std::shared_ptr<TimeInForceListModel> selected_model, QWidget& parent) {
+  auto model = std::make_shared<ArrayTableModel>();
+  auto& values = populate();
+  for(auto& value : values) {
+    model->push({TimeInForce(value.first), false});
```
kranar, line 39: Avoid doing the same work on every function invocation.

shaobinspire: The second column of the TableModel is variable for the different instances of the TimeInForceFilterPanel, so it is impossible to create a static pointer of the TableModel for different instances of TimeInForceFilterPanel. Besides, since the copy constructor of the TableModel is deleted, the static object of the TableModel can’t be copied for the different TimeInForceFilterPanels. Hence, I have to push data to the TableModel for a new TimeInForceFilterPanel instance.

kranar (approved):
Very good.

PR #744, opened 2021-11-15
Combo box

spireken:
The ComboBox API wasn't formally presented but a barebones implementation was added so the QueryModel could be declared within the ComboBox class.


shaobinspire (approved):
Looks good.

PR #751, opened 2021-12-22
ComboBox current

shaobinspire:


Applications/Spire/Source/Ui/CustomQtVariants.cpp
```diff
@@ -325,8 +325,7 @@ QString CustomVariantItemDelegate::displayText(const QVariant& value,
     }
     return QString::fromStdString(region.GetName());
   } else if(value.canConvert<Security>()) {
-    return QString::fromStdString(ToWildCardString(value.value<Security>(),
-      GetDefaultMarketDatabase(), GetDefaultCountryDatabase()));
+    return QString::fromStdString(ToString(value.value<Security>()));
```
kranar, line 328: This can not change.

Applications/Spire/Source/Ui/SecurityBox.cpp
```diff
@@ -54,6 +54,10 @@ SecurityBox::SecurityBox(std::shared_ptr<ComboBox::QueryModel> query_model,
       m_current(std::move(current)) {
   auto combo_box_current = make_transform_value_model(m_current,
     [] (const Security& current) {
+      if(current.GetMarket().IsEmpty() &&
```
kranar, line 57: What is the intention here?

shaobinspire: When displayTextAny in ComboBox::on_current converts a Security to a string, a “.” will be displayed following Security.Symbol, even though Security.Market is empty and CountryCode is None, so in the previous commit, I replaced ToWildCardString with ToString to display Security within the displayTextAny function in order to prevent “.” from appearing. Since I can’t change the implementation of displayTextAny, I have to convert Security to the Security.Symbol string before passing it to the ComboBox when Market is empty and CountryCode is None, and the input box of the ComboBox will not display the extra “.”.

kranar: Use ```displayTextAny``` to convert a Security to a ```QString```.

shaobinspire: I am using `displayTextAny` to convert a Security to a `QString`. In order to show this problem more intuitively, I removed this code and uploaded the demo to demos/combo_box_current/UiViewer-SecurityBox.exe. 

1. Open the SecurityBox;
2.  Type "a" to the current input box in the right panel;
3. The ComboBox will show "a.";
4. Keep on typing "b" to the current input box;
5. The ComboBox will show "ab."
6. Keep on typing "." to the current input box;
7. The ComboBox will show "ab.."

The extra "." is shown after the text input by the user. The current implementation of `displayTextAny` adds the "." when it converts the `Security` to the `QString`. What I did was to try to resolve the issue.

kranar: There's no issue with that behavior. Go back to using ```displayText``` to convert a value to a ```QString```.

shaobinspire: I am confused about it. Do you mean I should implement the `displayText' to convert a `Security` to a `QString`? There is a free function `ToString` in the Security which can convert a `Security` to a `std::string`.

kranar: I mixed up ```displayText``` with ```displayTextAny```. You can use ```displayTextAny``` to convert a ```Security``` to a ```QString```.

shaobinspire: I think that behavior is an issue. Sebastian proposed the task to clear the ComboBox's input box when a submission was added to filtered in the OpenFilterPanel. For the SecurityBox, there is no way to clear it because "." will be always shown in the SecurityBox when setting SecurityBox.current to be empty.

kranar: What does it mean to be empty? How can a SecurityBox.current be empty?

shaobinspire: Sorry, the `SecurityBox.current` can't be empty. I mean to make the input box of the `SecurityBox` to be empty. In order to achieve it, I will set `SecurityBox.current` to a Security with a empty `Security.Symbol` and the None `CountryCode`. I expected that `displayTextAny` could convert the empty `Security` to a empty `QString` rather than " .".

kranar: Okay I think I understand, try setting ```SecurityBox.current``` to ```Security()```

Applications/Spire/Source/Ui/ComboBox.cpp
```diff
@@ -238,6 +241,13 @@ void ComboBox::submit(const QString& query, bool is_passive) {
   m_submit_signal(value);
 }
 
+void ComboBox::on_current(const std::any& current) {
+  auto text = displayTextAny(current);
+  if(text != m_input_box->get_current()->get()) {
```
kranar, line 246: Do not use text to compare for equality, a value can have multiple text representations, for example "5", "5.0", "5.000" all represent the same value but have different text representations. You will need to test for true equality.

PR #752, opened 2022-01-12
Box painted style

spireken:


Applications/Spire/Source/Ui/Box.cpp
```diff
@@ -112,174 +321,285 @@ bool Box::event(QEvent* event) {
...
+    std::min(static_cast<double>(width()) /
+      (radius.m_bottom_left + radius.m_bottom_right),
+      static_cast<double>(height()) /
+      (radius.m_bottom_left + radius.m_top_left)))));
+}
+
+Box::BorderRadius Box::reduce_radius() const {
+  auto radius = m_style.m_border_radius;
+  auto factor = radius_reduction_factor(radius);
+  while(factor < 1.0f) {
```
shaobinspire, line 433: I don't understand why the while loop is needed here. Sometimes, it may cause a infinite loop.

Applications/Spire/Source/Ui/Box.cpp
```diff
@@ -1,12 +1,203 @@
...
+    auto end = [&] {
+      if(next_border_width <= 0) {
+        return QPointF(widget_width, radius);
+      }
+      return get_curve_position({QPointF(widget_width - radius, 0),
+        QPointF(widget_width, 0), QPointF(widget_width, radius)},
+        get_transition_point(border_width, next_border_width));
+    }();
+    return {QPointF(widget_width - radius, 0),
+      QPointF(widget_width - (radius / 2), 0), end};
```
shaobinspire, line 114: radius should be cast to `double`

shaobinspire (changes requested):

Applications/Spire/Include/Spire/Ui/Box.hpp
```diff
@@ -158,17 +158,63 @@ namespace Styles {
...
+      struct CornerProperty {
+        T m_top_left;
+        T m_top_right;
+        T m_bottom_right;
+        T m_bottom_left;
+
+        bool are_equal() const;
+      };
+      using Padding = SideProperty<int>;
+      using BorderWidth = SideProperty<int>;
```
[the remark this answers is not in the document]
spireken: It's not necessary because we don't support negative widths, but I use them so it's consistent with the types that the style system uses.

kranar:
On the one hand it's clear that a lot of effort and work went into this, but on the other hand the implementation does not look correct, it appears to be significantly more complex and heavy than what was anticipated. There are many issues with how the border is rendered and it does not handle many scenarios properly, for example given a square Box and a border radius half the width of the Box, the border must be a perfect circle.

That said, it's certainly possible that I underestimated the challenge involved in this task so I will do a review not only of this work but also some prototyping of my own to see if there are challenges I did not anticipate. The expectation for this task was not to be nearly as complicated and my understanding is that this is supposed to be implemented in a fairly straightforward manner using nothing more than circles and ellipsis. Draw a left border, draw a top border, draw an arc connecting the left to the top, do the same thing for the top/right, right/bottom, bottom/left.

But as I said if there's something involved here that requires this complexity I'll look into it. Until then I will close this branch.

PR #753, opened 2022-01-14
Responsive label

spireken:


Applications/Spire/Source/Ui/ResponsiveLabel.cpp
```diff
@@ -0,0 +1,230 @@
...
+  if(m_text_box->is_text_elided() && m_current_mapped_index != 0) {
+    adjustSize();
+    update_display_text();
+  }
+}
+
+void ResponsiveLabel::sort_mapped_labels() {
+  std::sort(m_mapped_labels.begin(), m_mapped_labels.end(),
+    [=] (const auto& first, const auto& second) {
+      return first.m_pixel_width < second.m_pixel_width;
```
shaobinspire, line 104: When sorting the mapped_labels, the index isn't considered. According to the description of  the ResponsiveLabel, the value with the lowest list index among the same length values is displayed.

Applications/Spire/Source/Ui/ResponsiveLabel.cpp
```diff
@@ -0,0 +1,230 @@
...
+      }
+      return std::prev(mapped_label)->m_pixel_width;
+    }();
+    auto upper = [&] {
+      if(std::next(mapped_label) == m_mapped_labels.end()) {
+        return mapped_label->m_pixel_width;
+      }
+      return std::next(mapped_label)->m_pixel_width;
+    }();
+    if(mapped_label->m_pixel_width <= lower ||
```
shaobinspire, line 206: When `m_pixel_width == lower` or `m_pixel_width == upper`, do we need to sort labels?

Applications/Spire/Source/Ui/ResponsiveLabel.cpp
```diff
@@ -0,0 +1,230 @@
...
+        upper <= mapped_label->m_pixel_width) {
+      sort_mapped_labels();
+    }
+  }
+  set_current(m_mapped_labels.size() - 1);
```
shaobinspire, line 211: Cast `m_mapped_labels.size()` to int

Applications/Spire/Source/Ui/ResponsiveLabel.cpp
```diff
@@ -0,0 +1,230 @@
...
+  set_current(std::distance(m_mapped_labels.begin(), std::prev(mapped_label)));
+}
+
+void ResponsiveLabel::on_label_added(int index) {
+  m_mapped_labels.push_back({index, get_pixel_width(m_labels->get(index))});
+  if(m_mapped_labels.size() > 1 && m_mapped_labels.back().m_pixel_width <
+      m_mapped_labels.at(m_mapped_labels.size() - 2).m_pixel_width) {
+    sort_mapped_labels();
+  }
+  set_current(m_mapped_labels.size() - 1);
```
shaobinspire, line 162: Cast `m_mapped_labels.size()` to int

shaobinspire (changes requested):

Applications/Spire/Source/Ui/ResponsiveLabel.cpp
spireken: I believe we do because of the case where two equal length labels need to be sorted by list index.

The case where they would be equal would be something like a list: [CCC, AA, BB, DD]. Assuming each single character has the same pixel width, the mapped labels would be sorted as [AA, BB, DD, CCC]. If CCC was then updated to be 'CC', it would become equal in pixel width to the previous mapped label, DD, and the mapped labels would then need to be sorted to [CC, AA, BB, DD].

Applications/Spire/Source/Ui/ResponsiveLabel.cpp
```diff
@@ -101,7 +97,10 @@ void ResponsiveLabel::set_current(const optional<int> mapped_label_index) {
 void ResponsiveLabel::sort_mapped_labels() {
   std::sort(m_mapped_labels.begin(), m_mapped_labels.end(),
     [=] (const auto& first, const auto& second) {
-      return first.m_pixel_width < second.m_pixel_width;
+      if(first.m_pixel_width != second.m_pixel_width) {
+        return first.m_pixel_width < second.m_pixel_width;
+      }
+      return first.m_list_index < second.m_list_index;
```
shaobinspire, line 103: I think `<` should be `>`, because the list index should be sorted in descending order and the lower list index should be placed in the back of the `m_mapped_labels`.

shaobinspire (changes requested):

Applications/Spire/Source/Ui/ResponsiveLabel.cpp
spireken: They're sorted in ascending order, so if indices 2, 7, and 5 are the same pixel width, then they're sorted as 2, 5, 7, and the item at index 2 would be displayed.

shaobinspire (approved):

Applications/Spire/Include/Spire/Ui/ResponsiveLabel.hpp
```diff
@@ -0,0 +1,68 @@
...
+      QFont m_text_box_font;
+      std::vector<MappedLabel> m_mapped_labels;
+      boost::optional<int> m_current_mapped_index;
+
+      int get_current_label_pixel_width() const;
+      int get_next_label_pixel_width() const;
+      int get_pixel_width(const QString& text) const;
+      bool is_outside_current_bounds(int width) const;
+      void reset_mapped_labels();
+      void set_current(const boost::optional<int> mapped_index);
```
kranar, line 56: Pass a mutable value or pass by const reference.

Applications/Spire/Source/Ui/ResponsiveLabel.cpp
```diff
@@ -0,0 +1,229 @@
...
+  if(m_text_box->is_text_elided() && m_current_mapped_index != 0) {
+    adjustSize();
+    update_display_text();
+  }
+}
+
+void ResponsiveLabel::sort_mapped_labels() {
+  std::sort(m_mapped_labels.begin(), m_mapped_labels.end(),
+    [=] (const auto& first, const auto& second) {
+      if(first.m_pixel_width != second.m_pixel_width) {
```
kranar, line 100: return std::tie(first.m_pixel_width, -first.m_list_index) <
      std::tie(second.m_pixel_width, -second.m_list_index);

Applications/Spire/Source/Ui/ResponsiveLabel.cpp
```diff
@@ -0,0 +1,229 @@
...
+}
+
+void ResponsiveLabel::update_current_font() {
+  auto updated_font = QFont();
+  auto updated_font_size = m_text_box_font.pixelSize();
+  auto& stylist = find_stylist(*m_text_box);
+  for(auto& property : stylist.get_computed_block()) {
+    property.visit(
+      [&] (const Font& font) {
+        stylist.evaluate(font, [&] (const auto& font) {
```
kranar, line 114: Evaluations are not local, so you will need to capture by value.

Applications/Spire/Source/Ui/ResponsiveLabel.cpp
```diff
@@ -0,0 +1,229 @@
...
+          updated_font = font;
+        });
+      },
+      [&] (const FontSize& size) {
+        stylist.evaluate(size, [&] (auto size) {
```
kranar, line 119: See above.

Applications/Spire/Source/Ui/TextBox.cpp
```diff
@@ -342,6 +342,10 @@ void TextBox::set_read_only(bool read_only) {
   update_placeholder_text();
 }
 
+bool TextBox::is_text_elided() const {
```
kranar, line 345: Find a way to remove this and have the ```ReponsiveLabel``` determine if the text fits.

kranar (changes requested):
Looks pretty good overall. One major issue is that the default size/size hint of this component is unspecified, so it will not work well when laid out based on its intrinsic size, it will end up using the size of its layout. which in turn will end up using the size of the ```TextBox```, but the size of the ```TextBox``` is not the intrinsic size of the ```ResponsiveLabel```.

kranar:
This is very good.

PR #754, opened 2022-01-14
Fix SecurityBox

shaobinspire:
Added the functionality that the SecurityBox can query by the security symbol and the security name.

### DIFF
--- a/Applications/Spire/Source/Ui/SecurityBox.cpp
+++ b/Applications/Spire/Source/Ui/SecurityBox.cpp
@@ -33,7 +33,7 @@ struct SecurityBox::SecurityQueryModel : ComboBox::QueryModel {
           }
         }();
         auto result = std::vector<std::any>();
-        std::unordered_set<Security> securities;
+        auto securities = std::unordered_set<Security>();
         for(auto& value : matches) {
           auto& security = std::any_cast<SecurityInfo&>(value).m_security;
           if(!securities.contains(security)) {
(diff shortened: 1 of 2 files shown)


Applications/Spire/Source/Ui/SecurityBox.cpp
```diff
@@ -33,8 +33,13 @@ struct SecurityBox::SecurityQueryModel : ComboBox::QueryModel {
           }
         }();
         auto result = std::vector<std::any>();
+        std::unordered_set<Security> securities;
```
spireken, line 36: Initialize using auto.

spireken (changes requested):

spireken (approved):

Applications/Spire/Source/Ui/SecurityBox.cpp
```diff
@@ -33,8 +33,13 @@ struct SecurityBox::SecurityQueryModel : ComboBox::QueryModel {
           }
         }();
         auto result = std::vector<std::any>();
+        auto securities = std::unordered_set<Security>();
         for(auto& value : matches) {
-          result.push_back(std::any_cast<SecurityInfo&>(value).m_security);
+          auto& security = std::any_cast<SecurityInfo&>(value).m_security;
+          if(!securities.contains(security)) {
```
kranar, line 39: This does two lookups when you can perform this with a single lookup:

https://en.cppreference.com/w/cpp/container/unordered_set/insert

    if(securities.insert(security).second) {
      result.push_back(security);
    }

kranar (changes requested):
Looks good, short and sweet.

PR #756, opened 2022-01-18
Open filter panel

shaobinspire:


Applications/Spire/Source/Ui/ComboBox.cpp
```diff
@@ -354,6 +355,7 @@ void ComboBox::on_drop_down_submit(const std::any& submission) {
   }
   m_submission = submission;
   m_submission_text = text;
+  QApplication::setActiveWindow(window());
```
spireken, line 358: The [documentation](https://doc.qt.io/qt-5/qapplication.html#setActiveWindow) for this is confusing. It says it doesn't set the keyboard focus on the window, but then below that says it sets the focusWidget (which will receive key input). Is this equivalent to QWidget::activateWindow, or does it actually activate the window without setting keyboard focus?

Applications/Spire/Include/Spire/Ui/OpenFilterPanel.hpp
```diff
@@ -0,0 +1,153 @@
...
+      enum class FilterMode {
+
+        /** The filtered values are included. */
+        INCLUDE,
+
+        /** The filtered values are excluded. */
+        EXCLUDE
+      };
+
+      /** The OpenInputBox using to get input inside the OpenFilterPanel. */
```
spireken, line 66: I think this should say 'used to get the input', or some variation of that.

Applications/Spire/Include/Spire/Ui/OpenFilterPanel.hpp
```diff
@@ -0,0 +1,153 @@
...
+       * @param submit_handler The handler used to handle the SubmitSignal.
+       * @return The InputBoxBuilder that shall take input.
+       */
+      using InputBoxBuilder = std::function<
+        OpenInputBox (const std::shared_ptr<QueryModel>& query_model,
+          const std::shared_ptr<AnyListModel>& filtered_model,
+          const SubmitHandler& submit_handler)>;
+
+      /**
+       * Signals the list of filted values and the filter mode.
```
spireken, line 91: Typo for 'filtered'.

Applications/Spire/Include/Spire/Ui/OpenFilterPanel.hpp
```diff
@@ -0,0 +1,153 @@
...
+       * @param submission The list of values that the user wants to filter.
+       * @param mode The filter mode.
+       */
+      using SubmitSignal =
+        Signal<void (const std::shared_ptr<AnyListModel>& submission,
+          FilterMode mode)>;
+
+      /** The default InputBox builder which builds an OpenInputBox */
```
spireken, line 99: Missing a period.

spireken (changes requested):
Looks nice.

Applications/Spire/Source/Ui/ComboBox.cpp
shaobinspire: I am also confused about that it doesn't set the keyboard focus and it still sends `FocusIn` event to the widget. The problem here I am trying to solve is to avoid the `OpenFilterPanel` being closed when clicking on the `DropDownList` to choose an item rather than to resolve focus issue. 

Both the `OverlayPanel` of the `OpenFilterPanel` and the `OverlayPanel` of the `DropDownList` inside the `ComboBox` have `Qt::Tool` window flag, and they don't become active by default when they are shown. When clicking on `DropDownList`, the `OverlayPanel` of the `OpenFilterPanel` will gets `FocusObserver::State::NONE` and be closed, which is not supposed to be closed. Calling `QApplication::setActiveWindow` when getting a submit signal from the `DropDownList` can prevent the `OverlayPanel` receiving `FocusObserver::State::NONE`.

spireken: You might be able to verify this by setting a breakpoint in [setActiveWindow](https://code.woboq.org/qt5/qtbase/src/widgets/kernel/qapplication.cpp.html#2080), but it looks like what the documentation means when it 'sets' the focusWidget is that it actually clears it.

shaobinspire: I debugged it by setting a breakpoint in 'QApplication::setActiveWindow', For this case, it calls `w->setFocus(Qt::ActiveWindowFocusReason);` on line 2084. 'QApplication::setActiveWindow' does two things: 

1. to deactivate current active window and activate the window passed in; 
2. to set the foucsWidget, which makes the widget in the window passed in get focus.

It does what the documentation says.

Applications/Spire/Source/Ui/FocusObserver.cpp
```diff
@@ -32,6 +32,9 @@ struct FocusObserver::FocusEventFilter : QObject {
   }
 
   void on_focus_changed(QWidget* old, QWidget* now) {
+    if(!now) {
```
spireken, line 35: Is it always the case that if no widget is focused, then the focus isn't considered updated?

shaobinspire: Good point. The fix is not correct, and I make another fix.
In the new fix, my thinking is when the current focused widget "now" is not m_widget and not the descendant of m_widget, the last focused widget "old" needs to be checked. If the "old" is m_widget or the descendant of m_widget, the focus state is set to `NONE`.

Note that there are two different functions used to check if a widget is another widget's ancestor. Spire::is_ancestor is used to check "now", while QWidget::isAncestorOf is used to check "old". The widget and its ancestor can be in different windows for Spire::is_ancestor, but the widget and its ancestor must be in a same window for QWidget::isAncestorOf.

Suppose there are two OverlayPanels A and B, and they are in different windows.  A is the parent of B, and A is shown. If B is opened, and gets activated and focus, using Spire::is_ancestor to check "now" can get the result that A is B's ancestor and the focus state of A is `FocusIn`, so the focused B doesn't make A closed. When B is closed and loses focus, using QWidget::isAncestorOf to check "old" can get the result that A isn't B's ancestor and the focus state of A doesn't update ('FocusIn'), so the closed B doesn't make A closed.

In this case, A is the OverlayPanel of the OpenFilterPanel and B is the OverlayPanel of the DropDownList inside the ComboBox.

kranar:
It's definitely very good work, the implementation is really solid but unfortunately it suffers from a bad API. A bad implementation is temporary, can always fix the implementation in the future, but a bad API is forever and spreads across the codebase. Nevertheless there is a ton of good stuff in here and I'd say 90% of it will end up going into the final implementation, but the API needs to be cleaned up significantly. That is mostly my fault in this case since I said I'd help out with the API, we'd go over some concepts like type erasure and how to work with generic code, and then I forgot all about it over the holidays and didn't give you any good feedback on it, so don't worry too much. I will take some time today to think it over and update the Asana task, overall it's impressive.

PR #757, opened 2022-01-20
Lazy text box

spireken:


kranar:
Looks good but still needs a lot more work. The implementation can not change any semantics, for example this implementation has ```is_read_only() == true``` initially but the original implementation has ```is_read_only() == false```. The submission is also incorrect for a read-only/disabled TextBox. But overall it looks good.

spireken:
I address the read-only and submission issues, and added a 'rejected' test property that can be used to test the rejected signal and styling, which can be demonstrated by submitting the rejected string (by default, 'deny').

Applications/Spire/Source/Ui/TextBox.cpp
```diff
@@ -263,85 +398,60 @@ TextBox::TextBox(QString current, QWidget* parent)
...
+  return m_read_only->get();
 }
 
 void TextBox::set_read_only(bool read_only) {
-  if(m_line_edit->isReadOnly() == read_only) {
+  if(read_only == m_read_only->get()) {
+    return;
+  }
+  m_read_only->set(read_only);
+  if(!m_line_edit && m_read_only->get()) {
```
kranar, line 442: Fix up these conditions, apply boolean algebra to avoid extraneous checks and clauses.

Applications/Spire/Source/Ui/TextBox.cpp
```diff
@@ -263,85 +398,60 @@ TextBox::TextBox(QString current, QWidget* parent)
...
+    match(*this, ReadOnly());
     return;
+  } else if(!m_line_edit) {
+    initialize_line_edit();
   }
-  m_line_edit->setReadOnly(read_only);
-  m_line_edit->setCursorPosition(0);
   setCursor(m_line_edit->cursor());
-  if(read_only) {
+  if(m_read_only->get()) {
```
kranar, line 449: Similarly there's a lot of context switching here. It really is as simple as if the component is read-only, match the read only selector, otherwise unmatch it. Avoid intertwining different non-local aspects of a model.

Applications/Spire/Source/Ui/TextBox.cpp
```diff
@@ -263,85 +398,60 @@ TextBox::TextBox(QString current, QWidget* parent)
 TextBox::TextBox(std::shared_ptr<TextModel> current, QWidget* parent)
     : QWidget(parent),
       m_current(std::move(current)),
-      m_submission(m_current->get()),
+      m_submission(std::make_shared<LocalTextModel>(m_current->get())),
+      m_read_only(std::make_shared<LocalBooleanModel>(false)),
```
kranar, line 402: I can maybe sympathize with the submission being a model, but read-only should not be a model. Models introduce a form of non-local reasoning along with all the problems that go with it (reentrancy bugs being the big one). If something is a model then I have to go over all connections to it to verify that there's no reentrancy issues, as opposed to if something is a local field then all of its uses are isolated.

Applications/Spire/Source/Ui/TextBox.cpp
```diff
@@ -263,85 +398,60 @@ TextBox::TextBox(QString current, QWidget* parent)
...
 }
 
 const std::shared_ptr<HighlightModel>& TextBox::get_highlight() const {
   return m_highlight;
 }
 
 void TextBox::set_placeholder(const QString& placeholder) {
-  m_box->set_placeholder(placeholder);
-  update_placeholder_text();
+  m_placeholder = placeholder;
```
kranar, line 427: Remove redundant assignments.

Applications/Spire/Source/Ui/TextBox.cpp
```diff
@@ -364,264 +474,125 @@ QSize TextBox::sizeHint() const {
...
     return 1;
   }();
+  auto metrics = QFontMetrics(m_text_style.m_font);
   m_size_hint.emplace(
-    m_line_edit->fontMetrics().horizontalAdvance(m_current->get()) +
-      cursor_width, m_line_edit->fontMetrics().height());
-  *m_size_hint += compute_decoration_size();
+    metrics.horizontalAdvance(m_current->get()) +
+      cursor_width, metrics.height());
+  *m_size_hint += m_geometry.get_geometry().size() -
```
kranar, line 481: This is nice.

Applications/Spire/Source/UiViewer/StandardUiProfiles.cpp
```diff
@@ -519,6 +519,21 @@ namespace {
...
+      }
+      QValidator::State get_state() const {
+        if(get() == m_rejected) {
+          return QValidator::Invalid;
+        }
+        return QValidator::Acceptable;
+      }
+    private:
+      QString m_rejected;
+  };
```
kranar, line 536: Very nice.

Applications/Spire/Include/Spire/Ui/TextBox.hpp
```diff
@@ -158,56 +160,46 @@ namespace Styles {
...
+        QLineEdit::EchoMode m_echo_mode;
 
-        StyleProperties(std::function<void ()> commit);
-        void clear();
+        TextStyleProperties();
       };
       struct TextValidator;
-      class PlaceholderBox;
+      class LineEdit;
+      friend class LineEdit;
```
kranar, line 180: Remove the ```friend```:

https://en.cppreference.com/w/cpp/language/nested_types

Applications/Spire/Source/Ui/TextBox.cpp
```diff
@@ -105,154 +105,289 @@ struct TextBox::TextValidator : QValidator {
...
+        const BoxGeometry& geometry, const TextStyleProperties& text_style) {
+      m_text_box->layout()->setContentsMargins(
+        geometry.get_padding_left(), geometry.get_padding_top(),
+        geometry.get_padding_right(), geometry.get_padding_bottom());
+      auto stylesheet = QString(
+        R"(#0x%1 {
+          background: transparent;
+          border-width: 0px;
+          padding: 0px;)").arg(reinterpret_cast<std::intptr_t>(this));
+      auto map = StyleSheetMap([] {});
```
kranar, line 184: You don't need to construct an entire ```StyleSheetMap``` in order to specify the text color.

Applications/Spire/Source/Ui/TextBox.cpp
```diff
@@ -105,154 +105,289 @@ struct TextBox::TextValidator : QValidator {
...
+      map.set("color", text_style.m_text_color);
+      map.write(stylesheet);
+      if(text_style.m_alignment != alignment()) {
+        setAlignment(text_style.m_alignment);
+      }
+      auto font = text_style.m_font;
+      font.setPixelSize(text_style.m_size);
```
kranar, line 191: Can you set the pixel size directly in ```text_style.m_font``` instead of having a separate field ```text_style.m_size``` for it?

Applications/Spire/Source/Ui/TextBox.cpp
```diff
@@ -105,154 +105,289 @@ struct TextBox::TextValidator : QValidator {
...
+          QLineEdit::keyPressEvent(e);
+          m_is_handling_key_press = false;
+          if(e->isAccepted()) {
+            return true;
+          }
+        }
+      }
+      return QLineEdit::eventFilter(watched, event);
+    }
+    void focusInEvent(QFocusEvent* event) override {
```
kranar, line 235: Vertical whitespace missing.

Applications/Spire/Source/Ui/TextBox.cpp
```diff
@@ -105,154 +105,289 @@ struct TextBox::TextValidator : QValidator {
...
+          focus_event->reason() != Qt::PopupFocusReason) {
+        m_text_validator->m_is_text_elided = false;
+        if(text() != m_current->get()) {
+          setText(m_current->get());
+        }
+      }
+      QLineEdit::focusInEvent(event);
+    }
+    void focusOutEvent(QFocusEvent* event) override {
+      auto focusEvent = static_cast<QFocusEvent*>(event);
```
kranar, line 247: What's the purpose of this cast?

kranar (changes requested):
Overall it's very good but still needs a lot more work.

Applications/Spire/Source/Ui/TextBox.cpp
spireken: This keeps coming up, so do we always add whitespace after every method definition, regardless of access specifier?

kranar: No different from ordinary function definitions.

shaobinspire (changes requested):

kranar:
This is good. I am going to make some minor changes to it but overall the PR is good and will be approved.

shaobinspire (approved):

kranar (changes requested):
Almost ready to be merged.

kranar:
Well done, these concepts will be revisited and built upon further in future components.

PR #758, opened 2022-01-21
Any ref

shaobinspire:


Applications/Spire/Include/Spire/Spire/AnyRef.hpp
```diff
@@ -0,0 +1,355 @@
...
+        virtual const void* get_const() const = 0;
+        virtual volatile void* get_volatile() const = 0;
+        virtual const volatile void* get_const_volatile() const = 0;
+        virtual const std::type_info& get_type() const = 0;
+        virtual constexpr bool is_const() const = 0;
+        virtual constexpr bool is_volatile() const = 0;
+        virtual constexpr bool is_const_volatile() const = 0;
+        virtual ~AnyRefBase() = default;
+      };
+      template<typename T>
```
kranar, line 101: Look at implementing this without these classes, this class is intended to be very lightweight and efficient.

kranar:
This implementation does use one strategy for type-erasure but it's the full blown heavyweight approach. This class is intended to very lightweight, so it can not use this approach, instead the implementation should be in terms of the following:

    enum class Qualifiers : std::uint8_t {
      NONE = 0,
      CONST = 1,
      VOLATILE = 2
    };
    void* m_ptr;
    std::type_index m_type;
    Qualifiers m_qualifiers;

PR #759, opened 2022-01-24
AnyValueModel

shaobinspire:


Applications/Spire/Include/Spire/Spire/AnyValueModel.hpp
```diff
@@ -0,0 +1,49 @@
...
+    public:
+      using Type = ValueModel<AnyRef>::Type;
+
+      using UpdateSignal = ValueModel<AnyRef>::UpdateSignal;
+
+      /**
+       * Constructs an AnyValueModel over an existing ValueModel.
+       * @param source The ValueModel to wrap.
+       */
+      template<typename T>
```
spireken, line 24: Should these use Type instead of T?

spireken (commented):
Looks good.

Applications/Spire/Include/Spire/Spire/AnyValueModel.hpp
shaobinspire: `Type` and `T` are different. `Type` of the AnyValueModel is `AnyRef`, while `T` used in the constructor could be of any type. Taking the unit test in AnyValueModelTester for example, T is `int`.

spireken (approved):

Applications/Spire/Include/Spire/Spire/AnyValueModel.hpp
```diff
@@ -0,0 +1,49 @@
...
+        const UpdateSignal::slot_type& slot) const override;
+
+    private:
+      mutable UpdateSignal m_update_signal;
+      Type m_value;
+      std::function<void (const Type&)> m_set_value;
+  };
+
+  template<typename T>
+  AnyValueModel::AnyValueModel(std::shared_ptr<ValueModel<T>> source)
```
kranar, line 43: ```source``` is a ```std::shared_ptr``` passed by value but no ownership of it is ever taken, that will result in a dangling reference.

Applications/Spire/Include/Spire/Spire/AnyValueModel.hpp
```diff
@@ -0,0 +1,49 @@
...
+    : m_value(const_cast<T&>(source->get())),
```
kranar, line 44: This violates const-correctness, can engender undefined behavior, and violates several of model's invariants, such as all mutations should result in a signal.

Applications/Spire/Include/Spire/Spire/AnyValueModel.hpp
```diff
@@ -0,0 +1,49 @@
...
+      m_set_value([=] (const Type& value) {
+        any_cast<T>(m_value) = any_cast<const T>(value); }) {}
```
kranar, line 46: This implementation leads to dangling references (```value```'s ownership is never expressed).

Applications/Spire/Source/Common/AnyValueModel.cpp
```diff
@@ -0,0 +1,30 @@
...
+using namespace boost;
+using namespace boost::signals2;
+using namespace Spire;
+
+const AnyValueModel::Type& AnyValueModel::get() const {
+  return m_value;
+}
+
+QValidator::State AnyValueModel::test(const Type& value) const {
+  if(m_value.has_value()) {
```
kranar, line 12: ```value``` is never used, it's not clear what this method is doing.

kranar (changes requested):
Too many issues with this, almost nothing about this implementation is correct. Revise it and resubmit a new PR.

PR #762, opened 2022-01-29
OpenFilterPanel

shaobinspire:


kranar:
Lot of good work, looks almost done but some minor changes need to be made.

Applications/Spire/Include/Spire/Spire/AnyValueModel.hpp
```diff
@@ -101,6 +101,9 @@ namespace Spire {
   template<typename T>
   QValidator::State AnyValueModel::WrapperValueModel<T>::set(
       const Type& value) {
+    if(!value.has_value()) {
```
kranar, line 104: Undo this change.

Applications/Spire/Source/Ui/OpenFilterPanel.cpp
```diff
@@ -0,0 +1,377 @@
...
+    style.get((Hover() || Press()) / Body() >> is_a<Icon>()).
+      set(BackgroundColor(QColor(Qt::transparent)));
+    style.get(Disabled() / Body() >> is_a<Icon>()).
+      set(Visibility::NONE);
+    return style;
+  }
+
+  auto LIST_ITEM_STYLE() {
+    auto style = StyleSheet();
+    style.get((Any() || Hover() || Press() || Focus() || Selected())).
```
kranar, line 34: Remove extraneous parenthesis.

Applications/Spire/Source/Ui/OpenFilterPanel.cpp
```diff
@@ -0,0 +1,377 @@
...
+      set(BackgroundColor(QColor(0xFFFFFF))).
+      set(border_size(0)).
+      set(PaddingRight(scale_width(10)));
+    return style;
+  }
+
+  auto display_text(OpenFilterPanel::Mode mode) {
+    if(mode == OpenFilterPanel::Mode::INCLUDE) {
+      static const auto value = QObject::tr("Include");
+      return value;
```
kranar, line 44: Avoid making copies.

Applications/Spire/Source/Ui/OpenFilterPanel.cpp
```diff
@@ -0,0 +1,377 @@
...
+    } else {
+      static const auto value = QObject::tr("Exclude");
+      return value;
+    }
+  }
+}
+
+class DeletableItem : public QWidget {
+  public:
+    using DeleteSignal = Signal<void()>;
```
kranar, line 54: Space after ```void```.

Applications/Spire/Source/Ui/OpenFilterPanel.cpp
```diff
@@ -0,0 +1,377 @@
...
+    }
+  }
+
+  std::any parse(const QString& query) override {
+    auto value = m_source->parse(query);
+    if(!value.has_value()) {
+      return value;
+    }
+    if(m_matches_set.contains(displayTextAny(value))) {
+      static auto result = std::any();
```
kranar, line 106: What is the purpose of this ```static``` variable?

Applications/Spire/Source/Ui/OpenFilterPanel.cpp
```diff
@@ -0,0 +1,377 @@
...
+  QtPromise<std::vector<std::any>> submit(const QString& query) override {
+    return m_source->submit(query).then([=] (auto&& source_result) {
+      auto result = [&] {
+        try {
+          return source_result.Get();
+        } catch(const std::exception&) {
+          return std::vector<std::any>();
+        }
+      }();
+      std::erase_if(result, [=] (const auto& value) {
```
kranar, line 121: Incorrect capture.

Applications/Spire/Source/Ui/OpenFilterPanel.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+    void on_update(Mode mode, bool value) {
+      get_button(mode)->get_current()->set(value);
+      if(value) {
+        auto blocker = shared_connection_block(m_connection);
+        m_model->set(mode);
+      }
+    }
+
+    void on_model_update(const Mode mode) {
```
kranar, line 181: Don't pass by ```const``` value.

Applications/Spire/Source/Ui/OpenFilterPanel.cpp
```diff
@@ -0,0 +1,377 @@
...
+  return item;
+}
+
+void OpenFilterPanel::on_input_box_submission(const AnyRef& submission) {
+  m_input_box->get_current()->set({});
+  m_matches->push(submission);
+}
+
+void OpenFilterPanel::on_list_view_current(const optional<int>& current) {
+  m_list_view->setFocusPolicy(Qt::NoFocus);
```
kranar, line 338: Why is the list view's focus policy constantly being set?

Applications/Spire/Source/Ui/OpenFilterPanel.cpp
```diff
@@ -0,0 +1,377 @@
...
+void OpenFilterPanel::on_mode_current(Mode mode) {
+  submit();
+}
+
+void OpenFilterPanel::on_matches_operation(
+    const AnyListModel::Operation& operation) {
+  visit(operation,
+    [&] (const AnyListModel::AddOperation& operation) {
+      auto item = m_list_view->get_list_item(operation.m_index);
+      set_style(*item, LIST_ITEM_STYLE());
```
kranar, line 350: Why is the style being set here on every add operation?

Applications/Spire/Include/Spire/Spire/AnyValueModel.hpp
shaobinspire: The intention is to clear the value of the `AnyValueModel` from the `OpenFilterPanel`. The `OpenFilterPanel` knows that the type of the current of the `AnyInputBox` is `AnyRef`, and it can set an empty `AnyRef` to the `current` to clear the `AnyInputBox`. If this code gets removed, `any_cast` will return a invalid pointer because the type of the empty `AnyRef` is void but the casted type is `std::any` which is the type of the `current` of the `ComboBox`, and there is no way to clear the `AnyInputBox` from the `OpenFilterPanel`.

Applications/Spire/Source/Ui/OpenFilterPanel.cpp
shaobinspire: I tried to set the style of the `ListItem` in the `ListView`, but the style couldn’t be applied to the newly added `ListItem` through the `ListView`.

shaobinspire: I found a solution to solve it.

Applications/Spire/Source/Ui/OpenFilterPanel.cpp
```diff
@@ -0,0 +1,358 @@
...
+
+void OpenFilterPanel::on_mode_current(Mode mode) {
+  submit();
+}
+
+void OpenFilterPanel::on_matches_operation(
+    const AnyListModel::Operation& operation) {
+  visit(operation,
+    [&] (const AnyListModel::AddOperation& operation) {
+      update_style(*m_list_view, [] (auto&) {});
```
kranar, line 333: What does this do?

Applications/Spire/Source/Ui/ScrollBar.cpp
```diff
@@ -236,6 +236,7 @@ void ScrollBar::update_thumb() {
       style.get(Any()).set(PaddingLeft(m_thumb_position));
     }
   });
+  update();
```
kranar, line 239: What does this do?

kranar (changes requested):
Major improvement.

Applications/Spire/Source/Ui/OpenFilterPanel.cpp
shaobinspire: When a ListItem is added to the ListView, the newly added ListItem cannot be styled, so I have to style the ListItem or update the style of the ListView when the add operation happens.

kranar: That's good, when you come across bugs in the future, instead of trying to work around them, report them in Asana so they can get fixed.

Applications/Spire/Source/Ui/ScrollBar.cpp
shaobinspire: Here I am trying to solve the bug of the ScrollBar. When enough ListItems are added into the ListView and the vertical scroll bar becomes visible, the thumb is not draggable even though the ListItems can move with the mouse dragging.

I will remove it and report the issue in Asana.

kranar: You can keep this in here, just wasn't sure what it was doing.

Applications/Spire/Source/Ui/OpenFilterPanel.cpp
shaobinspire: For this one, should I remove it or keep it?

kranar: You can keep it, but I need to fix this bug, so in the future just find a way to make it clear that this is working around a bug, preferably by creating a task in Asana. In your Asana task you can refer to this line (along with the commit) and point out that this is a workaround to a bug.

kranar (approved):

PR #763, opened 2022-01-31
 FocusObserver Issue

shaobinspire:
Fixed the re-entrancy issue which causes the iterators of m_entries to be unavailable during erasing.


kranar (approved):
Good catch and approach to fixing it. I made some minor modifications.

PR #767, opened 2022-02-07
ComboBox

shaobinspire:
Replaced the TextBox in the ComboBox with the AnyInputBox.

### DIFF
--- a/Applications/Spire/Source/Ui/ComboBox.cpp
+++ b/Applications/Spire/Source/Ui/ComboBox.cpp
@@ -16,11 +16,6 @@ using namespace boost::signals2;
 using namespace Spire;
 using namespace Spire::Styles;
 
-AnyInputBox* ComboBox::default_input_box_builder() {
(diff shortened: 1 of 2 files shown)


spireken (approved):

Applications/Spire/Source/Ui/ComboBox.cpp
```diff
@@ -39,7 +51,7 @@ ComboBox::ComboBox(std::shared_ptr<QueryModel> query_model,
       m_has_autocomplete_selection(false),
       m_current_connection(m_current->connect_update_signal(
         std::bind_front(&ComboBox::on_current, this))) {
-  m_input_box = new TextBox();
+  m_input_box = input_box_builder();
```
kranar, line 54: If you don't need a function then use a plain value. As discussed in our meeting, a ```function``` is used only when the return value depends on side-effects or some combination of arguments.

kranar (approved):
Beautiful.

PR #768, opened 2022-02-09
Reposition the Overlay Panel.

shaobinspire:
Fixed the issue where the OverlayPanel doesn’t move with the height change of the parent widget.

When the position of the OverlayPanel’s ancestor (not direct parent) changes, whether it is caused by moving window or layout changing, the OverlayPanel might not get the Move event from its direct parent, because the position of direct parent might have not changed relative to the ancestor.

The added ParentMoveObserver is used to listen to the Move event of all ancestors of the OverlayPanel, including the top-level window. When the position of any one of the ancestors changes, the OverlayPanel can know the change and reposition to follow its parent.

### DIFF
--- a/Applications/Spire/Source/Ui/OverlayPanel.cpp
+++ b/Applications/Spire/Source/Ui/OverlayPanel.cpp
@@ -1,5 +1,4 @@
 #include "Spire/Ui/OverlayPanel.hpp"
-#include <QApplication>
 #include <QGraphicsDropShadowEffect>
 #include <QHBoxLayout>
 #include <QMouseEvent>
@@ -45,55 +44,6 @@ namespace {
   }
 }
 
-class OverlayPanel::ParentMoveObserver : public QObject {
-  public:
-    using MoveSignal = Signal<void()>;
-
-    ParentMoveObserver(QWidget& widget) {
-      install_parent_event_filter(widget);
-    }
-
-    ~ParentMoveObserver() {
-      for(auto parent : m_parents) {
-        parent->removeEventFilter(this);
-      }
-    }
-
-    connection connect_move_signal(const MoveSignal::slot_type& slot) const {
-      return m_move_signal.connect(slot);
-    }
-
-    bool eventFilter(QObject* watched, QEvent* event) override {
-      if(event->type() == QEvent::Move) {
-        m_move_signal();
-      } else if(event->type() == QEvent::ParentChange) {
-        auto i = std::find(m_parents.begin(), m_parents.end(), watched);
-        auto j = std::next(i, 1);
-        while(j != m_parents.end()) {
-          (*j)->removeEventFilter(this);
-          j = m_parents.erase(j);
-        }
-        if(i != m_parents.end()) {
-          install_parent_event_filter(*(*i));
-        }
-      }
-      return QObject::eventFilter(watched, event);
-    }
-
-  private:
-    mutable MoveSignal m_move_signal;
-    std::list<QWidget*> m_parents;
-
-    void install_parent_event_filter(const QWidget& widget) {
-      auto parent = widget.parentWidget();
-      while(parent) {
-        m_parents.push_back(parent);
-        parent->installEventFilter(this);
-        parent = parent->parentWidget();
-      }
-    }
-};
-
 OverlayPanel::OverlayPanel(QWidget& body, QWidget& parent)
     : QWidget(&parent,
         Qt::Tool | Qt::FramelessWindowHint | Qt::NoDropShadowWindowHint),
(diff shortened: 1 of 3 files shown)


Applications/Spire/Source/Ui/OverlayPanel.cpp
```diff
@@ -45,6 +45,55 @@ namespace {
...
+        if(i != m_parents.end()) {
+          install_parent_event_filter(*(*i));
+        }
+      }
+      return QObject::eventFilter(watched, event);
+    }
+
+  private:
+    mutable MoveSignal m_move_signal;
+    std::list<QWidget*> m_parents;
```
spireken, line 85: Could you use a std::unordered_set for this instead? It looks like it's iterated over, searched, and written to, so the search could become constant while still supporting the other operations.

Applications/Spire/Source/Ui/OverlayPanel.cpp
```diff
@@ -45,6 +45,55 @@ namespace {
...
+      }
+    }
+
+    connection connect_move_signal(const MoveSignal::slot_type& slot) const {
+      return m_move_signal.connect(slot);
+    }
+
+    bool eventFilter(QObject* watched, QEvent* event) override {
+      if(event->type() == QEvent::Move) {
+        m_move_signal();
```
spireken, line 68: Is it possible that a single move could cause a move event to be sent to multiple parents, or is it always the case that only an immediate widget is affected by a move?

Applications/Spire/Source/Ui/OverlayPanel.cpp
shaobinspire: The parents are stored in the list in order from the most direct parent to the top-level parent. When the parent of a widget stored in the list (m_parents) changes, all parents behind the widget in the list need to be removed and new parents of the widget need to be added in order. 

For example, suppose m_parents has a list of [`parent1`, `parent2`, `parent3`, `parent4`], `parent2` is the parent of `parent1`, and `parent3` is the parent of `parent2`, …  and `parent4` is the top-level parent. When the parent of `parent2` changes, `parent3` and `parent4` need to be removed from the list, and new parents are added after `parent2`, like [`parent1`, `parent2`, `new-parent3`, `new-parent4`, `new-parent5`].

`std::unordered_set` can’t store values in order, so it is not a suitable container. Another alternative is `std::vector`, but the list might need to erase and add items multiple times, so I think `std::list` is better than `std::vector`.

Applications/Spire/Source/Ui/OverlayPanel.cpp
shaobinspire: Refer to Qt documentation: “Move events are sent to widgets that have been moved to a new position relative to their parent.”
https://doc.qt.io/qt-5/qmoveevent.html#details

It means that only when the position relative to the parent changes, the move event is emitted. The parent moves, hence the rect of the widget moves, but children's rects are still left exactly the same relative to their parent's rect, so they aren’t moving, hence no move event.

However, it is possible that multiple move events are sent from different widgets when a widget is moving. For example, a parent widget moves and one of its children moves to a new position relative to the parent widget for some reason, the parent and the child respectively trigger move events, but these two move events are different.

spireken (approved):

Applications/Spire/Source/Ui/OverlayPanel.cpp
```diff
@@ -45,6 +45,55 @@ namespace {
...
+  public:
+    using MoveSignal = Signal<void()>;
+
+    ParentMoveObserver(QWidget& widget) {
+      install_parent_event_filter(widget);
+    }
+
+    ~ParentMoveObserver() {
+      for(auto parent : m_parents) {
+        parent->removeEventFilter(this);
```
kranar, line 58: Why is this destructor needed?

Applications/Spire/Source/Ui/OverlayPanel.cpp
```diff
@@ -45,6 +45,55 @@ namespace {
   }
 }
 
+class OverlayPanel::ParentMoveObserver : public QObject {
+  public:
+    using MoveSignal = Signal<void()>;
```
kranar, line 50: ```void ()```

Applications/Spire/Source/Ui/OverlayPanel.cpp
```diff
@@ -45,6 +45,55 @@ namespace {
   }
 }
 
+class OverlayPanel::ParentMoveObserver : public QObject {
```
kranar, line 48: What functionality does this provide that the ```GlobalPositionObserver``` does not?

shaobinspire: The observer provides the same functionality as the `GlobalPositionObserver`.  I will replace it with `GlobalPositionObserver`.

kranar (approved):

PR #779, opened 2022-02-18
Key filter panel

shaobinspire:


Applications/Spire/Source/Ui/KeyFilterPanel.cpp
```diff
@@ -0,0 +1,103 @@
...
+  }
+};
+
+AnyInputBox* key_input_box_builder(
+    std::shared_ptr<KeySequenceValueModel> current,
+    std::shared_ptr<AnyListModel> matches) {
+  auto input_box = new AnyInputBox(*(new KeyInputBox(
+    std::make_shared<KeyInputBoxValueModel>(std::move(current), matches))));
+  input_box->connect_submit_signal([=] (const auto& submission) {
+    if(!any_cast<QKeySequence>(submission).isEmpty()) {
```
kranar, line 75: Eliminate common subexpressions.

Applications/Spire/Source/Ui/KeyFilterPanel.cpp
```diff
@@ -0,0 +1,103 @@
...
+      [&] (const AnyListModel::RemoveOperation& operation) {
+        m_matches_set.erase(m_matches_list[operation.m_index]);
+        m_matches_list.erase(m_matches_list.begin() + operation.m_index);
+      });
+  }
+
+  void add_match(int index) {
+    auto value = displayTextAny(m_matches->get(index));
+    m_matches_set.insert(value);
+    m_matches_list.insert(m_matches_list.begin() + index, value);
```
kranar, line 65: What does ```m_matches_list``` provide that ```m_matches``` does not?

kranar (changes requested):
This looks like a big improvement.

Applications/Spire/Source/Ui/KeyFilterPanel.cpp
shaobinspire: When `KeyInputBoxValueModel` gets the `RemoveOperation` signal, the item has been removed from `m_matches`, and `m_matches_set` doesn't know which item should be removed. `m_matches_list` helps `m_matches_set` to quickly remove items.

Applications/Spire/Source/Ui/KeyFilterPanel.cpp
```diff
@@ -0,0 +1,103 @@
...
+  QValidator::State set(const Type& value) {
+    if(m_matches_set.contains(value.toString())) {
+      return QValidator::State::Invalid;;
+    }
+    return m_source->set(value);;
+  }
+
+  connection connect_update_signal(
+      const UpdateSignal::slot_type& slot) const override {
+    return m_source->connect_update_signal(slot);
```
kranar, line 48: Because ```KeyInputBoxValueModel``` doesn't own ```m_source```, it can not just directly forward slots to ```m_source```. When ```KeyInputBoxValueModel``` is destroyed, all connections have to be destroyed as well.

Applications/Spire/Source/Ui/KeyFilterPanel.cpp
shaobinspire: It prevents an empty submission from adding into matches when the `KeyInputBox` loses focus.

shaobinspire: When an empty `KeyInputBox` loses focus, it will submit an empty value. Maybe I need to create a task for the issue.

Applications/Spire/Source/Ui/KeyFilterPanel.cpp
```diff
@@ -45,7 +49,11 @@ struct KeyInputBoxValueModel : ValueModel<QKeySequence> {
 
   connection connect_update_signal(
       const UpdateSignal::slot_type& slot) const override {
-    return m_source->connect_update_signal(slot);
+    return m_update_signal.connect(slot);
+  }
+
+  void on_current(const QKeySequence& current) {
+    m_update_signal(current);
```
kranar, line 56: You can elide this entirely my passing ```m_update_signal``` to ```m_source->connect_update_signal```.

PR #780, opened 2022-02-18
DestinationFilterPanel

shaobinspire:


spireken (approved):

Applications/Spire/Source/Ui/DestinationFilterPanel.cpp
```diff
@@ -0,0 +1,105 @@
+#include "Spire/Ui/DestinationFilterPanel.hpp"
+#include "Spire/Ui/AnyInputBox.hpp"
+#include "Spire/Ui/CustomQtVariants.hpp"
+#include "Spire/Ui/DestinationBox.hpp"
+
+using namespace boost::signals2;
+using namespace Nexus;
+using namespace Spire;
+
+struct DestinationFilterQueryModel : ComboBox::QueryModel {
```
kranar, line 10: Avoid repetition, if you find yourself copying/pasting code, consider finding a way of implementing a solution that reduces the amount of copying/pasting.

shaobinspire: I proposed the `ExcludingQueryModel` to avoid implementing two similar models which are used for the `SecurityFIlterModel` and `DestinationFilterModel`. The model accepts a function to convert `std::any` to `QString`, which can be customized by the `SecurityFilterModel` and `DestinationFilterModel`.

https://gist.github.com/shaobinspire/1cffead77a60807b26a534d2665d90d1

kranar: What is the purpose of the the ```Converter```?

shaobinspire: The `Converter` is used to convert `std::any` to a `QString` based on the specific type. For example, for the DestinationFilterPanel, the `Converter` casts the `Destination` to `QString`:
```
  auto box = new DestinationBox(std::make_shared<ExcludingQueryModel>(model, matches,
      [] (const std::any& value) {
        return displayTextAny(std::any_cast<const DestinationDatabase::Entry&>(value).m_id);
      }));
```

For the SecurityFilterPanel, the `Converter` casts the `Security` to `QString`:
```
  auto box = new SecurityBox(std::make_shared<ExcludingQueryModel>(model, matches,
      [] (const std::any& value) {
        return displayTextAny(std::any_cast<SecurityInfo&>(value).m_security);
      }));
```

kranar: That tells me what it does, it does not tell me its purpose.

kranar: I don't follow the API. ```ExcludingQueryModel``` is now a template parameterized over a type ```T``` but ```T``` is not used anywhere. Think carefully about the API and then feel free to propose it on Asana.

Applications/Spire/Source/Common/ExcludingQueryModel.cpp
```diff
@@ -0,0 +1,74 @@
...
+      add_exclusion(operation.m_index);
+    },
+    [&] (const AnyListModel::RemoveOperation& operation) {
+      m_exclusion_set.erase(m_exclusion_list[operation.m_index]);
+      m_exclusion_list.erase(m_exclusion_list.begin() + operation.m_index);
+    });
+}
+
+QString ExcludingQueryModel::to_string(const std::any& value) {
+  if(value.type() == typeid(DestinationDatabase::Entry)) {
```
kranar, line 67: Remove lines 67-72.

kranar (changes requested):

Applications/Spire/Source/Common/ExcludingQueryModel.cpp
shaobinspire: When the `ExcludingQueryModel` is used for `DestinationFilterPanel`, the type of the query result is `DestinationDatabase::Entry`, but the type of data in exclusions is `Destination`, so they are needed to get `Destination` from `DestinationDatabase::Entry` in order to compare the query result and exclusion.

shaobinspire: A alternative would be to add a virtual member function `to_string` to the ExcludingQueryModel's API:
```
      virtual QString to_string(const std::any& value);
```
When the DestinationFilterPanel uses this model, a derived class like `DestinationExcludingQueryModel` is created to override the virtual function where `Destination` is retrieved from `DestinationDatabase::Entry`.

```
class DestinationExcludingQueryModel : public ExcludingQueryModel {
  public:
    using ExcludingQueryModel::ExcludingQueryModel;

    QString to_string(const std::any& value) override {
      return displayTextAny(
        std::any_cast<const DestinationDatabase::Entry&>(value).m_id);
    }
};
```

kranar:
Resubmit when complete.

kranar:
Get feedback on it and then submit it for review. I unfortunately am too busy to do a lot of back and forths, back and forths and this is entirely the wrong approach to take.

shaobinspire:
I moved to_string from public to private according to the guideline2 "Prefer to make virtual functions private" in the following article:
http://www.gotw.ca/publications/mill18.htm

PR #783, opened 2022-02-28
Local imbalance table model

spireken:


shaobinspire (changes requested):

shaobinspire (approved):

Applications/Spire/Source/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorTableModel.cpp
```diff
@@ -0,0 +1,95 @@
...
+    }
+    return std::make_any<T>(potential);
+  }
+}
+
+void LocalOrderImbalanceIndicatorTableModel::add(
+    const OrderImbalance& imbalance) {
+  auto& previous_imbalance = m_imbalances[imbalance.m_security];
+  if(previous_imbalance.m_imbalance != OrderImbalance()) {
+    if(previous_imbalance.m_imbalance.m_timestamp < imbalance.m_timestamp) {
```
kranar, line 23: Remove redundant/nested if statements.

Applications/Spire/Source/OrderImbalanceIndicator/LocalOrderImbalanceIndicatorTableModel.cpp
```diff
@@ -0,0 +1,95 @@
...
+}
+
+std::vector<std::any> LocalOrderImbalanceIndicatorTableModel::make_row(
+    const OrderImbalance& imbalance) const {
+  return {imbalance.m_security, imbalance.m_side, imbalance.m_size,
+    imbalance.m_referencePrice, imbalance.m_size * imbalance.m_referencePrice,
+    imbalance.m_timestamp.date(), imbalance.m_timestamp.time_of_day()};
+}
+
+std::vector<boost::optional<std::any>>
```
kranar, line 84: Remove ```boost::```

Applications/Spire/Include/Spire/SpireTester/TableOperationLog.hpp
```diff
@@ -0,0 +1,56 @@
+#ifndef SPIRE_TABLE_OPERATION_LOG_HPP
+#define SPIRE_TABLE_OPERATION_LOG_HPP
+#include "Spire/Ui/TableModel.hpp"
+
+namespace Spire {
+
+  /** Logs operations performed on a TableModel. */
+  class TableOperationLog {
+    public:
```
kranar, line 9: Look at how other models are tested and either reuse that or justify the purpose of this class.

kranar (changes requested):

Applications/Spire/Include/Spire/SpireTester/TableOperationLog.hpp
spireken: Since I'll have three different imbalance table models to test my goal with this class was to avoid repeating the same or similar operation logging code for each test case like the other TableModel implementations do in their tests, which also avoids having to manually manipulate the data structure that contains the logged operations. Though, an improvement would be to remove the getters for each vector of the typed Operations and instead have a single getter that takes an index and returns the operation performed at that index in the sequence of operations, or returns none if the index is invalid. This would allow the ordering of operations to be tested as well instead of treating each type of operation as a separate sequence.

kranar: What similar logging code are you referring to? The code to a log an operation is to connect a slot to the operation signal that pushes the operation to a ```std::deque<Operation>```.

spireken: In looking over the Array, Filtered, Sorted table model tests the most common cases when testing operations are verifying operation counts (total and individual counts), verifying the operation was performed on specific rows and/or columns, and verifying what order the operations were performed in. Every test case that tests operations has to then connect to the operation signal and store additional variables to track what it needs to test, so I would prefer to offload that to a class that can encapsulate all of the logging behavior. I was too focused on my particular use case so the API isn't sufficient for every case, but it could be improved by changing the internal vector to a deque and having getters that return and remove the first and last elements.

A general use case would roughly look like this, and avoids having to manually maintain the counts , deque, and connection:

`auto model = std::make_shared<SomeTableModel>();
auto log = TableOperationLog(model);
model->do_something(abc);
REQUIRE(log.operation_count() == 1);
REQUIRE(log.add_count() == 1);
auto operation1 = get<TableModel::AddOperation>(log.get_front());
REQUIRE(operation1);
REQUIRE(operation1->m_index == 1);`

kranar: I see, have a look at the ```ArrayListModelTester``` to see how to unit test this using a ```std::deque```.

kranar (approved):

PR #785, opened 2022-03-01
Order imbalance table model

spireken:
I originally proposed an OrderImbalanceIndicatorModelSubscriber class that would manage the subscription to a model, but I removed it because it didn't provide enough value in practice to make maintaining it worthwhile.


Applications/Spire/Include/Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorTableModel.hpp
```diff
@@ -0,0 +1,45 @@
...
+        std::shared_ptr<OrderImbalanceIndicatorModel> source);
+
+      int get_row_size() const override;
+
+      int get_column_size() const override;
+
+      const std::any& at(int row, int column) const override;
+
+      boost::signals2::connection connect_operation_signal(
+        const typename OperationSignal::slot_type& slot) const override;
```
shaobinspire, line 32: Remove `typename`.

Applications/Spire/Source/OrderImbalanceIndicator/OrderImbalanceIndicatorTableModel.cpp
```diff
@@ -0,0 +1,46 @@
...
+  return m_table.get_column_size();
+}
+
+const std::any& OrderImbalanceIndicatorTableModel::at(
+    int row, int column) const {
+  return m_table.at(row, column);
+}
+
+connection OrderImbalanceIndicatorTableModel::connect_operation_signal(
+    const typename OperationSignal::slot_type& slot) const {
```
shaobinspire, line 32: Remove `typename`.

Applications/Spire/Source/OrderImbalanceIndicator/OrderImbalanceIndicatorTableModel.cpp
```diff
@@ -0,0 +1,46 @@
+#include "Spire/OrderImbalanceIndicator/OrderImbalanceIndicatorTableModel.hpp"
+
+using namespace boost::signals2;
+using namespace Nexus;
+using namespace Spire;
+
+OrderImbalanceIndicatorTableModel::OrderImbalanceIndicatorTableModel(
+    TimeInterval interval,
+    std::shared_ptr<OrderImbalanceIndicatorModel> source)
+    : m_source(std::move(source)) {
```
shaobinspire, line 10: Why is the source stored in the class? It isn't used in other member functions except in the constructor.

shaobinspire (changes requested):

Applications/Spire/Source/OrderImbalanceIndicator/OrderImbalanceIndicatorTableModel.cpp
spireken: It keeps a pointer to it so the subscription won't be cancelled if whatever else is holding a pointer to the OrderImbalanceIndicatorModel is destroyed.

shaobinspire (approved):

Applications/Spire/Include/Spire/SpireTester/SpireTester.hpp
```diff
@@ -29,6 +29,12 @@ namespace Spire {
       });
     app.exec();
   }
+
+  /**
+   * Processes events in the Qt event loop until the given predicate returns
+   * true.
+   */
+  void wait_until(const std::function<bool ()>& predicate);
```
kranar, line 37: Changes to the public API need to be approved. Remove this function.

spireken: I'm looking into a TestOrderImbalanceIndicatorModel than can resolve subscriptions manually.

kranar (approved):

PR #791, opened 2022-03-09
Current imbalance model

spireken:


Applications/Spire/Source/SpireTester/CurrentOrderImbalanceIndicatorTableModelTester.cpp
```diff
@@ -0,0 +1,292 @@
...
+    REQUIRE(model.get_row_size() == 3);
+    REQUIRE(rows_equal(model, {B350, C110, A100}));
+    environment.AdvanceTime(seconds(25));
+    REQUIRE(model.get_row_size() == 2);
+    REQUIRE(rows_equal(model, {B350, C110}));
+    environment.AdvanceTime(seconds(25));
+    REQUIRE(model.get_row_size() == 1);
+    REQUIRE(row_imbalance(model, 0) == B350);
+    source->publish(B150);
+    REQUIRE(model.get_row_size() == 1);
```
shaobinspire, line 260: From my understanding, since the current time is 325 (250+25+25+25) and the expiring time is 125 (325-200), B150 is later than 125 and it should be added into model and the row size should be 2.

shaobinspire (changes requested):

Applications/Spire/Source/SpireTester/CurrentOrderImbalanceIndicatorTableModelTester.cpp
spireken: In this case, based on the timestamp B350 is a newer imbalance than B150, and since B350 is published first B150 should be ignored by the table model. If B150 was published first then B350 would replace it, but the opposite isn't true.

It's good you pointed this out, though, because on review I didn't put any condition to prevent an imbalance that is already expired from being inserted into the table model.

shaobinspire: I see.

shaobinspire (approved):
Looks good.

Applications/Spire/Include/Spire/OrderImbalanceIndicator/CurrentOrderImbalanceIndicatorTableModel.hpp
```diff
@@ -0,0 +1,85 @@
...
+
+      int get_column_size() const override;
+
+      const std::any& at(int row, int column) const override;
+
+      boost::signals2::connection connect_operation_signal(
+        const OperationSignal::slot_type& slot) const override;
+
+    private:
+      struct Expiring {
```
kranar, line 51: Types are nouns, not verbs. It's not even clear what could be meant by an object called ```Expiring``` since that's a state.

Applications/Spire/Include/Spire/OrderImbalanceIndicator/CurrentOrderImbalanceIndicatorTableModel.hpp
```diff
@@ -0,0 +1,85 @@
...
+        Nexus::Security m_security;
+        boost::posix_time::ptime m_timestamp;
+      };
+      class CallbackTimer {
+        public:
+          CallbackTimer(std::unique_ptr<Beam::Threading::TimerBox> timer,
```
kranar, line 57: Look at removing the use of ```std::unique_ptr```.

Applications/Spire/Source/OrderImbalanceIndicator/CurrentOrderImbalanceIndicatorTableModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+    m_queue.GetSlot<Timer::Result>([=] (auto result) {
+      if(result == Timer::Result::EXPIRED) { m_expiration_callback(); }
+    }));
+  m_timer->Start();
+}
+
+CurrentOrderImbalanceIndicatorTableModel::CurrentOrderImbalanceIndicatorTableModel(
+    time_duration offset, TimeClientBox clock, TimerFactory timer_factory,
+    std::shared_ptr<OrderImbalanceIndicatorModel> source)
+    : m_source(std::move(source)),
```
kranar, line 29: The ordering is inconsistent here. Order things as you would according to English grammar and use a consistent ordering.

Applications/Spire/Source/OrderImbalanceIndicator/CurrentOrderImbalanceIndicatorTableModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+      m_offset(offset),
+      m_clock(std::move(clock)),
+      m_timer_factory(std::move(timer_factory)) {
+  m_operation_connection = m_table.connect_operation_signal(std::bind_front(
+    &CurrentOrderImbalanceIndicatorTableModel::on_operation, this));
+  auto subscription = m_source->subscribe(
+    TimeInterval::closed(m_clock.GetTime() - m_offset,
+      std::numeric_limits<ptime>::max()), std::bind_front(
```
kranar, line 37: Does ```pos_infin``` work instead of using ```numeric_limits```?

Applications/Spire/Source/OrderImbalanceIndicator/CurrentOrderImbalanceIndicatorTableModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+    &CurrentOrderImbalanceIndicatorTableModel::on_imbalance, this));
+  m_subscription_connection = std::move(subscription.m_connection);
+  m_load = subscription.m_snapshot.then(
+    [=] (auto&& result) { on_load(result.Get()); });
```
kranar, line 41: Why is this being passed as a forwarding reference?

Applications/Spire/Source/OrderImbalanceIndicator/CurrentOrderImbalanceIndicatorTableModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+}
+
+void CurrentOrderImbalanceIndicatorTableModel::update_next_expiring() {
+  if(m_table.get_row_size() == 0) {
+    m_next_expiring = none;
+    return;
+  }
+  auto next_expiration = std::numeric_limits<ptime>::max();
+  auto row = 0;
+  auto time_interval_lower = m_clock.GetTime() - m_offset;
```
kranar, line 69: Just as in English, you don't call it a "balloon red", you call it a "red balloon". Use the same grammar rules when naming variables.

Applications/Spire/Source/OrderImbalanceIndicator/CurrentOrderImbalanceIndicatorTableModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+    }
+    ++i;
+  }
+  if(m_table.get_row_size() == 0) {
+    m_next_expiring = none;
+    return;
+  }
+  m_next_expiring = Expiring{m_table.get<Security>(row, 0),
+    ptime(m_table.get<date>(row, 5), m_table.get<time_duration>(row, 6))};
+  m_timers.emplace_back(std::make_unique<TimerBox>(m_timer_factory(
```
kranar, line 88: Use ```push_back```.

Applications/Spire/Source/OrderImbalanceIndicator/CurrentOrderImbalanceIndicatorTableModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+    (m_next_expiring->m_timestamp + m_offset) - m_clock.GetTime())),
```
kranar, line 89: What do the parenthesis do?

Applications/Spire/Source/OrderImbalanceIndicator/CurrentOrderImbalanceIndicatorTableModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+    update_next_expiring();
+  });
+}
+
+void CurrentOrderImbalanceIndicatorTableModel::on_imbalance(
+    const OrderImbalance& imbalance) {
+  if(imbalance.m_timestamp < m_clock.GetTime() - m_offset) {
+    return;
+  }
+  auto is_update_required =
```
kranar, line 106: Why is this on line 106 instead of on line 111?

kranar (changes requested):

Applications/Spire/Source/OrderImbalanceIndicator/CurrentOrderImbalanceIndicatorTableModel.cpp
spireken: I'd like to avoid copying the result (which could be large), but the value returned by the promise will be a temporary so I can't take it by reference. My understanding of this is that result's lifetime is equal to it's use, so it will persist until the call to on_load has completed.

Applications/Spire/Source/OrderImbalanceIndicator/CurrentOrderImbalanceIndicatorTableModel.cpp
spireken: Are you referring to the ordering of the parameters in the API in contrast to the ordering as they're declared in the class? I'm just not sure what that has to do with grammar.

Applications/Spire/Source/OrderImbalanceIndicator/CurrentOrderImbalanceIndicatorTableModel.cpp
kranar: Follow up on the rules for how references work and then fix this up. As is it's too inconsistent and I'm not sure what the purpose of it is.

Applications/Spire/Source/OrderImbalanceIndicator/CurrentOrderImbalanceIndicatorTableModel.cpp
kranar: The whole thing, use a consistent ordering according to how most English speakers speak and keep that ordering consistent. Ordering matters, it's not arbitrary and it informs things such as dependencies and relationships. For example it's why almost all programming languages structure things as ```subject.verb(noun1, noun2, ...)``` or ```verb(noun1, noun2, ...)```.

kranar (changes requested):

PR #794, opened 2022-05-05
Quantity box update

shaobinspire:


Applications/Spire/Source/Ui/QuantityBox.cpp
```diff
@@ -34,4 +35,50 @@ QuantityBox::QuantityBox(std::shared_ptr<OptionalQuantityModel> model,
...
+}
+
+bool QuantityBox::eventFilter(QObject* watched, QEvent* event) {
+  if(watched == &get_decimal_box() && event->type() == QEvent::Show) {
+    auto proxy = find_focus_proxy(get_decimal_box());
+    if(proxy) {
+      proxy->installEventFilter(this);
+    }
+  } else if(event->type() == QEvent::KeyPress && !is_read_only()) {
+    auto& key_event = *static_cast<QKeyEvent*>(event);
```
kranar, line 48: Why is an event filter used instead of having the ```QKeyEvent``` bubble up?

Applications/Spire/Source/Ui/QuantityBox.cpp
```diff
@@ -34,4 +35,50 @@ QuantityBox::QuantityBox(std::shared_ptr<OptionalQuantityModel> model,
...
+    auto multiplier = 1.0;
```
kranar, line 49: Use an IIFE.

Applications/Spire/Source/Ui/QuantityBox.cpp
```diff
@@ -34,4 +35,50 @@ QuantityBox::QuantityBox(std::shared_ptr<OptionalQuantityModel> model,
...
+    if(key_event.key() == Qt::Key_K) {
+      multiplier = 1000;
+    } else if(key_event.key() == Qt::Key_H) {
+      multiplier = 100;
+    } else if(key_event.key() == Qt::Key_D) {
+      if(key_event.modifiers().testFlag(Qt::AltModifier)) {
+        multiplier = 10;
+      } else {
+        multiplier = 0.1;
```
kranar, line 58: Work with integers when possible. Instead of ```0.1 * x```, rewrite it as ```x / 10```. This avoids introducing any additional floating point issues.

kranar (changes requested):
Looks good.

Applications/Spire/Source/Ui/QuantityBox.cpp
shaobinspire: Since the `QKeyEvent` is set to accepted state by the `QLineEdit` contained in the `TextBox` which is used by the `DecimalBox` when the `QLineEdit` receives alphanumeric input, it can't be propagated to its parent widget according to the description of `QKeyEvent`:

> Calling [ignore](https://doc.qt.io/qt-5/qevent.html#ignore)() on a key event will propagate it to the parent widget. The event is propagated up the parent widget chain until a widget accepts it or an event filter consumes it.

kranar: Sounds good.

kranar (approved):

PR #795, opened 2022-05-10
Market box

shaobinspire:


Applications/Spire/Source/Ui/MarketBox.cpp
```diff
@@ -0,0 +1,110 @@
...
+    auto value = m_source->parse(query);
+    if(value.has_value()) {
+      return to_destination(std::any_cast<MarketDatabase::Entry&>(value));
+    }
+    return value;
+  }
+
+  QtPromise<std::vector<std::any>> submit(const QString& query) override {
+    return m_source->submit(query).then([=] (auto&& source_result) {
+        auto matches = [&] {
```
kranar, line 36: This makes an extra copy.

Applications/Spire/Source/Ui/MarketBox.cpp
```diff
@@ -0,0 +1,110 @@
...
+MarketBox::MarketBox(
+    std::shared_ptr<ComboBox::QueryModel> query_model,
+    std::shared_ptr<CurrentModel> current, QWidget* parent)
+    : QWidget(parent),
+      m_query_model(
+        std::make_shared<MarketQueryModel>(std::move(query_model))),
+      m_current(std::move(current)) {
+  auto destination_box_current = make_transform_value_model(m_current,
+    [=] (const MarketCode& current) {
+      return GetDefaultMarketDatabase().FromCode(current).m_displayName;
```
kranar, line 67: Do not use default databases, they are only for testing.

kranar (changes requested):
I think reusing ```DestinationBox``` was a good idea overall and results in a very elegant solution.

PR #796, opened 2022-05-16
Transition view

shaobinspire:


Applications/Spire/Include/Spire/Ui/TransitionView.hpp
```diff
@@ -0,0 +1,50 @@
+#ifndef SPIRE_TRANSITION_VIEW_HPP
+#define SPIRE_TRANSITION_VIEW_HPP
+#include <QWidget>
```
kranar, line 3: Missing the "Ui.hpp" include.

Applications/Spire/Source/Ui/TransitionView.cpp
```diff
@@ -0,0 +1,82 @@
...
+using namespace Spire::Styles;
+
+TransitionView::TransitionView(QWidget* body, QWidget* parent)
+    : QWidget(parent),
+      m_body(body),
+      m_timer(new QTimer(this)),
+      m_status(Status::NONE) {
+  make_vbox_layout(this);
+  make_pending_widget();
+  m_timer->setSingleShot(true);
```
kranar, line 19: This assumes that a widget is shown on construction, which is almost never the case.

kranar (changes requested):

PR #800, opened 2022-05-31
Security view

shaobinspire:


Applications/Spire/Include/Spire/Ui/SecurityView.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+      /**
+       * Constructs a SecurityView.
+       * @param query_model The model used to query matches.
+       * @param body The component that represents the current security.
+       * @param current The current value's model.
+       * @param parent The parent widget.
+       */
+      SecurityView(std::shared_ptr<ComboBox::QueryModel> query_model,
+        QWidget* body, std::shared_ptr<CurrentModel> current,
```
kranar, line 37: Write your code in a consistent order. The constructor orders things as ```query_model```, ```body```, ```current```, then the methods order things as ```query_model```, ```current```, ```body```, and then the fields order things as ```body```, ```query_model```, ```current```.

Determine what the conceptual order of these fields are and use a consistent ordering.

Applications/Spire/Source/Ui/SecurityView.cpp
```diff
@@ -0,0 +1,165 @@
...
+    m_current->set(m_securities[m_current_index]);
+  } else if(event->key() == Qt::Key_PageDown && !m_securities.empty()) {
+    m_current_index = (m_current_index + 1) % m_securities.size();
+    m_current->set(m_securities[m_current_index]);
+  }
+  QWidget::keyPressEvent(event);
+}
+
+void SecurityView::on_submit(const Security& security) {
+  if(auto iter = std::find(m_securities.begin(), m_securities.end(), security);
```
kranar, line 151: ```iter``` -> ```i```.

Applications/Spire/Source/Ui/TextBox.cpp
```diff
@@ -513,6 +513,7 @@ void TextBox::resizeEvent(QResizeEvent* event) {
 void TextBox::showEvent(QShowEvent* event) {
   if(!m_line_edit && (!is_read_only() || isEnabled())) {
     initialize_line_edit();
+    m_line_edit->show();
```
kranar, line 516: Provide a brief explanation for what this fixes.

Applications/Spire/Source/Ui/OverlayPanel.cpp
```diff
@@ -217,8 +217,10 @@ void OverlayPanel::on_parent_move(QPoint) {
 }
 
 void OverlayPanel::update_mask() {
-  auto intersection = geometry().intersected(
-    QRect(parentWidget()->mapToGlobal(QPoint()), parentWidget()->size()));
-  setMask(QPolygon(rect()).subtracted(
-    QRect(mapFromGlobal(intersection.topLeft()), intersection.size())));
+  if(m_positioning == Positioning::PARENT) {
```
kranar, line 220: Briefly explain what this fixes.

Applications/Spire/Source/Ui/ComboBox.cpp
```diff
@@ -125,7 +125,6 @@ bool ComboBox::eventFilter(QObject* watched, QEvent* event) {
         } else {
           revert_to(m_submission_text, false);
         }
-        return true;
```
kranar, line 128: Briefly explain what this fixes.

Applications/Spire/Include/Spire/Ui/SecurityView.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * Displays securities and provides the functionality to quickly search and
+   * cycle through securities.
+   */
+  class SecurityView : public QWidget {
+    public:
+
+      /** A ValueModel over a Nexus::Security. */
+      using CurrentModel = ValueModel<Nexus::Security>;
```
kranar, line 18: ```ValueModel``` isn't included as a header.

Applications/Spire/Include/Spire/Ui/SecurityView.hpp
```diff
@@ -0,0 +1,69 @@
...
+      void keyPressEvent(QKeyEvent* event) override;
+
+    private:
+      class SecuritySearchWindow;
+      QWidget* m_body;
+      TextBox* m_prompt;
+      SecuritySearchWindow* m_search_window;
+      QStackedWidget* m_layers;
+      std::shared_ptr<CurrentModel> m_current;
+      std::vector<Nexus::Security> m_securities;
```
kranar, line 62: I don't see any include for ```vector```.

kranar (changes requested):
This is excellent work, very impressive.

Applications/Spire/Source/Ui/OverlayPanel.cpp
shaobinspire: The purpose of the function `update_mask` is to mask the overlapping region between the `OverlayPanel` and its parent in order to reveal the border style of parent component when `Positioning` is `Positioning::PARENT`. To be precise, the area is the shared borders between the `OverlayPanel` and its parent. However, if `Positioning` is `Positioning::NONE`, the `OverlayPanel` might be completely covered by its parent when it is placed inside the area of the parent. Hence, this functionality of `update_mask` is revised to limit to `Positioning = PARENT`.

Applications/Spire/Source/Ui/ComboBox.cpp
shaobinspire: When the `ComboBox` is placed in the `OverlayPanel`, the Escape event is prevented from being sent to the `OverlayPanel` because of returning true after handling the Escape event in `ComboBox::eventFilter`, so the `OverlayPanel` can’t receive the Escape event to close itself.

kranar: What alternative approach can you think of to resolve this issue?

Applications/Spire/Source/Ui/OverlayPanel.cpp
kranar: Approved.

Applications/Spire/Source/Ui/ComboBox.cpp
shaobinspire: An alternative is that the `SecurityBox` listens to the Escape key press of the `TextBox` through `eventFilter`, when the `SecurityBox` receives the event, it resends the event to the `OverlayPanel`. It’s feasible because the `SecurityBox` receives the press event prior to the `ComboBox`.

Applications/Spire/Source/Ui/TextBox.cpp
shaobinspire: This change is unnecessary. I will restore it.

Applications/Spire/Include/Spire/Ui/SecurityView.hpp
kranar: The ordering is still not consistent, the constructor is now ```query_model```, ```body```, ```current``` and the methods are ```query_model```, ```current```, ```body```. These subtle details add up in the long run. Think about conceptually what the order of dependencies are and order things from the most independent to the least independent. What is more likely, that ```current``` depends on ```query_model``` or that ```query_model``` depends on ```current```? Is it more likely that ```body``` depends on ```current``` or that ```current``` depends on ```body```?

Once you know what the most common logical order of dependencies are, arrange everything in order from independent to dependent, constructor, methods, fields, algorithms, anything and everything.

kranar (commented):
Changes look good.

PR #807, opened 2022-07-07
Popup box

shaobinspire:


Applications/Spire/Include/Spire/Ui/PopupBox.hpp
```diff
@@ -0,0 +1,69 @@
...
+      int m_above_space;
+      int m_below_space;
+      int m_right_space;
+      mutable QSize m_size_hint;
+      boost::signals2::scoped_connection m_focus_connection;
+      boost::signals2::scoped_connection m_body_focus_connection;
+      boost::signals2::scoped_connection m_position_connection;
+
+      bool has_popped_up() const;
+      void on_focus(FocusObserver::State state);
```
kranar, line 59: Place the event handlers at the end. All the ```on_xxx``` go to the end.

Applications/Spire/Include/Spire/Ui/TagBox.hpp
```diff
@@ -84,16 +85,18 @@ namespace Styles {
...
+      int m_input_box_horizontal_padding;
       int m_list_view_horizontal_padding;
       int m_scroll_bar_end_range;
+      boost::signals2::scoped_connection m_focus_connection;
       boost::signals2::scoped_connection m_style_connection;
       boost::signals2::scoped_connection m_list_view_style_connection;
       boost::signals2::scoped_connection m_text_box_style_connection;
 
       QWidget* make_tag(const std::shared_ptr<AnyListModel>& model, int index);
+      void on_focus(FocusObserver::State state);
```
kranar, line 99: See above about ```on_xxx```.

Applications/Spire/Include/Spire/Ui/TagComboBox.hpp
```diff
@@ -108,10 +98,8 @@ namespace Spire {
       void on_operation(const AnyListModel::Operation& operation);
```
kranar, line 98: See above.

Applications/Spire/Source/Scratch/main.cpp
```diff
@@ -1,12 +1,105 @@
 #include <QApplication>
```
kranar, line 1: Revert changes to this and incorporate this demo to the best that you can into the Ui Viewer. The UI Viewer doesn't need to have an exact copy of this demo but it should be good enough to test out most functionality.

Applications/Spire/Source/Ui/PopupBox.cpp
```diff
@@ -0,0 +1,182 @@
...
+      m_focus_observer(*this),
+      m_body_focus_observer(*m_body),
+      m_position_observer(*this),
+      m_alignment(Alignment::NONE),
+      m_min_height(m_body->sizeHint().height()),
+      m_max_height(QWIDGETSIZE_MAX),
+      m_above_space(0),
+      m_below_space(0),
+      m_right_space(0),
+      m_focus_connection(m_focus_observer.connect_state_signal(
```
kranar, line 21: There is no need for an object to establish a connection to another object that is owned. Since ```PopupBox``` owns ```m_focus_observer``` then no connection is needed since the ```m_focus_observer``` will be destroyed before the ```PopupBox``` is destroyed.

Applications/Spire/Source/Ui/PopupBox.cpp
```diff
@@ -0,0 +1,182 @@
...
+const QWidget& PopupBox::get_body() const {
+  return *m_body;
+}
+
+QWidget& PopupBox::get_body() {
+  return *m_body;
+}
+
+QSize PopupBox::sizeHint() const {
+  if(has_popped_up()) {
```
kranar, line 41: Can you describe the goal of this method? It looks like it could be rewritten as:

    if(has_popped_up()) {
      return m_body->sizeHint();
    }
    return QWidget::sizeHint();

Applications/Spire/Source/Ui/PopupBox.cpp
```diff
@@ -0,0 +1,182 @@
...
+  } else if(minimumHeight() == maximumHeight()) {
+    m_body->setFixedHeight(minimumHeight());
+    if(m_body->sizeHint().height() > height()) {
+      m_body->setFixedWidth(std::max(width(), minimumWidth()));
+    } else {
+      m_body->setMinimumWidth(std::max(width(), minimumWidth()));
+      m_body->setMaximumWidth(std::min(m_right_space, maximumWidth()));
+    }
+  } else {
+    m_body->setMinimumHeight(std::max(height(), minimumHeight()));
```
kranar, line 157: How can ```height()``` be less than ```minimumHeight()```. By Qt's definition the following is always true: ```height() >= minimumHeight()```.

Applications/Spire/Source/Ui/ScrollBox.cpp
```diff
@@ -460,19 +460,9 @@ void ScrollBox::update_layout() {
   auto scroll_bar_size = QSize(
     m_scrollable_layer->get_vertical_scroll_bar().sizeHint().width(),
     m_scrollable_layer->get_horizontal_scroll_bar().sizeHint().height());
-  auto viewport_size = [&] {
```
kranar, line 463: What is this change related to? Is there a task for it?

shaobinspire: The purpose of this code is to solve the issue of the vertical `ScrollBar` blinking when adding a new Tag. The `TagBox` has to access the view port of the `ScrollBox` and set maximum size to it. In addition, Setting the maximum size of the view port might cause the `ScrollBox` placed in a layout sometimes not to fill the entire layout (The `ScrollBox` is partially displayed in its parent).  Now I fixed the blinking by setting the vertical display policy (ON_OVERFLOW and NEVER) within the `TagBox`, so the fix was processed in the `TagBox` instead of in the `ScrollBox`, which doesn't need to hack into the `ScrollBox`. Hence, I removed these code and reverted to the original implementation of the ScrollBox.

Applications/Spire/Source/Ui/PopupBox.cpp
shaobinspire: The goal of the method is to keep the PopupBox the same size as before the body popping up. When the body poppes out from the layout of the PopupBox, the size of PopupBox will become zero, so the method prevents it from becoming zero. I don’t think the rewritten code is same as it. 
 m_body->sizeHint() can return different sizes as the content changes, while the method needs to return a constant size which is the size before the body popping up. This constant size is stored in m_size_hint.

kranar (changes requested):

Applications/Spire/Include/Spire/Ui/PopupBox.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+    private:
+      enum class Alignment : std::uint8_t {
+        NONE,
+        ABOVE,
+        BELOW
+      };
+      QWidget* m_body;
+      QWidget* m_window;
+      FocusObserver m_focus_observer;
```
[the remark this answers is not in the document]
shaobinspire: `m_focus_observer` is used to observe getting focus of the `PopupBox`, and `m_body_focus_observer` is used to observe loss of focus of the body. When the body pops out from the `PopupBox`, it no longer belongs to the `PopupBox` (it has re-parented to the window), so there is no way to get the focus state of body from the `PopupBox`, and we need to observe the body and the `PopupBox` separately.

shaobinspire:
Since this submission involves the old TagBox and TagComboBox which have been updated now, I will close it and submit a new branch. All issues related to the PopupBox which were found in this submission will be fixed in the new one.

PR #809, opened 2022-07-22
Added padding to the InfoTip.

shaobinspire:


Applications/Spire/Source/Ui/InfoTip.cpp
```diff
@@ -235,7 +235,10 @@ QPixmap InfoTip::render_background() const {
   shadow.setOffset(0, scale_height(3));
   shadow.setBlurRadius(scale_width(5));
   auto path = get_arrow_path();
-  path.addRect(rect().marginsRemoved(get_margins()));
+  auto half_border_size = m_border_size / 2.0;
+  auto border_margins = QMargins(half_border_size, half_border_size,
+    std::round(half_border_size), std::round(half_border_size));
```
kranar, line 240: Why is round used instead of ```ceil``` or ```floor```?

Applications/Spire/Source/Ui/InfoTip.cpp
```diff
@@ -282,10 +286,31 @@ void InfoTip::on_style() {
         stylist.evaluate(size, [=] (auto size) {
           m_border_size = size;
         });
+      },
+      [&] (const PaddingRight& size) {
+        stylist.evaluate(size, [&] (auto size) {
```
kranar, line 291: Capture by reference only when the closure does not outlive its lexical scope.

kranar (changes requested):

kranar (approved):

PR #822, opened 2022-09-12
Region box display initial current

shaobinspire:


Applications/Spire/Source/Ui/ListCurrentController.cpp
```diff
@@ -69,11 +69,14 @@ void ListCurrentController::move(int source, int destination) {
     return 1;
   }();
   if(direction == 1) {
-    std::rotate(m_views.begin() + source, m_views.begin() + source + 1,
-      m_views.begin() + destination + 1);
+    std::rotate(m_views.begin() + destination, m_views.begin() + source,
```
kranar, line 72: What does this have to do with the task?

Applications/Spire/Source/Ui/RegionBox.cpp
```diff
@@ -125,6 +173,19 @@ void RegionBox::resizeEvent(QResizeEvent* event) {
   QWidget::resizeEvent(event);
 }
 
+void RegionBox::region_to_list_of_tags(const Region& region,
```
kranar, line 176: Why is this a method?

Applications/Spire/Source/Ui/TagComboBox.cpp
```diff
@@ -25,10 +25,23 @@ namespace {
   void copy_list_model(const std::shared_ptr<AnyListModel>& from,
       const std::shared_ptr<AnyListModel>& to) {
     to->transact([&] {
+      auto find = [=] (int index) {
```
kranar, line 28: Why is this capturing by value?

Applications/Spire/Include/Spire/Ui/RegionBox.hpp
```diff
@@ -70,7 +70,10 @@ namespace Spire {
       std::shared_ptr<CurrentModel> m_current;
       TagComboBox* m_tag_combo_box;
       boost::signals2::scoped_connection m_current_connection;
+      boost::signals2::scoped_connection m_tag_operation_connection;
 
+      void region_to_list_of_tags(const Nexus::Region& region,
```
kranar, line 75: Rename to ```to_tag_list```.

kranar (changes requested):
Good start.

Applications/Spire/Source/Ui/ListCurrentController.cpp
shaobinspire: When the regions inside the `RegionBox` are sorted,  the `ListItem` containing the region Tags will be moved, and this function is to move `ListItems` from the `source` position to the `destination` position. `std::rotate` was used in the wrong way, so I have to try to fix it.

> template< class ForwardIt >
> void rotate( ForwardIt first, ForwardIt n_first, ForwardIt last );
>
> It accepts three parameters: 
> first       -the beginning of the original range
> n_first   -the element that should appear at the beginning of the rotated range
> last        -the end of the original range

If `source` > `destination`, `first` is `destination`, `n_first` is `source` and `last` is `source + 1`, which means to move the value at the `source` position to the beginning of the range `[destination, source + 1)`;
if `source` < `destination`, it moves the value at the `source` position to the end of the range `(source -1, destination]`,  using a reverse iterator.

Applications/Spire/Source/Ui/RegionBox.cpp
shaobinspire: Because m_query_model is used inside the function. The function could become a free function by passing the query model in it.

Applications/Spire/Source/Ui/ListCurrentController.cpp
shaobinspire: There are similar snippets of moving items inside `ListView` (`ListView::move_item`, line 394 and line 405 in ListView.cpp). I will update the code here to be consistent with the snippets inside `ListView::move_item`.

kranar (approved):
Looks very good.

PR #823, opened 2022-09-16
The overflow of the TagBox in a wrapper

shaobinspire:
Instead of setting fixed size from parent to child, for example, the `TagComboBox` or `RegionBox` sets fixed size to the `TagBox`, the new approach is to get fixed size from the `TagBox’s` wrapper (I think the fixed size of the root wrapper would be the size of the `TagBox`). The `TagBox’s` wrapper is the component that wraps one and only one `TagBox` and provides additional functionality to the `TagBox`. The `TagBox` traverses up its parent hierarchy and looks for the root wrapper.


kranar (approved):
Excellent work.

PR #829, opened 2022-10-18
Enter event of the DropDownBox

shaobinspire:


Applications/Spire/Source/Ui/DropDownBox.cpp
```diff
@@ -163,6 +163,8 @@ bool DropDownBox::eventFilter(QObject* watched, QEvent* event) {
         if(!is_read_only()) {
           m_is_modified = true;
           submit();
+        } else {
```
kranar, line 166: Prefer to swap out negations when possible.

    if(is_read_only()) {
      event->ignore();
    } else {
      ...
    }

It's easier to reason in terms of ```X``` instead of ```!X```.

kranar (changes requested):

kranar (approved):

PR #831, opened 2022-10-18
Prevent the ComboBox from swallowing Escape.

shaobinspire:


Applications/Spire/Source/Ui/ComboBox.cpp
```diff
@@ -126,6 +126,7 @@ bool ComboBox::eventFilter(QObject* watched, QEvent* event) {
         } else {
           revert_to(m_submission_text, false);
         }
+        event->ignore();
```
kranar, line 129: Why is this event always ignored? It should only be ignored if the ```ComboBox``` takes no action on the key press.

shaobinspire: Yes, if the ComboBox takes some actions, the event should not be ignored.

kranar (approved):

PR #832, opened 2022-10-18
Tag combo box key event

shaobinspire:


Applications/Spire/Source/Ui/TagComboBox.cpp
```diff
@@ -226,6 +227,17 @@ bool TagComboBox::eventFilter(QObject* watched, QEvent* event) {
   return QWidget::eventFilter(watched, event);
 }
 
+void TagComboBox::keyPressEvent(QKeyEvent* event) {
+  if(event->key() == Qt::Key_Escape) {
+    m_tag_box->get_current()->set("");
+    copy_list_model(m_submission, get_current());
+    m_is_modified = false;
+    event->ignore();
```
kranar, line 235: Always ignoring this event is likely incorrect. An event should only be ignored if the widget takes no action/doesn't handle the event.

PR #834, opened 2022-10-20
The issue of the TagBox listening to QLineEdit

shaobinspire:


Applications/Spire/Source/Ui/TagBox.cpp
```diff
@@ -374,6 +377,14 @@ QWidget* TagBox::make_tag(
   return tag;
 }
 
+void TagBox::install_text_proxy_event_filter() {
+  if(!m_text_focus_proxy) {
+    if(m_text_focus_proxy = m_text_box->focusProxy()) {
```
kranar, line 382: Tests of the form ```x = y``` are error prone. Do not use assignment as a condition.

PR #840, opened 2022-11-14
OrderTaskTableModel

shaobinspire:
The OrderTaskTableModel is used by the TableView inside the OrderTasksPage. The first column has an empty value for the field of the grab handle and the last row is the empty row used to add a new row.


Applications/Spire/Include/Spire/KeyBindings/OrderTaskTableModel.hpp
```diff
@@ -0,0 +1,128 @@
...
+#include "Nexus/Definitions/Side.hpp"
+#include "Nexus/Definitions/TimeInForce.hpp"
+#include "Spire/Spire/ListModel.hpp"
+#include "Spire/Spire/TableModel.hpp"
+#include "Spire/Spire/TableModelTransactionLog.hpp"
+
+namespace Spire {
+
+  /** The index of the grab handle field. */
+  static const auto GRAB_HANDLE_INDEX = 0;
```
kranar, line 17: Why are these ```static```?
Why are they part of the ```Spire``` namespace?

Applications/Spire/Include/Spire/KeyBindings/OrderTaskTableModel.hpp
```diff
@@ -0,0 +1,128 @@
...
+
+        /** The key binding of the order task. */
+        QKeySequence m_key_sequence;
+      };
+
+      /**
+       * Constructs an OrderTasksTableModel.
+       * @param source A list of OrderTask.
+       */
+      explicit OrderTaskTableModel(
```
kranar, line 81: This API looks inconsistent. Either this class converts a list to a table, or this class represents a table. Pick one of those two functions and update the API to fulfill one and only one purpose.

Applications/Spire/Source/KeyBindings/OrderTaskTableModel.cpp
```diff
@@ -0,0 +1,174 @@
...
+  return QValidator::State::Invalid;
+}
+
+connection OrderTaskTableModel::connect_operation_signal(
+    const OperationSignal::slot_type& slot) const {
+  return m_transaction.connect_operation_signal(slot);
+}
+
+void OrderTaskTableModel::on_operation(
+    const AnyListModel::Operation& operation) {
```
kranar, line 130: Why does ```on_operation``` operate on the type erased ```AnyListModel::Operation```?

Applications/Spire/Source/KeyBindings/OrderTasksPage.cpp
```diff
@@ -0,0 +1,24 @@
+#include "Spire/KeyBindings/OrderTasksPage.hpp"
+
+using namespace Nexus;
+using namespace Spire;
+
+OrderTasksPage::OrderTasksPage(
```
kranar, line 6: This class does not look like it's been implemented, remove it from the pull request.

kranar (changes requested):
Looks like a good start.

Applications/Spire/Include/Spire/KeyBindings/OrderTasksToTableModel.hpp
```diff
@@ -0,0 +1,111 @@
...
+#include "Nexus/Definitions/Side.hpp"
+#include "Nexus/Definitions/TimeInForce.hpp"
+#include "Spire/Spire/ListModel.hpp"
+#include "Spire/Spire/TableModel.hpp"
+#include "Spire/Spire/TableModelTransactionLog.hpp"
+
+namespace Spire {
+
+  /** The order task with a keybinding. */
+  struct OrderTask {
```
kranar, line 17: Put this in its own file.

Applications/Spire/Source/KeyBindings/OrderTasksToTableModel.cpp
```diff
@@ -0,0 +1,130 @@
...
+    return order_task.m_key;
+  }
+  return {};
+}
+
+void OrderTasksToTableModel::on_operation(
+    const ListModel<OrderTask>::Operation& operation) {
+  visit(operation,
+    [&] (const ListModel<OrderTask>::AddOperation& operation) {
+      m_transaction.push(TableModel::AddOperation(operation.m_index,
```
kranar, line 110: Fix indentation.

Applications/Spire/Source/KeyBindings/OrderTasksTableViewModel.cpp
```diff
@@ -0,0 +1,87 @@
...
+  }
+  if(row < m_source->get_row_size()) {
+    return m_source->at(row, column - 1);
+  }
+  return empty_value;
+}
+
+QValidator::State OrderTasksTableViewModel::set(int row, int column,
+    const std::any& value) {
+  if(row < 0 || row >= get_row_size() || column < 0 ||
```
kranar, line 51: I don't believe these checks are needed, can you verify? Only perform checks if they are needed to satisfy the semantics of this class, there is no need to duplicate error handling that would need to be performed by the source model.

kranar (changes requested):
Big improvement.

Applications/Spire/Source/KeyBindings/OrderTasksTableViewModel.cpp
shaobinspire: Good points. I think checks for `row` might be needed and checks for `column` could be removed. If there are no checks for `row`, and `row` is -1 and `column` is 0, the function of set will return `Invalid`, but the function is supposed to throw an exception since the function is trying to access the row which is out of range. Validating `column` can be done by the source model.

Applications/Spire/Source/KeyBindings/OrderTasksTableViewModel.cpp
```diff
@@ -32,9 +32,8 @@ int OrderTasksTableViewModel::get_column_size() const {
 }
 
 AnyRef OrderTasksTableViewModel::at(int row, int column) const {
-  if(row < 0 || row >= get_row_size() || column < 0 ||
-      column >= get_column_size()) {
-    throw std::out_of_range("The row or column is out of range.");
+  if(row < 0 || row >= get_row_size()) {
+    throw std::out_of_range("The row is out of range.");
   }
   static auto empty_value = AnyRef();
```
kranar, line 38: Remove the use of this static variable, it inhibits copy elision:

https://en.cppreference.com/w/cpp/language/copy_elision

kranar (changes requested):

PR #844, opened 2022-12-02
GrabHandle

shaobinspire:


Applications/Spire/Source/KeyBindings/GrabHandle.cpp
```diff
@@ -0,0 +1,49 @@
...
+      set(BackgroundColor(QColor(Qt::transparent))).
+      set(Fill(QColor(0x535353)));
+  });
+  m_hover_observer.connect_state_signal(
+    std::bind_front(&GrabHandle::on_hover, this));
+}
+
+void GrabHandle::mousePressEvent(QMouseEvent* event) {
+  setCursor(Qt::ClosedHandCursor);
+  QWidget::mousePressEvent(event);
```
kranar, line 35: Why does this call the parent ```mousePressEvent```?

shaobinspire: I think the `GrabHandle` just do an update for the cursor shape and it should not prevent this mouse event bubbling up to its parent. The the `accept` flag of this event got by this handler is accepted, and the parent `mousePressEvent` will ignore this flag, so `GrabHandle`'s parent still has a chance to do some extra actions.

PR #849, opened 2022-12-15
Destination box escape

shaobinspire:


kranar:
Only merge into the ```key_bindings``` branch issues that are related to the key bindings task.

PR #855, opened 2022-12-19
TableCurrentController and TableSelectionController calucate wrong row index

shaobinspire:
It fixed the issue of [TableView crashes when it is sorting and the current row is the last row.](https://app.asana.com/0/1169210941620249/1203563868277689/f)


kranar (approved):
Excellent fix!

PR #866, opened 2023-01-23
The scrollbar scrolling to the top when a new row is added

shaobinspire:
It is related to the task [The scrollbar scrolling to the top when a new row is added](https://app.asana.com/0/0/1203756802316159/f).

Showing the newly added row can recalculate the correct row geometry which will be used to calculate the scrollbar position, before entering the Qt event loop.

### DIFF
--- a/Applications/Spire/Include/Spire/KeyBindings/OrderTasksRow.hpp
+++ b/Applications/Spire/Include/Spire/KeyBindings/OrderTasksRow.hpp
@@ -77,6 +77,9 @@ namespace Styles {
       /** Returns the row index. */
       int get_row_index() const;
 
+      /** Returns the row widget. */
+      QWidget* get_row() const;
+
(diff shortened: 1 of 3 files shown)


Applications/Spire/Include/Spire/KeyBindings/OrderTasksRow.hpp
```diff
@@ -77,6 +77,9 @@ namespace Styles {
       /** Returns the row index. */
       int get_row_index() const;
 
+      /** Returns the row widget. */
+      QWidget* get_row() const;
```
kranar, line 81: Can the return value be a ```nullptr```?

shaobinspire: Yes, it can.

PR #868, opened 2023-01-25
Order tasks row drag and drop

shaobinspire:
It resolved the task of [Drag and Drop table row for the KeyBindings](https://app.asana.com/0/0/1203649305772855/f).

### DIFF
--- a/Applications/Spire/Source/KeyBindings/OrderTasksPage.cpp
+++ b/Applications/Spire/Source/KeyBindings/OrderTasksPage.cpp
@@ -47,6 +47,8 @@ namespace {
     style.get(Any() > CurrentRow()).set(BackgroundColor(QColor(0x88E2E0FF)));
     style.get(Any() > CurrentColumn()).set(BackgroundColor(Qt::transparent));
     style.get(Any() > OutOfRangeRow()).set(BackgroundColor(QColor(0xFDF8DE)));
+    style.get(Any() > (DraggingCurrent() && Current())).
+      set(border_color(QColor(0xE0E0E0)));
     return style;
   }
 
@@ -269,7 +271,8 @@ OrderTasksPage::OrderTasksPage(
       m_region_query_model(std::move(region_query_model)),
       m_model(std::move(model)),
       m_destinations(std::move(destinations)),
-      m_markets(std::move(markets)) {
+      m_markets(std::move(markets)),
+      m_rows(std::make_shared<ArrayListModel<std::shared_ptr<TableRow>>>()) {
   auto layout = make_vbox_layout(this);
   layout->setContentsMargins({0, 0, 0, scale_height(20)});
   layout->addWidget(make_help_text_region());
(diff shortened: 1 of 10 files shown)


Applications/Spire/Source/KeyBindings/OrderTasksPage.cpp
```diff
@@ -141,14 +143,15 @@ namespace {
 
   auto display(const Region& region) {
     auto text = QString();
+    auto delimiter = QString(0x1f);
```
kranar, line 146: Why is a delimiter used to display text?

Applications/Spire/Source/KeyBindings/OrderTasksRow.cpp
```diff
@@ -469,10 +475,8 @@ EditableBox* OrderTasksRow::make_editor(
     return new EditableBox(*input_box);;
 }
 
-EditableBox* OrderTasksRow::make_empty_editor(
-    const std::shared_ptr<ComboBox::QueryModel>& region_query_model,
-    const DestinationDatabase& destinations, const MarketDatabase& markets,
-    const std::shared_ptr<TableModel>& table, int row, int column) {
+EditableBox* OrderTasksRow::make_empty_editor(int column,
```
kranar, line 478: What mutable operation does this perform?

Applications/Spire/Source/KeyBindings/TableRowDragDrop.cpp
```diff
@@ -0,0 +1,385 @@
...
+          item.m_order != TableHeaderItem::Order::NONE) {
+        item.m_order = TableHeaderItem::Order::NONE;
+        header_items->set(i, item);
+      }
+    }
+    m_is_sorted = false;
+  }
+}
+
+void TableRowDragDrop::drag_start(const QPoint& pos) {
```
kranar, line 334: Rename to ```start_drag```. Prefer naming functions after verbs.

kranar (changes requested):

Applications/Spire/Source/KeyBindings/OrderTasksPage.cpp
shaobinspire: The purpose of creating the function is to generate a string of the region for searching. It can be useful to use delimiters in searching.
For example, suppose a region includes "USA" and "AUS", if no delimiter is used, the string of the region might be created as "USAAUS". The user types "AA" in the search box, the region will meet the filter, but actually it shouldn't.

kranar: A function named ```display``` shall only be used to display a text representation of a value. If you're intention is to perform a search then you will need a function that serves that specific purpose.

Applications/Spire/Source/KeyBindings/OrderTasksRow.cpp
shaobinspire: `on_submit` which is a mutable operation is used for the submit signal of the input box within `make_empty_editor`.   Since 'make_editor' doesn't preform any mutable operation, it should be a const member function.

PR #870, opened 2023-01-26
TableView crash on delete the last row when current_row is last

shaobinspire:


Applications/Spire/Source/Ui/TableBody.cpp
```diff
@@ -392,15 +392,29 @@ TableItem* TableBody::get_current_item() {
 }
 
 TableBody::Cover* TableBody::find_row(int index) {
-  return static_cast<Cover*>(layout()->itemAt(index)->widget());
+  if(auto row = layout()->itemAt(index)) {
```
kranar, line 395: Why is this performing a check? Try to identify the root cause of the issue and fix it at the point of use.

kranar (approved):

PR #874, opened 2023-02-08
Set the size of left and right borders of the TableBody to be 0

shaobinspire:


Applications/Spire/Source/KeyBindings/OrderTasksPage.cpp
```diff
@@ -39,6 +39,9 @@ namespace {
       set(BorderBottomSize(scale_height(1))).
       set(BorderTopColor(QColor(0xE0E0E0))).
       set(BorderTopSize(scale_height(1)));
+    style.get(Any() > is_a<TableBody>()).
```
kranar, line 42: Use ```HorizontalPadding```.

kranar (approved):

PR #883, opened 2023-03-02
The TagBox jitters when inserting a new tag

shaobinspire:


Applications/Spire/Source/Ui/TagBox.cpp
```diff
@@ -412,13 +441,21 @@ void TagBox::update_tag_size_policy() {
   if(m_model->m_source->get_size() == 0) {
     m_list_view->set_direction_size_policy(QSizePolicy::Expanding);
   } else {
+    auto set_tag_text_size_policy = [&] (QSizePolicy::Policy horizontal) {
+      for(auto i = 0; i < m_list_view->get_list()->get_size() - 1; ++i) {
+        static_cast<Box*>(m_list_view->get_list_item(i)->get_body().layout()->
```
kranar, line 446: This looks good in terms of correctness, but break down the expression into sub-expressions to make clear what the sub-components/widgets are.

Applications/Spire/Source/Ui/TagBox.cpp
```diff
@@ -519,17 +558,31 @@ void TagBox::update_vertical_scroll_bar_visible() {
...
-      m_scrollable_list_box->get_scroll_box().set_vertical(
-        ScrollBox::DisplayPolicy::NEVER);
-    }
+    m_scrollable_list_box->get_scroll_box().set_vertical(
+      ScrollBox::DisplayPolicy::NEVER);
+  }
+}
+
+void TagBox::update_size_constraint() {
+  auto list_view_body =
```
kranar, line 570: For non-null references, use a reference instead of a pointer.

PR #886, opened 2023-03-15
The selection controller updates index.

shaobinspire:
This pull request is part of fixing a crash when out-of-range rows are removed.


kranar (approved):
Much appreciated.

PR #892, opened 2023-03-22
Key bindings sorting reorder row

shaobinspire:


shaobinspire:
This approach causes other issues (https://app.asana.com/0/0/1204204934146706/1204252179031991/f).

PR #908, opened 2023-05-15
Qt warnings

shaobinspire:
Qt outputs a warning of “QWidget::setMinimumSize: (/QWidget) Negative sizes (-13, 552) are not possible”, when showing/hiding the table columns. The reason is that the column minimum width is set to a negative value when the width of the window is less than the total width of all columns.

Although Qt can correct the minimum width to be a non-negative value, I think it should be handled before Qt deals with it.


kranar (approved):

kranar:
Thanks for catching this issue and fixing it.

PR #914, opened 2023-07-04
Added the boost namespace to disambiguate the function of get.

shaobinspire:
When I compiled Spire2018 recently, I got an error of "Spire/SpireTester/ListModelTester.hpp(37,31): error C2275: 'Spire::ListModel<int>::AddOperation': expected an expression instead of a type". When accessing the ListModel::Operation which is a `boost::variant`, the get function without a namespace may confuse the compiler.

### DIFF
--- a/Applications/Spire/Include/Spire/SpireTester/ListModelTester.hpp
+++ b/Applications/Spire/Include/Spire/SpireTester/ListModelTester.hpp
@@ -5,8 +5,6 @@
 #include <doctest/doctest.h>
 #include "Spire/Spire/ListModel.hpp"
 
-using namespace boost;
-
 namespace Spire {
 
   /**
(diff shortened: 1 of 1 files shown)


Applications/Spire/Include/Spire/SpireTester/ListModelTester.hpp
```diff
@@ -5,6 +5,8 @@
 #include <doctest/doctest.h>
 #include "Spire/Spire/ListModel.hpp"
 
+using namespace boost;
```
kranar, line 8: Don't add ```using namespace``` in header files.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-using-directive

PR #917, opened 2023-07-18
Fixed the issue that the DropdownBox current list item not being styled on initialization.

shaobinspire:


kranar (approved):
I am very impressed with your ability to catch these issues. Good job.

PR #931, opened 2023-10-10
ColorPicker

shaobinspire:


Applications/Spire/Source/Ui/ColorPicker.cpp
```diff
@@ -0,0 +1,336 @@
...
+#include "Spire/Ui/OverlayPanel.hpp"
+#include "Spire/Ui/Slider2D.hpp"
+
+using namespace boost;
+using namespace boost::signals2;
+using namespace Spire;
+using namespace Spire::Styles;
+
+namespace {
+  auto get_board_image() {
```
kranar, line 20: Why doesn't this return a constant reference?

Applications/Spire/Source/Ui/ColorPicker.cpp
```diff
@@ -0,0 +1,336 @@
...
+  scoped_connection m_spectrum_x_connection;
+  scoped_connection m_spectrum_y_connection;
+
+  ColorPickerModel(std::shared_ptr<ValueModel<QColor>> source)
+      : m_source(std::move(source)),
+        m_hue_slider_model(std::make_shared<LocalDecimalModel>()),
+        m_alpha_slider_model(std::make_shared<LocalDecimalModel>()),
+        m_spectrum_x_model(std::make_shared<LocalDecimalModel>()),
+        m_spectrum_y_model(std::make_shared<LocalDecimalModel>()) {
+    m_hue_slider_model->set_minimum(Decimal(0));
```
kranar, line 164: Does it fail if you omit the ```Decimal```?

    m_hue_slider_model->set_minimum(0);
    m_hue_slider_model->set_maximum(360);

Applications/Spire/Include/Spire/Ui/ColorPicker.hpp
```diff
@@ -0,0 +1,70 @@
...
+       * @param parent The parent widget.
+       */
+      explicit ColorPicker(QWidget& parent);
+
+      /**
+       * Constructs a ColorPicker with a predefined set of colors.
+       * @param current The model used for the current color.
+       * @param parent The parent widget.
+       */
+      ColorPicker(std::shared_ptr<ValueModel<QColor>> current,
```
kranar, line 30: Are there typedefs for these models?

    ColorValueModel
    LocalColorValueModel
    ...

kranar (changes requested):
Good start.

Applications/Spire/Source/Ui/ColorPicker.cpp
shaobinspire: Yes, because `set_minimum` and `set_maximum` of those models accept `boost::optional` as argument.

Applications/Spire/Include/Spire/Ui/ColorPicker.hpp
shaobinspire: There are no these typedefs here. These typedefs were added to the `ColorBox.hpp`, and `ColorBox` are the parent of `ColorPicker`. Maybe we can put these typedefs to a separate file, like the typedefs for decimal models in `Decimal.hpp`. In this case, the `ColorBox`, `ColorPicker`, `ColorCodePanel` and even `HexColorBox` can use these typedefs.

kranar: Feel free to include ```ColorBox.hpp```

shaobinspire: `ColorBox.hpp` does not exist yet. The ColorBox is a task I am currently implementing and has not yet been submitted.

kranar: Understood, in that case this is fine. When you implement ```ColorBox```, replace all references to ```ValueModel<QColor>``` with the typedef. For this task it's fine as is.

PR #934, opened 2023-10-18
ColorBox

shaobinspire:


Applications/Spire/Include/Spire/Ui/ColorBox.hpp
```diff
@@ -0,0 +1,93 @@
...
+#include "Spire/Ui/PressObserver.hpp"
+#include "Spire/Ui/Ui.hpp"
+
+namespace Spire {
+
+  /** A ValueModel over a QColor. */
+  using ColorModel = ValueModel<QColor>;
+
+  /** A ValueModel over an optional QColor. */
+  using OptionalColorModel = ValueModel<boost::optional<QColor>>;
```
kranar, line 17: What is the use case for the ```OptionalColorModel```? It doesn't look like it's used anywhere and I believe ```QColor``` already represents its own invalid value though ```QColor::isValid()```:

https://doc.qt.io/qt-6/qcolor.html#isValid

kranar (changes requested):
This is fantastic, your code reviews have very few issues.

PR #935, opened 2023-10-24
Highlight picker

shaobinspire:
The `ColorPicker` is using the `Oklch` color space and the `APCA` contrast algorithm. To make it easier to test them in unit tests, I put them into a separate file `ColorConversion`.

The test data of APCA comes from [APCA Script Test](https://github.com/Myndex/apca-w3/blob/master/test/test.html).
The test data for conversion between RGB and Oklab and between RGB and Oklch was calculated by [Online Color Converter](https://ajalt.github.io/colormath/converter/) and [Oklch](https://oklch.com/).

Their implementations refer to the following contents:
1. [ColorM](https://github.com/soreja/colorm/tree/master)
2. [farver](https://github.com/thomasp85/farver/tree/main)
3. [apca-w3](https://github.com/Myndex/apca-w3/#apca-w3-basic-math-srgb).


Applications/Spire/Include/Spire/Ui/ColorConversion.hpp
```diff
@@ -0,0 +1,78 @@
...
+#define SPIRE_COLOR_CONVERSION_HPP
+#include <QColor>
+
+namespace Spire {
+
+  /** Represents a Oklab color. */
+  struct OklabColor {
+
+    /** Specifies if the color is valid. */
+    bool m_is_valid;
```
kranar, line 11: When would this be false?

shaobinspire: When `QColor` converts to `OklabColor` and `QColor` is invalid, `OklabColor` will be invalid and `m_is_valid` will be false.

kranar: Okay, in that case write a function to check if an ```OklabColor``` is valid, and reserve some kind of representation for an invalid ```OklabColor```, for example if any component is ```-1``` then ```is_valid``` returns ```false```.

Applications/Spire/Include/Spire/Ui/ColorConversion.hpp
```diff
@@ -18,14 +15,20 @@ namespace Spire {
...
     double m_b;
+
+    /** Constructs an invalid OklabColor with the lab value (-1, -1, -1) */
+    OklabColor();
+
+    /** Constructs an OklabColor. */
+    OklabColor(double l, double a, double b);
+
+    /** Returns <code>true</code> iff the OklabColor is valid. */
+    bool is_valid() const;
```
kranar, line 26: Make this a free function.

Applications/Spire/Source/Ui/ColorConversion.cpp
```diff
@@ -50,13 +50,46 @@ namespace {
...
+    m_a(-1),
+    m_b(-1) {}
+
+OklabColor::OklabColor(double l, double a, double b)
+  : m_l(l),
+    m_a(a),
+    m_b(b) {}
+
+bool OklabColor::is_valid() const {
+  if(m_l < 0 || m_l > 1.0 || m_a < -0.4 || m_a > 0.4 ||
```
kranar, line 64: return m_l >= 0. && m_l <= 1. && m_a >= -0.4 && m_a <= 0.4 && m_b >= -0.4 && m_b <= 0.4;

Applications/Spire/Source/Ui/ColorConversion.cpp
```diff
@@ -50,13 +50,46 @@ namespace {
...
+    m_c(-1),
+    m_h(0) {}
+
+OklchColor::OklchColor(double l, double c, double h)
+  : m_l(l),
+    m_c(c),
+    m_h(h) {}
+
+bool OklchColor::is_valid() const {
+  if(m_l < 0 || m_l > 1.0 || m_c < 0 || m_c > 0.4) {
```
kranar, line 82: m_h can be any value?

kranar (changes requested):

Applications/Spire/Source/Ui/ColorConversion.cpp
shaobinspire: Yes, because m_h represents the hue angle of a color and there is no limit for the angle value.
https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/oklch

I will add some unit tests for it.

PR #943, opened 2023-12-05
TextBox missing submit signal

shaobinspire:
The updated TextBox is missing the submit signal when the Enter key is pressed or the TextBox focuses out. This PR is trying to fixed it.


kranar (approved):
Thanks for this fix.

PR #947, opened 2024-01-15
The read-only DropDownBox is editable.

shaobinspire:


Applications/Spire/Source/Ui/DropDownBox.cpp
```diff
@@ -353,6 +353,9 @@ void DropDownBox::submit() {
 }
 
 void DropDownBox::on_button_press_end(PressObserver::Reason reason) {
+  if(is_read_only()) {
```
kranar, line 356: Can you consider placing this check closer to its consequence?

    if(m_down_list->isVisible()) {
      ...
    } else if(!is_read_only()) {
      ...
    }

kranar (approved):

PR #950, opened 2024-01-23
Color swatch

shaobinspire:


Applications/Spire/Source/Ui/ColorSwatch.cpp
```diff
@@ -0,0 +1,87 @@
...
+    style.get(Any()).
+      set(border(scale_width(1), QColor(Qt::transparent))).
+      set(horizontal_padding(scale_width(1))).
+      set(vertical_padding(scale_height(1)));
+    style.get(Hover() || Focus()).
+      set(border_color(QColor(0x4B23A0)));
+    style.get(Drag()).
+      set(BackgroundColor(QColor(255, 255, 255, 204))).
+      set(border_color(QColor(0x4B23A0)));
+    style.get(Any() >> Body()).
```
kranar, line 27: Why is the descendant selector used here? Could you instead use either the top selector or select the component directly?

Applications/Spire/Source/Ui/ColorSwatch.cpp
```diff
@@ -0,0 +1,87 @@
...
+  }
+}
+
+void ColorSwatch::on_current(const QColor& color) {
+  update_style(*this, [&] (auto& style) {
+    style.get(Any() >> Body()).set(BackgroundColor(color));
+    style.get(Drag() >> Body()).
+      set(BackgroundColor(
+        QColor(color.red(), color.green(), color.blue(), 204)));
+    style.get(Highlighted() >> Body()).
```
kranar, line 79: I don't see a use of ```color``` here. Could this not be moved to the default style?

kranar (changes requested):
Good start.

Applications/Spire/Source/Ui/ColorSwatch.cpp
shaobinspire: When both the Drag and Highlighted selectors are applied to Body(), the Highlighted style is obscured by the Drag style. From my experiment, it appears that the Highlighted style remains visible only when it is applied after the Drag style. Hence, it was placed here.

Applications/Spire/Source/Ui/ColorSwatch.cpp
shaobinspire: Initially, I used the top selector, which worked well for ColorSwatch alone. However, when the ColorSwatch is used within the ColorPicker, it seems that styles are applied to the entire ColorSwatch instead of its Body().
I can update code to apply these styles to the component directly.

shaobinspire: I will replace the Body selector with the Swatch selector to resolve this issue.

Applications/Spire/Source/Ui/ColorSwatch.cpp
shaobinspire: Considering that highlighted style is rarely set while dragging, I don't think it would be an issue that the Highlighted style obscures the Drag style. I will update this code.

PR #957, opened 2024-02-08
Updated the width of the menu based on the new specs

shaobinspire:


Applications/Spire/Source/Ui/MenuButton.cpp
```diff
@@ -123,3 +137,33 @@ void MenuButton::mouseReleaseEvent(QMouseEvent* event) {
...
+    {MINIMUM_MENU_WIDTH(), width(), m_menu->sizeHint().width()});
+  auto window_width = max_width + margins.left() + margins.right();
+  if(max_width == m_menu->sizeHint().width()) {
+    window_width += m_menu_border_size;
+  }
+  m_menu_window->setFixedWidth(window_width);
+}
+
+void MenuButton::on_menu_window_style() {
+  m_menu_border_size = 0;
```
kranar, line 154: No update is being performed, when the style is changed there should be a corresponding change to the visual appearance of the widget.

PR #962, opened 2024-02-21
Window Update

shaobinspire:


Applications/Spire/Source/KeyBindings/KeyBindingsWindow.cpp
```diff
@@ -32,7 +32,6 @@ KeyBindingsWindow::KeyBindingsWindow(
   setWindowTitle(tr("Key Bindings"));
```
kranar, line 32: Changes to the KeyBindingsWindow don't belong in this branch.

kranar (changes requested):
Separate the changes to the Window and the KeyBindingsWindow, everything else looks good!

PR #964, opened 2024-02-21
Window Updates

shaobinspire:


Applications/Spire/Include/Spire/Ui/Window.hpp
```diff
@@ -49,6 +49,7 @@ namespace Spire {
       mutable ClosedSignal m_closed_signal;
       TitleBar* m_title_bar;
       QWidget* m_body;
+      QLayout* m_layout;
```
kranar, line 52: Why was this field added?

kranar (changes requested):

Applications/Spire/Include/Spire/Ui/Window.hpp
shaobinspire: It is not required and will be removed.

kranar (approved):

PR #965, opened 2024-02-21
Key bindings window

shaobinspire:
The changes of the EditableBox and TableView come from the old version of KeyBindings.


kranar (approved):
Looks good, keep the momentum going!

PR #969, opened 2024-02-23
Cancel key bindings form

shaobinspire:


Applications/Spire/Source/KeyBindings/CancelKeyBindingsModel.cpp
```diff
@@ -38,29 +38,30 @@ QValidator::State CancelKeyBindingsModel::on_validate(Operation operation,
...
   }
   return QValidator::Invalid;
 }
 
 void CancelKeyBindingsModel::on_update(Operation operation,
     const QKeySequence& sequence) {
-  if(sequence.isEmpty()) {
-    return;
-  }
+  auto update = [=] {
```
kranar, line 52: Capture by reference unless you need the lambda expression to outlive it's current scope.

Applications/Spire/Source/KeyBindings/CancelKeyBindingsModel.cpp
```diff
@@ -38,29 +38,30 @@ QValidator::State CancelKeyBindingsModel::on_validate(Operation operation,
...
+    m_bindings_map.erase(m_previous_bindings[static_cast<int>(operation)]);
+    if(!sequence.isEmpty()) {
+      m_bindings_map[sequence] = operation;
+    }
+    m_previous_bindings[static_cast<int>(operation)] = sequence;
```
kranar, line 57: What does ```m_previous_bindings``` do?

kranar (changes requested):
Looks really good now.

Applications/Spire/Source/KeyBindings/CancelKeyBindingsModel.cpp
shaobinspire: It records the previous bindings and is used to remove the old binding from the `m_bindings_map`. `m_bindings_map` can not remove old key sequence when the update signal triggers (a new key sequence is created),  because the signal doesn't provide the value before it is updated.

kranar: Thanks for your explanation.

PR #973, opened 2024-02-28
InteractionsKeyBindingsForm

shaobinspire:
This pull request can only compile successfully if the https://app.asana.com/0/0/1206686076467409/f is fixed.  I temporarily commented out the following code in `ScalarValueModelDecorator` to get it compiled for the demo.

```cpp
 if(unwrapped_value != m_model->get() &&
     fmod(unwrapped_value, *m_increment) != static_cast<Type>(0)) {
   return QValidator::State::Invalid;
 }
```

Another thing is that I added a function auto make_interactions_page in KeyBindingsWindow to display the InteractionKeyBindingsForm. This is function will be removed when the entire interactions page implementation is complete.


Applications/Spire/Source/KeyBindings/InteractionsKeyBindingsForm.cpp
```diff
@@ -0,0 +1,201 @@
...
+    return new MoneyBox(
+      std::make_shared<OptionalScalarValueModelDecorator<Money>>(
+        std::move(model)));
+  }
+
+  template<typename T>
+  std::shared_ptr<ScalarValueModel<T>> get_increment(
+      const std::shared_ptr<InteractionsKeyBindingsModel>& bindings,
+      Qt::KeyboardModifier modifier) {
+    return std::shared_ptr<ScalarValueModel<T>>();
```
kranar, line 102: Both ```bindings``` and ```modifier``` are unused.

PR #976, opened 2024-03-02
Interactions page

shaobinspire:


kranar (changes requested):
Too many issues with the implementation. A lot of concepts that we went over the past couple of years are not translating into the implementation. Things like separating logic from presentation, using models to represent data, using data transformations instead of hardcoding mutable state all over the place.

For small pull requests I can usually point out specific issues, but this is a much larger pull request than usual and it's full of these problems. Unfortunately the concepts I mentioned are absolutely important and are mandatory in order to implement the more complex functionality that Spire needs, otherwise we end up with a bunch of really hard to maintain buggy code.

If you can, please go over the implementation and do what you can to improve upon it, otherwise I can look at implementing it on my end when I get time.

shaobinspire:
I implemented two custom data model `RegionInteractionsListModel` and `RegionInteractionsKeyBindingsModel` to separate logic from representation. However, they didn’t fully adhere to the interface of data models. For instance, **`RegionInteractionsListModel`** did not inherit the interfaces provided by **`ListModel`**. I refactored these two models to align them as closely as possible with their interface definitions in this update.

I also added a method to set a region to InteractionsKeyBindingsForm, which can provide the ability to modify the text title at the top of InteractionsKeyBindingsForm.

kranar:
Okay, feel free to start working on the tasks tab. I will review this branch and make changes as needed.

PR #977, opened 2024-03-12
EditableBox update

shaobinspire:
Since the EditableBox in the master branch is different from that in the key_bindings branch, the EditableBox in key_bindings seems to be newer, I updated its implementation based on the key_bindings branch.


kranar (approved):
Good work!

PR #982, opened 2024-03-25
EditableTableView 

shaobinspire:


Applications/Spire/Include/Spire/Ui/EditableTableView.hpp
```diff
@@ -0,0 +1,114 @@
...
+      PopupBox* m_popup_box;
+      QWidget* m_tip_window;
+      bool m_has_sent_event;
+  };
+
+  /** Represents an editable TableView. */
+  class EditableTableView : public TableView {
+    public:
+
+      using HeaderModel = TableView::HeaderModel;
```
kranar, line 36: Why are all these typedefs here?

Applications/Spire/Include/Spire/Ui/EditableTableView.hpp
```diff
@@ -0,0 +1,114 @@
...
+        std::shared_ptr<TableFilter> table_filter,
+        std::shared_ptr<CurrentModel> current,
+        std::shared_ptr<SelectionModel> selection, ViewBuilder view_builder,
+        Comparator comparator, Filter filter, QWidget* parent = nullptr);
+
+      /** Applies a new filter to this TableView. */
+      void set_filter(const Filter& filter);
+
+      /** Connects a slot to the DeleteSignal. */
+      boost::signals2::connection connect_delete_signal(
```
kranar, line 78: Why is there a signal instead of using the ```TableModel```? The ```TableModel``` is how changes are intended to be communicated.

Applications/Spire/Include/Spire/Ui/EditableTableView.hpp
```diff
@@ -0,0 +1,114 @@
...
+       */
+      EditableTableView(std::shared_ptr<TableModel> table,
+        std::shared_ptr<HeaderModel> header,
+        std::shared_ptr<TableFilter> table_filter,
+        std::shared_ptr<CurrentModel> current,
+        std::shared_ptr<SelectionModel> selection, ViewBuilder view_builder,
+        Comparator comparator, Filter filter, QWidget* parent = nullptr);
+
+      /** Applies a new filter to this TableView. */
+      void set_filter(const Filter& filter);
```
kranar, line 75: I don't see a purpose for this method, changes to the model should be handled by the ```TableModel```.

kranar (changes requested):
You will need to break this down into a smaller pull request. There are too many issues with this implementation and it's not something that can be fixed in one shot.

Applications/Spire/Include/Spire/Ui/EditableTableView.hpp
shaobinspire: The `TableModel` doesn't provide the `Remove` or `Delete` method, so the `EditableTableView` can't remove a row internally. This signal is used to tell the user of the `EditableTableView` to remove the specified row.

shaobinspire:
> You will need to break this down into a smaller pull request. There are too many issues with this implementation and it's not something that can be fixed in one shot.

I removed code related to removing and adding functionalities from the EditableTableView.

kranar:
Create a branch that implements just the delete row functionality and submit a PR for it.

PR #983, opened 2024-03-27
EditableTableView deletes a row

shaobinspire:
In terms of deleting a row, I can’t think of any other solution than triggering a Delete signal to the `TableView` user to delete a row. I think the `TableModel` should provide the operation interfaces, such as push, remove and move, similar to `ListModel`. If the `TableModel` offers these interfaces, the Delete signal would no longer be necessary.

Let me explain some structures used for the `EditableTableView`:

1. `EditableTableModel`: It adds an extra row as the last empty row and two columns as the first column and the last column.
2. `EditableTableHeaderModel`: It adds two header items as the first column and the last column.
3. `RevertColumnTableModel`: It removes two columns added by the EditableTableModel.
4. `CustomListValueModel`: It is used to resolve the `out_of_range` issue. After a row is deleted, the `TextBox` in the `TableView` receives the style update signal and attempt to access the data that has been removed from the model to render the content. This results in an `out_of_range` exception, as the index in the `ListValueModel` has been set to -1.

### DIFF
--- a/Applications/Spire/Source/Ui/EditableTableView.cpp
+++ b/Applications/Spire/Source/Ui/EditableTableView.cpp
@@ -1,0 +1,441 @@
+#include "Spire/Ui/EditableTableView.hpp"
+#include "Spire/Spire/Dimensions.hpp"
+#include "Spire/Ui/Box.hpp"
+#include "Spire/Ui/Button.hpp"
+#include "Spire/Ui/DropDownBox.hpp"
+#include "Spire/Ui/EditableBox.hpp"
+#include "Spire/Ui/Icon.hpp"
+#include "Spire/Ui/Layouts.hpp"
+#include "Spire/Ui/ScrollBox.hpp"
+#include "Spire/Ui/TableBody.hpp"
+#include "Spire/Ui/TableItem.hpp"
+
+using namespace boost;
+using namespace boost::signals2;
+using namespace Spire;
+using namespace Spire::Styles;
+
+namespace {
+  using DeleteButton = StateSelector<void, struct DeleteButtonSeletorTag>;
+  using EmptyCell = StateSelector<void, struct EmptyCellSeletorTag>;
+
+  const auto DELETE_TIMEOUT_MS = 100;
+
+  auto TABLE_VIEW_STYLE() {
+    auto style = StyleSheet();
+    style.get(Any() > is_a<ScrollBox>()).
+      set(BackgroundColor(QColor(0xFFFFFF))).
+      set(PaddingBottom(64));
+    style.get(Any() > Current()).
+      set(BackgroundColor(Qt::transparent));
+    style.get(Any() > HoverItem()).set(border_color(QColor(0xA0A0A0)));
+    style.get((Any() > DeleteButton()) << is_a<TableItem>()).
+      set(padding(scale_width(2)));
+    style.get((Any() > DeleteButton()) << (HoverItem() || Current())).
+      set(border_color(QColor(Qt::transparent)));
+    style.get(Any() > (DeleteButton() && Hover()) > is_a<Icon>()).
+      set(BackgroundColor(QColor(0xDFDFEB))).
+      set(Fill(QColor(0xB71C1C)));
+    style.get((Any() > EmptyCell()) << (HoverItem() || Current())).
+      set(border_color(QColor(Qt::transparent)));
+    style.get(Any() > CurrentRow() > DeleteButton()).
+      set(Visibility(Visibility::VISIBLE));
+    style.get(Any() > CurrentRow() > DeleteButton() > is_a<Icon>()).
+      set(Fill(QColor(0x535353)));
+    style.get(Any() > CurrentRow() > (DeleteButton() && Hover()) >
+        is_a<Icon>()).
+      set(BackgroundColor(QColor(0xD0CEEB))).
+      set(Fill(QColor(0xB71C1C)));
+    style.get(Any() > CurrentRow()).set(BackgroundColor(QColor(0xE2E0FF)));
+    style.get(Any() > CurrentColumn()).set(BackgroundColor(Qt::transparent));
+    return style;
+  }
+
+  auto DELETE_BUTTON_STYLE() {
+    auto style = StyleSheet();
+    style.get(Any()).set(Visibility(Visibility::INVISIBLE));
+    style.get(Any() > is_a<Box>()).
+      set(BackgroundColor(QColor(Qt::transparent))).
+      set(horizontal_padding(scale_width(2))).
+      set(vertical_padding(scale_height(2)));
+    style.get(Any() > is_a<Icon>()).
+      set(BackgroundColor(QColor(Qt::transparent)));
+    return style;
+  }
+
+  auto TABLE_ROW_STYLE(StyleSheet style) {
+    style.get(Any() > is_a<DropDownBox>() >
+        (is_a<TextBox>() && !(+Any() << is_a<ListItem>()))).
+      set(horizontal_padding(scale_width(8)));
+    style.get(Any() > is_a<EditableBox>()).
+      set(horizontal_padding(scale_width(8)));
+    style.get(Hover()).set(BackgroundColor(0xF2F2FF));
+    style.get(Hover() > DeleteButton()).
+      set(Visibility(Visibility::VISIBLE));
+    style.get(Hover() > DeleteButton() > is_a<Icon>()).
+      set(Fill(QColor(0x535353)));
+    return style;
+  }
+
+  QWidget* make_empty_cell() {
+    auto cell = new QWidget();
+    cell->setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Expanding);
+    match(*cell, EmptyCell());
+    return cell;
+  }
+
+  EditableBox* find_editable_box(QWidget& widget) {
+    if(auto editable_box = dynamic_cast<EditableBox*>(&widget)) {
+      return editable_box;
+    }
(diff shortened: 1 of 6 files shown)


Applications/Spire/Source/Ui/EditableTableView.cpp
```diff
@@ -0,0 +1,431 @@
...
+          std::bind_front(&EditableTableView::delete_current_row, this));
+      });
+      return button;
+    } else {
+      return make_empty_cell();
+    }
+  } else if(column == table->get_column_size() - 1) {
+    return make_empty_cell();
+  } else {
+    return source_view_builder(std::make_shared<RevertColumnTableModel>(table),
```
kranar, line 381: Why is a ```RevertColumnTableModel``` being used instead of passing the ```table``` argument from the constructor?

Applications/Spire/Source/Ui/EditableTableView.cpp
```diff
@@ -0,0 +1,431 @@
...
+    std::shared_ptr<CurrentModel> current,
+    std::shared_ptr<SelectionModel> selection, ViewBuilder view_builder,
+    Comparator comparator, QWidget* parent)
+    : TableView(std::make_shared<EditableTableModel>(std::move(table), header),
+        std::make_shared<EditableTableHeaderModel>(header),
+        std::move(table_filter), std::move(current), std::move(selection),
+        std::bind_front(&EditableTableView::make_table_item, this,
+          std::move(view_builder)),
+        std::move(comparator), parent) {
+  auto table_header = static_cast<TableHeader*>(static_cast<Box*>(
```
kranar, line 329: Do not code against implementation details, prefer to code against an API. If you need to modify an API then propose the modification and code against it.

Applications/Spire/Source/Ui/EditableTableView.cpp
```diff
@@ -0,0 +1,431 @@
...
+  scroll_box.setFocusPolicy(Qt::NoFocus);
+  m_table_body = static_cast<TableBody*>(&scroll_box.get_body());
+  m_table_body->installEventFilter(this);
+  set_style(*this, TABLE_VIEW_STYLE());
+  for(auto i = 0; i < get_table()->get_row_size(); ++i) {
+    m_rows.push({});
+  }
+  for(auto i = 0; i < m_table_body->get_table()->get_row_size(); ++i) {
+    auto row = m_table_body->layout()->itemAt(i)->widget();
+    update_style(*row, [] (auto& style) {
```
kranar, line 347: What is the advantage of styling row by row instead using the top selector?

Applications/Spire/Source/Ui/EditableTableView.cpp
```diff
@@ -0,0 +1,431 @@
...
+    }
+  } else if(column == table->get_column_size() - 1) {
+    return make_empty_cell();
+  } else {
+    return source_view_builder(std::make_shared<RevertColumnTableModel>(table),
+      row, column - 1);
+  }
+}
+
+void EditableTableView::set_column_cover_mouse_events_transparent() {
```
kranar, line 386: Is this related to deletion? If not then remove it.

kranar (changes requested):
It's a good start so far.

Applications/Spire/Source/Ui/EditableTableView.cpp
```diff
@@ -0,0 +1,431 @@
...
+      m_header(std::move(header)),
+      m_source_connection(m_source->connect_operation_signal(
+        std::bind_front(&EditableTableModel::on_operation, this))) {}
+
+  int get_row_size() const {
+    return m_source->get_row_size() + 1;
+  }
+
+  int get_column_size() const {
+    return m_header->get_size() + 2;
```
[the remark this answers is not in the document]
shaobinspire: When the column size is obtained from the source and all rows are removed from the source, this function will return 2. This will result in no vertical gridlines starting from the second column in the last empty row. Using header to retrieve the column count resolves this issue. 

Jon noted [this issue](https://app.asana.com/0/0/1206604117618379/1206823596078434/f) in Asana.

Applications/Spire/Source/Ui/EditableTableView.cpp
```diff
@@ -0,0 +1,431 @@
...
+        column >= get_column_size()) {
+      throw std::out_of_range("The row or column is out of range.");
+    }
+    if(column == 0) {
+      static auto row_index = 0;
+      row_index = row;
+      return row_index;
+    }
+    column -= 1;
+    if(row < m_source->get_row_size() && column < m_source->get_column_size()) {
```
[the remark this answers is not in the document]
shaobinspire: Line 168 checks if the row or column is outside the range of this model, while line 178 checks whether the row or column is within the range of the source. The model appends a row and a column to the source's row-end and column-end, so when row is equal to the source row size or column is equal to the source column size, it will return none.

Applications/Spire/Source/Ui/EditableTableView.cpp
shaobinspire: `table` in the callback function `view_builder` is different from `table` in the constructor. `table` in the callback function `view_builder` is the transformed table after sorting and filtering. `table` from the constructor can be passed to it, but the second parameter `row` must be converted to source's `row`.

Applications/Spire/Source/Ui/EditableTableView.cpp
```diff
@@ -0,0 +1,431 @@
...
+    return m_header->get_size() + 2;
+  }
+
+  AnyRef at(int row, int column) const {
+    if(row < 0 || row >= get_row_size() || column < 0 ||
+        column >= get_column_size()) {
+      throw std::out_of_range("The row or column is out of range.");
+    }
+    if(column == 0) {
+      static auto row_index = 0;
```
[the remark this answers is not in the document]
shaobinspire: `AnyRef(row, AnyRef::by_value)` doesn’t work for `TableModel::get`. `AnyRef(row, AnyRef::by_value)` is returned from `EditableTableModel::at`, but it is destroyed before returning from `TableModel::get`, as a result, `TableModel::get` returns an invalid value.

```
template<typename T>
const T& TableModel::get(int row, int column) const {
  return any_cast<const T>(at(row, column));
}
```

To avoid it, I used `TableModel::at ` directly instead of `TableModel::get`:
```
m_rows.set(any_cast<int>(m_table_body->get_table()->at(i, 0)), row);
```

Applications/Spire/Source/Ui/EditableTableView.cpp
```diff
@@ -0,0 +1,334 @@
...
+using namespace Spire::Styles;
+
+namespace {
+  using DeleteButton = StateSelector<void, struct DeleteButtonSeletorTag>;
+  using EmptyCell = StateSelector<void, struct EmptyCellSeletorTag>;
+
+  const auto DELETE_TIMEOUT_MS = 100;
+
+  auto TABLE_VIEW_STYLE() {
+    auto style = StyleSheet();
```
kranar, line 25: Remove any styles that are not related to deletion.

Applications/Spire/Source/Ui/EditableTableView.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+  explicit EditableTableModel(std::shared_ptr<TableModel> source,
+    std::shared_ptr<EditableTableView::HeaderModel> header)
+    : m_source(std::move(source)),
+      m_header(std::move(header)),
+      m_source_connection(m_source->connect_operation_signal(
+        std::bind_front(&EditableTableModel::on_operation, this))) {}
+
+  int get_row_size() const {
+    return m_source->get_row_size() + 1;
```
kranar, line 96: Is this row added in order to be able to add a row? If so remove it since it doesn't involve deletion.

Applications/Spire/Source/Ui/TableHeader.cpp
```diff
@@ -64,6 +64,10 @@ Button& TableHeader::get_filter_button(int column) {
   return m_item_views[column]->get_filter_button();
 }
 
+TableHeaderItem& TableHeader::get_item(int column) {
```
kranar, line 67: What if the column is out of bounds?

kranar (changes requested):
Much better.

kranar:
Merge master into this branch. The TableModel has added support for removing rows. Use that functionality to remove rows instead of signaling a delete.

shaobinspire:
There is an issue that I can’t find a proper solution for. When the row containing a `TextBox` is deleted, `TextBox’s` contents are updated before its model `ColumnViewListModel` gets operation signal from the source to update its index. It results in an issue that `ColumnViewListModel` accesses data from the updated the source model with the un-updated row index.  For example, there are three rows initially and the first row is deleted. The destruction of the `TextBox` causes the content to be updated (see attached screenshot). However, the row index of the `TextBox’s` model in the first row is still 1 (it is supposed to be 0) as the model hasn’t received the source operation signal yet. Hence, the TextBox in the first row gets the data of the TextBox in the second row. That is why I created the `CustomColumnViewListModel`, but it doesn't solve the problem very well.

To reproduce it:

1. Open the EditableTableView.
2. Click on the DeleteButton on the first row.
3. The first row is deleted.
4. The second row becomes the first row, and its name becomes “Test3”.
5. Expected: Its name should be “Test2”.
6. Hover on the name column of the first row.
7. The name is updated to “Test2”.
image

kranar:
We can meet tomorrow to go over your example.

kranar:
Please review and test the changes I made. If you are satisfied let me know and I will merge them into master.

PR #985, opened 2024-04-03
adding a new row to EditableTableView

shaobinspire:
I have put the demo to demos/editable_table_view_adding_row.

### DIFF
--- a/Applications/Spire/Source/Ui/EditableTableView.cpp
+++ b/Applications/Spire/Source/Ui/EditableTableView.cpp
@@ -90,7 +90,7 @@ struct EditableTableModel : TableModel {
         std::bind_front(&EditableTableModel::on_operation, this))) {}
 
   int get_row_size() const override {
-    return m_source->get_row_size();
+    return m_source->get_row_size() + 1;
   }
 
   int get_column_size() const override {
@@ -267,18 +267,22 @@ EditableTableView::EditableTableView(
 QWidget* EditableTableView::make_table_item(const ViewBuilder& view_builder,
     const std::shared_ptr<TableModel>& table, int row, int column) {
   if(column == 0) {
-    auto button = make_delete_icon_button();
-    button->setMaximumHeight(scale_height(26));
-    match(*button, DeleteButton());
-    auto tracker = std::make_shared<Tracker>(row);
-    tracker->m_connection = get_table()->connect_operation_signal(
-      std::bind_front(&TableRowIndexTracker::update, &tracker->m_index));
-    button->connect_click_signal([=] {
-      QTimer::singleShot(0, this, [=] {
-        delete_row(tracker->m_index);
+    if(row < table->get_row_size() - 1) {
+      auto button = make_delete_icon_button();
+      button->setMaximumHeight(scale_height(26));
+      match(*button, DeleteButton());
+      auto tracker = std::make_shared<Tracker>(row);
+      tracker->m_connection = get_table()->connect_operation_signal(
+        std::bind_front(&TableRowIndexTracker::update, &tracker->m_index));
+      button->connect_click_signal([=] {
+        QTimer::singleShot(0, this, [=] {
+          delete_row(tracker->m_index);
+        });
       });
-    });
-    return button;
+      return button;
+    } else {
+      return make_empty_cell();
+    }
   } else if(column == table->get_column_size() - 1) {
     return make_empty_cell();
   } else {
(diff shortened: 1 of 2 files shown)


Applications/Spire/Source/UiViewer/StandardUiProfiles.cpp
```diff
@@ -2101,7 +2168,10 @@ UiProfile Spire::make_editable_table_view_profile() {
         std::make_shared<ListSingleSelectionModel>(),
         std::make_shared<ListEmptySelectionModel>()),
       [=] (const auto& table, auto row, auto column) -> QWidget* {
-        return make_row_cell(table, row, column);
+        if(row < table->get_row_size()) {
+          return make_row_cell(table, row, column);
+        }
+        return make_empty_row_cell(array_table_model, column);
```
kranar, line 2174: Explain this change please.

shaobinspire: It creates cell components for the last empty row.

Applications/Spire/Source/Ui/EditableTableView.cpp
```diff
@@ -267,18 +267,22 @@ EditableTableView::EditableTableView(
...
-    auto button = make_delete_icon_button();
-    button->setMaximumHeight(scale_height(26));
-    match(*button, DeleteButton());
-    auto tracker = std::make_shared<Tracker>(row);
-    tracker->m_connection = get_table()->connect_operation_signal(
-      std::bind_front(&TableRowIndexTracker::update, &tracker->m_index));
-    button->connect_click_signal([=] {
-      QTimer::singleShot(0, this, [=] {
-        delete_row(tracker->m_index);
+    if(row < table->get_row_size() - 1) {
```
kranar, line 270: This looks awkward, help me understand it. The spec says that the last row should be a series of empty cells, if that's the case then that is the dominant condition to check. After that check, then you can check if the column is 0 and insert a delete button, and then after that check you can check if the column is the last column to insert an empty cell, and then finally as a last resort you use the view builder. If I misunderstand something please correct me.

Applications/Spire/Source/UiViewer/StandardUiProfiles.cpp
kranar: Why is it doing that?

shaobinspire: My understanding is that although the last line is a blank line, it does not mean that this line does not contain any components. Each cell in this row is the same as the cell in the previous row except that their value is null. They need to receive user input and generate a new row when submitted. This is why the function is used here.

Applications/Spire/Source/Ui/EditableTableView.cpp
shaobinspire: What component is contained in the empty cell of the last row?

Applications/Spire/Source/UiViewer/StandardUiProfiles.cpp
kranar: Why is test code implementing specifications? Test code is for testing, it's not for implementing functionality.

shaobinspire: Yeah, it should be a part of the EditableTableView.

shaobinspire: If creating empty cells was placed inside the EditableTableView, how to add a new row to the model when a cell submitted a value? Because TableModel has no push or insert interface.

kranar: You will need to consider whatever scenarios are needed and propose a solution. I am not familiar enough with the spec or the requirements to give an answer at this time. If after thinking it over thoroughly you are unable to come up with a proposal or a solution then I can review it and provide some feedback.

shaobinspire:
I worked out two solutions for adding a row through the last row:

1. The EditableTableView adds an extra row at the end (The number of rows returned by EditableTableModel is the number of rows in the source model plus 1.), but each cell in this last row holds none value. Since the EditableTableView doesn’t know the type of the cell in the last row, it can’t create cell components internally. The user of the EditableTableView therefore has the responsibility for creating cell components for the last row. This approach was implemented in the version I submitted last time.

2. The EditableTableView accepts the table where the last row contains the empty values for each cell. Hence, the EditableTableView doesn’t need to add an extra row as the last one. Consequently, The user of the EditableTableView is no longer responsible for creating empty cells for the last row. However, a downside of this approach is that the last row will be involved in sorting. I am unable to find a way to exclude the last row from sorting,  since sorting relies solely on values rather than indices. The first approach can circumvent this issue, as the cell values in the last row are none and none values can be excluded from sorting operations.

The challenge of the approach 2 is that since TableView doesn’t provide an insertion interface, EditableTableView cannot implement adding new rows to the model internally. As an alternative, the responsibility of inserting new data into the table would fall upon the user of the EditableTableView. However, I think it should be handled by the EditableTableView itself, rather than letting the user take on this responsibility.

PR #994, opened 2024-04-16
TaskkeysPage

shaobinspire:


Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  auto make_button(const QString& path, const QString& tooltip) {
+    auto button = make_icon_button(imageFromSvg(path, scale(16, 16)), tooltip);
+    button->setFixedSize(scale(26, 26));
+    return button;
+  }
+}
+
+TaskKeysPage::TaskKeysPage(std::shared_ptr<KeyBindingsModel> key_bindings,
```
kranar, line 38: Why is this a class? I feel like a repeating issue is the use of classes when they are not needed.

A class is used to encapsulate and manage both state and behavior. If all you need is state you use a struct, if all you need is behavior you use a function. If what you need is to manage both state and behavior then you use a class.

What state and behavior is needed for this?

kranar (changes requested):

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
shaobinspire: Since there is no state that the `TaskKeysPage` needs to manage, it is suitable to use a function to create it. However, the `TaskKeysPage` requires key press events to add and duplicate rows.  I found `KeyObserver` can monitor the key press events, but I don’t know how to use it in a function to monitor and get key events.

kranar (approved):
Okay, I did not realize this task was just for the layout and styles.

PR #995, opened 2024-04-19
Task keys operations

shaobinspire:


Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
```diff
@@ -73,18 +87,155 @@ TaskKeysPage::TaskKeysPage(std::shared_ptr<KeyBindingsModel> key_bindings,
...
-    std::move(markets)));
+    std::move(markets));
+  m_table_view->installEventFilter(this);
+  layout->addWidget(m_table_view);
   auto box = new Box(body);
   update_style(*box, [] (auto& style) {
     style.get(Any()).set(BackgroundColor(QColor(0xFFFFFF)));
   });
   enclose(*this, *box);
+  m_new_task_form = new LineInputForm(tr("New Task"), *this);
```
kranar, line 101: Can you only create this form when it's needed and used?

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
```diff
@@ -73,18 +87,155 @@ TaskKeysPage::TaskKeysPage(std::shared_ptr<KeyBindingsModel> key_bindings,
...
+  window->move(
+    mapToGlobal(QPoint(0, 0)) + rect().center() - window->rect().center());
+}
+
+void TaskKeysPage::on_duplicate_task_action() {
+  auto& selection = m_table_view->get_selection()->get_row_selection();
+  if(selection->get_size() == 0) {
+    return;
+  }
+  m_table_view->setFocus();
```
kranar, line 164: Why is focus put back on the table view at this point?

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
```diff
@@ -73,18 +87,155 @@ TaskKeysPage::TaskKeysPage(std::shared_ptr<KeyBindingsModel> key_bindings,
...
+  auto last_current_row = m_table_view->get_current()->get()->m_row;
+  auto sorted_selection = std::vector<int>();
```
kranar, line 166: Would this work?

    auto sorted_selection = std::vector(selection->begin(), selection->end());

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
```diff
@@ -73,18 +87,155 @@ TaskKeysPage::TaskKeysPage(std::shared_ptr<KeyBindingsModel> key_bindings,
...
+  sorted_selection.assign(selection->begin(), selection->end());
+  std::sort(sorted_selection.begin(), sorted_selection.end(), std::greater<int>());
+  auto& table = m_table_view->get_body().get_table();
+  for(auto index : sorted_selection) {
+    auto order_task = OrderTaskArguments();
```
kranar, line 171: This is too brittle, magic numbers everywhere and too many assumptions.

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
```diff
@@ -73,18 +87,155 @@ TaskKeysPage::TaskKeysPage(std::shared_ptr<KeyBindingsModel> key_bindings,
...
+    order_task.m_side = table->get<Side>(index, 5);
+    order_task.m_quantity = table->get<optional<Quantity>>(index, 6);
+    order_task.m_time_in_force = table->get<TimeInForce>(index, 7);
+    order_task.m_additional_tags =
+      table->get<std::vector<Nexus::Tag>>(index, 8);
+    m_key_bindings->get_order_task_arguments()->insert(order_task,
+      m_table_view->get_current()->get()->m_row);
+  }
+  QTimer::singleShot(0, this, [=] {
+    m_table_view->get_current()->set(TableView::Index(last_current_row, 1));
```
kranar, line 185: Why is this done after the current event loop?

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
```diff
@@ -73,18 +87,155 @@ TaskKeysPage::TaskKeysPage(std::shared_ptr<KeyBindingsModel> key_bindings,
...
+  });
+}
+
+void TaskKeysPage::on_delete_task_action() {
+  m_table_view->setFocus();
```
kranar, line 190: Why is ```setFocus``` here?

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
```diff
@@ -73,18 +87,155 @@ TaskKeysPage::TaskKeysPage(std::shared_ptr<KeyBindingsModel> key_bindings,
...
+  auto& selection = m_table_view->get_selection()->get_row_selection();
+  for(auto i = 0; i < selection->get_size(); ++i) {
```
kranar, line 192: Can this be done?

    for(auto i : *m_table_view->get_selection()->get_row_selection()) {
      m_table_view->get_body().get_table()->remove(i);
    }

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
```diff
@@ -73,18 +87,155 @@ TaskKeysPage::TaskKeysPage(std::shared_ptr<KeyBindingsModel> key_bindings,
...
+    m_table_view->get_body().get_table()->remove(selection->get(i));
+  }
+}
+
+void TaskKeysPage::on_new_task_submission(const QString& name) {
+  m_table_view->setFocus();
```
kranar, line 198: See above.

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
```diff
@@ -73,18 +87,155 @@ TaskKeysPage::TaskKeysPage(std::shared_ptr<KeyBindingsModel> key_bindings,
...
+          m_added_region_item = m_table_view->get_body().get_item(index);
+          m_added_region_item->installEventFilter(this);
+        });
+        m_is_row_added = false;
+      }
+    },
+    [&] (const TableModel::RemoveOperation& operation) {
+      if(m_table_view->get_body().get_table()->get_row_size() == 0) {
+        QTimer::singleShot(0, this, [=] {
+          m_table_view->setFocus();
```
kranar, line 237: A lot of these focus issues need to be reconsidered. Changing focus can be a very blunt user experience.

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
```diff
@@ -73,18 +87,155 @@ TaskKeysPage::TaskKeysPage(std::shared_ptr<KeyBindingsModel> key_bindings,
...
+  visit(operation,
+    [&] (const TableModel::AddOperation& operation) {
+      if(m_is_row_added) {
+        QTimer::singleShot(0, this, [=] {
+          auto index = TableView::Index(operation.m_index, 2);
+          m_table_view->get_current()->set(index);
+          m_added_region_item = m_table_view->get_body().get_item(index);
+          m_added_region_item->installEventFilter(this);
+        });
+        m_is_row_added = false;
```
kranar, line 231: This is too brittle, I suspect that this should be moved to line 208. Let me know your thoughts.

kranar (changes requested):

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
```diff
@@ -73,18 +87,155 @@ TaskKeysPage::TaskKeysPage(std::shared_ptr<KeyBindingsModel> key_bindings,
...
 }
 
 const std::shared_ptr<KeyBindingsModel>&
     TaskKeysPage::get_key_bindings() const {
   return m_key_bindings;
 }
+
+bool TaskKeysPage::eventFilter(QObject* watched, QEvent* event) {
+  if(watched == m_table_view) {
+    if(event->type() == QEvent::KeyPress) {
```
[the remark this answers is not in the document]
shaobinspire: Jon expects that these key sequences are handled when the table body is focused and a cell is not in an editing state.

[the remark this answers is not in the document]
kranar: Why isn't the table body propagating the key press up the widget hierarchy?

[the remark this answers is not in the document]
shaobinspire: When the EditableTableView has a current cell, the EditableBox inside the current cell will intercept the key press and trigger the editing state.

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
```diff
@@ -73,18 +87,155 @@ TaskKeysPage::TaskKeysPage(std::shared_ptr<KeyBindingsModel> key_bindings,
...
+        return true;
+      } else if(key_event.modifiers() & Qt::ControlModifier &&
+          key_event.key() == Qt::Key_D) {
+        on_duplicate_task_action();
+        return true;
+      }
+    }
+  } else if(watched == m_added_region_item) {
+    if(event->type() == QEvent::Show) {
+      auto enter_event =
```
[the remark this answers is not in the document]
shaobinspire: My purpose is to start to edit the RegionBox by sending the Enter event to the EditableBox. I will look into it and find a better way to achieve this.

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
shaobinspire: Sure.

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
shaobinspire: `SetFocus` is use to resolve the [issue](https://app.asana.com/0/0/1207094151238171/1207118094988040/f). When the RegionBox pops up, the current row is set incorrectly after adding, copying or deleting rows. I therefore did `SetFocus` for the add, copy and delete operations to resolve this issue. 

The cause reason of the issue is the `EditablePopupBox` invokes `SetFocus` when receiving the end edit signal, causing the current row to be reset. I will think about it for a better solution.

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
shaobinspire: After copying rows, the current cell should be `row_selection[0][1]` based on specs. When new rows are inserted to the table, the current row will remain unchanged instead of the newly added row.

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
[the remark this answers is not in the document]
shaobinspire: I didn't find a better solution for this case.

shaobinspire:
Since EditableBox depends on focus to change the editing status, and the PopupBox relies on focus to popup displays, focus changes become frequent. Particularly the EditablePopupBox combines them, making the situation even more complicated.

I updated the implementation of the EditableBox to reduce the use of “SetFocus”.

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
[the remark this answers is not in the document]
kranar: The current approach cannot be approved. This is precisely the kind of code that tries to force a solution in terms of implementation details rather than working with well-designed APIs.

Your implementation must work with respect to an API, never with respect to implementation details.

Applications/Spire/Source/Ui/EditableBox.cpp
```diff
@@ -171,21 +182,13 @@ void EditableBox::select_all_text() {
 }
 
 void EditableBox::on_focus(FocusObserver::State state) {
-  if(isHidden()) {
+  if(m_input_box->isHidden()) {
     return;
   }
   if(state == FocusObserver::State::NONE) {
```
kranar, line 188: Avoid doing things like:

    if(condition) {
      action(true);
    } else {
      action(false);
    }

Instead it's just:

    action(condition);

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
```diff
@@ -73,18 +88,145 @@ TaskKeysPage::TaskKeysPage(std::shared_ptr<KeyBindingsModel> key_bindings,
...
+    m_key_bindings->get_order_task_arguments()->insert(order_task,
+      m_table_view->get_current()->get()->m_row);
+  }
+  QTimer::singleShot(0, this, [=] {
+    m_table_view->get_current()->set(TableView::Index(last_current_row, 1));
+  });
+}
+
+void TaskKeysPage::on_delete_task_action() {
+  for(auto i : *m_table_view->get_selection()->get_row_selection()) {
```
kranar, line 184: This should be a transaction, in general if a series of operations are being performed that all have the same cause, then those operations should be a transaction.

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
```diff
@@ -73,18 +89,140 @@ TaskKeysPage::TaskKeysPage(std::shared_ptr<KeyBindingsModel> key_bindings,
...
 }
 
 const std::shared_ptr<KeyBindingsModel>&
     TaskKeysPage::get_key_bindings() const {
   return m_key_bindings;
 }
+
+bool TaskKeysPage::eventFilter(QObject* watched, QEvent* event) {
+  if(watched == m_added_region_item) {
+    if(event->type() == QEvent::Show) {
```
kranar, line 118: Why do you need to wait for the Show event in order to ```set_editing```?

kranar (commented):
This looks much much better.

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
shaobinspire: The `ComboBox` within the RegionBox uses a delay initialization to initialize `m_input_focus_proxy`. If the function `set_editing` was invoked at line 219, the `RegionBox` would pop up before the `ComboBox` was displayed, resulting in a crash because the `ComboBox` attempt to send events to `m_input_focus_proxy` which haven’t been initialized (Line 437 in ComboBox.cpp).

Although sending mouse events to uninitialized `m_input_focus_proxy` can be avoided by adding check code before sending events, the `RegionBox` will be displayed before the newly added row is shown. I have put a demo `KeyBindingsUiTester-RegionBox.exe` into demos/task_keys_page_operation to illustrate this case. You will see that the RegionBox is displayed briefly before the new row is shown.

kranar: Okay, thanks.

PR #997, opened 2024-04-24
Unique task key binding

shaobinspire:


Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
```diff
@@ -118,6 +118,7 @@ bool TaskKeysPage::eventFilter(QObject* watched, QEvent* event) {
     if(event->type() == QEvent::Show) {
       static_cast<EditableBox*>(
         &m_added_region_item->get_body())->set_editing(true);
+      m_added_region_item->removeEventFilter(this);
```
kranar, line 121: If you don't want any further events, then this should be placed before line 119.

Applications/Spire/Source/KeyBindings/TaskKeysTableView.cpp
```diff
@@ -161,6 +161,21 @@ namespace {
     return make_list_value_model(
       std::make_shared<ColumnViewListModel<T>>(table, column), row);
   }
+
+  struct RegionKeyHash {
+    std::size_t operator()(
+        const std::pair<Region, QKeySequence>& region_key) const {
+      auto seed = std::size_t(0);
+      hash_combine(seed, hash_range(region_key.first.GetCountries().begin(),
```
kranar, line 169: Why are you reimplementing a hash for ```Region```?

Applications/Spire/Source/KeyBindings/TaskKeysTableView.cpp
```diff
@@ -368,6 +383,144 @@ struct OrderTaskTableModel : TableModel {
   }
 };
 
+struct UniqueTaskKeyTableModel : TableModel {
+  static const auto KeyIndex = static_cast<int>(TableColumn::KEY);
```
kranar, line 387: SNAKE_CASE for constants.

Applications/Spire/Source/KeyBindings/TaskKeysTableView.cpp
```diff
@@ -368,6 +383,144 @@ struct OrderTaskTableModel : TableModel {
...
+    return m_source->remove(row);
+  }
+
+  connection connect_operation_signal(
+      const OperationSignal::slot_type& slot) const override {
+    return m_transaction.connect_operation_signal(slot);
+  }
+
+  void on_operation(const Operation& operation) {
+    auto check_conflicting_row = [=] (const RegionKey& region_key) {
```
kranar, line 446: I don't understand how this works.

My understanding is if the key or region column is updated, then a scan is performed on the table to find another row that has the same key and a region and resets the key for that row.

Why is all this additional work being performed? Realistically the implementation for this should be in the ```set``` method.

kranar (changes requested):

Applications/Spire/Source/KeyBindings/TaskKeysTableView.cpp
shaobinspire: Since resetting the key based on specs is performed when the user submits the region or key, rather than when modifying them.  The work in `on_operation` is to find the conflict row from the update of the region and key. The function `update` performs the reset when the user submits the update.

kranar: Thanks for explaining this. Reimplement the model to reset in the ```set``` method instead of on submit. The specs will be updated to reflect that.

Applications/Spire/Source/KeyBindings/TaskKeysTableView.cpp
```diff
@@ -430,10 +411,34 @@ struct UniqueTaskKeyTableModel : TableModel {
   }
 
   QValidator::State set(int row, int column, const std::any& value) override {
-    return m_source->set(row, column, value);
+    auto update_key = [&] (const Region& region, const QKeySequence& key) {
```
kranar, line 414: When a series of operations need to be performed that are all related to one another, use a transaction to avoid introducing inconsistent states.

Applications/Spire/Source/KeyBindings/TaskKeysTableView.cpp
```diff
@@ -430,10 +411,34 @@ struct UniqueTaskKeyTableModel : TableModel {
...
+      if(m_region_keys.insert({region, key}).second) {
+        return;
+      }
+      for(auto i = 0; i < get_row_size(); ++i) {
+        if(i != row) {
+          if(get<Region>(i, REGION_INDEX) == region &&
```
kranar, line 420: Can you merge these if's together?

kranar (changes requested):

Applications/Spire/Source/KeyBindings/TaskKeysTableView.cpp
```diff
@@ -443,79 +448,14 @@ struct UniqueTaskKeyTableModel : TableModel {
   }
 
   void on_operation(const Operation& operation) {
```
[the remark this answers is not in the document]
shaobinspire: This function is used to forward the operations of the source model.

PR #999, opened 2024-04-30
Task keys search

shaobinspire:


Applications/Spire/Source/KeyBindings/OrderTaskArgumentsMatch.cpp
```diff
@@ -0,0 +1,177 @@
...
+  for(auto i = 0; i < name.length(); ++i) {
+    if(name.right(name.length() - i).startsWith(query, Qt::CaseInsensitive)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+bool Spire::matches(const CountryCode& country, const QString& query) {
+  auto& entry = GetDefaultCountryDatabase().FromCode(country);
```
kranar, line 32: The default databases are for testing only, they can not be used in implementations.

Applications/Spire/Source/KeyBindings/OrderTaskArgumentsMatch.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+bool Spire::matches(const Destination& destination, const QString& query) {
+  auto& entry = GetDefaultDestinationDatabase().FromId(destination);
+  return !entry.m_id.empty() &&
+    (QString::fromStdString(entry.m_id).startsWith(
+        query, Qt::CaseInsensitive) ||
+      matches_by_word(QString::fromStdString(entry.m_description), query));
+}
+
+bool Spire::matches(const OrderType& order_type, const QString& query) {
```
kranar, line 89: Pay attention to whether an argument is passed by reference or passed by value:

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f16-for-in-parameters-pass-cheaply-copied-types-by-value-and-others-by-reference-to-const

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
```diff
@@ -42,16 +44,219 @@ namespace {
   }
 }
 
+struct OrderTaskMatchCache {
```
kranar, line 47: Do not pollute the global namespace.

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
```diff
@@ -42,16 +44,219 @@ namespace {
...
+
+struct TaskKeysPage::FilteredTaskKeysListModel : OrderTaskArgumentsListModel {
+  std::shared_ptr<OrderTaskArgumentsListModel> m_source;
+  std::vector<OrderTaskMatchCache> m_caches;
+  std::vector<int> m_filtered_data;
+  QString m_query;
+  ListModelTransactionLog<OrderTaskArguments> m_transaction;
+  scoped_connection m_source_connection;
+
+  explicit FilteredTaskKeysListModel(
```
kranar, line 70: This looks significantly more complicated than using the existing ```FilteredTableModel```, can you give some reasons why this was used instead?

kranar (changes requested):

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
shaobinspire: The model is a "ListModel". Since `ListModel` lacks a filter model, I implemented this `FilteredModel` based on the implementation of `FilteredTableModel`. If using an existing `FilteredTableModel`, the approach I can think of is to convert the `ListModel` to a `TableModel` for filtering. Because "TaskKeysTableView" has already performed a conversion from "ListModel" to "TableModel", I think converting again would be redundant and inefficient. Do you have any suggestions for using the FilteredTableModel?

Applications/Spire/Source/KeyBindings/OrderTaskArgumentsMatch.cpp
shaobinspire: How to get the name, letter code of a country using country code?

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
kranar: Convert from a ```ListModel``` to a ```TableModel``` and then use a ```FilterTableModel``` on it.

Applications/Spire/Source/KeyBindings/OrderTaskArgumentsMatch.cpp
kranar: Pass in a ```CountryDatabase```.

Applications/Spire/Source/KeyBindings/TaskKeysPage.cpp
shaobinspire: Can I update the first parameter of `make_task_keys_table_view` from `std::shared_ptr<OrderTaskArgumentsListModel>` to `std::shared_ptr<TableModel>` to reduce the need for conversion from `ListModel` to `TableModel`? Otherwise, `ListModel` would have to be converted to `TableModel` to use `FilterTableModel`, and the converted `TableModel` would be converted back to `ListModel` before being passed to `make_task_keys_table_view`.

kranar: Yes there's no issue with that.

kranar (approved):

kranar:
Resolve conflicts.

PR #1005, opened 2024-05-13
TagNameBox

shaobinspire:


Applications/Spire/Source/KeyBindings/TagNameBox.cpp
```diff
@@ -0,0 +1,196 @@
...
+connection TagNameBox::connect_submit_signal(
+    const SubmitSignal::slot_type& slot) const {
+  return m_drop_down_box->connect_submit_signal([=] (const auto&) {
+    slot(m_current->get());
+  });
+}
+
+void TagNameBox::create_info_tip(
+    const std::shared_ptr<AdditionalTag>& current) {
+  if(auto tag = std::dynamic_pointer_cast<OrderFieldAdditionalTag>(current);
```
kranar, line 183: Don't use a dynamic cast, ```TagNameBox``` there's no reason why special treatment should be given to any specific tag.

Applications/Spire/Source/KeyBindings/TagNameBox.cpp
```diff
@@ -0,0 +1,196 @@
...
+          OrderFieldInfoTip::Model())), parent) {}
+
+TagNameBox::TagNameBox(
+    std::shared_ptr<ListModel<std::shared_ptr<AdditionalTag>>> tags,
+    std::shared_ptr<ValueModel<std::shared_ptr<AdditionalTag>>> current,
+    QWidget* parent)
+    : QWidget(parent),
+      m_tags(std::move(tags)),
+      m_current(std::move(current)) {
+  auto name_list_model = std::make_shared<AdditionalTagNameListModel>(m_tags);
```
kranar, line 139: Use ```EnumBox``` instead of reimplementing it.

Applications/Spire/Source/KeyBindings/OrderFieldAdditionalTag.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+using namespace boost;
+using namespace boost::gregorian;
+using namespace boost::posix_time;
+using namespace Nexus;
+using namespace Spire;
+
+namespace {
+  auto make_quantity_modifiers() {
+    return QHash<Qt::KeyboardModifier, Quantity>(
```
kranar, line 21: Why is this constantly repeated everywhere in the codebase?

Applications/Spire/Source/KeyBindings/OrderFieldAdditionalTag.cpp
```diff
@@ -0,0 +1,92 @@
...
+    m_value(std::make_shared<LocalValueModel<optional<Nexus::Tag::Type>>>(
+      tag.GetValue())) {}
+
+const OrderFieldInfoTip::Model& OrderFieldAdditionalTag::get_model() const {
+  return m_model;
+}
+
+const std::shared_ptr<AdditionalTagValueModel>&
+    OrderFieldAdditionalTag::make_value() const {
+  return m_value;
```
kranar, line 56: This is not consistent with the API which states:

    /** Makes a new value suitable for this tag. */

Applications/Spire/Source/KeyBindings/TagNameBox.cpp
```diff
@@ -0,0 +1,196 @@
...
+        [&] (const ListModel<std::shared_ptr<AdditionalTag>>::UpdateOperation&
+            operation) {
+          m_transaction.push(UpdateOperation(operation.m_index,
+            operation.get_previous()->get_name(),
+            operation.get_value()->get_name()));
+        });
+    }
+  };
+
+  struct AdditionalTagNameValueModel : ValueModel<QString> {
```
kranar, line 70: What's wrong with using a transform?

    make_transform_value_model(std::move(source),
      [] (const auto& tag) {
        return tag->get_name();
      },
      [list] (const auto& name) {
        auto i = std::find_if(list->begin(), list->end(),
          [&] (const auto& tag) {
            return name == tag->get_name();
          });
          if(i == list->end()) {
            throw std::invalid_argument("Tag name not found.");
          }
          return *i;
      });

Applications/Spire/Source/KeyBindings/TagNameBox.cpp
```diff
@@ -0,0 +1,196 @@
...
+#include "Spire/Ui/DropDownBox.hpp"
+#include "Spire/Ui/Layouts.hpp"
+
+using namespace boost;
+using namespace boost::signals2;
+using namespace Spire;
+using namespace Spire::Styles;
+
+namespace {
+  struct AdditionalTagNameListModel : ListModel<QString> {
```
kranar, line 15: Instead of always repeating and copying/pasting code over and over again, try to come up with general solutions that can be composed together. There is too much boilerplate and repetition.

kranar (changes requested):

Applications/Spire/Source/KeyBindings/TagNameBox.cpp
shaobinspire: How to fill the `OrderFieldInfoTip` up used within the `TagNameBox`? The `AdditionalTag` doesn't provide information for the `OrderFieldInfoTip` except providing the tag name.

kranar: I don't know, the issue is if I have to spend the time to think over the design of this, then I may as well implement it myself.

Try to think through the tasks assigned to you, come up with use cases, think about possible scenarios and implementations, and then only after you have a really good understanding of the problem, proceed to implement it. When I see hacky solutions like dynamic casts and other tricks that subvert the type system it's a strong indication that the problem or the task is not well understood and wasn't properly thought through and I can't approve these solutions.

Applications/Spire/Source/KeyBindings/OrderFieldAdditionalTag.cpp
kranar: Okay sounds good. Create a task for this issue and implement it in a separate branch.

Applications/Spire/Source/KeyBindings/OrderFieldAdditionalTag.cpp
shaobinspire: If the function returns a new value every time it’s called, I think it would be reasonable to return a value instead of a reference, since returning the address of a local object would be invalid.

I propose updating the return type of `AdditionalTag::make_value`:
```
virtual std::shared_ptr<AdditionalTagValueModel>
    make_value() const = 0;
```

kranar: Good catch, that makes sense.

Applications/Spire/Source/KeyBindings/TagNameBox.cpp
shaobinspire: `EnumBox<std::shared_ptr<AdditionalTag>>` can’t be instantiated because `EnumBox<T>::default_view_builder` uses `to_text`, but `to_text` lacks the version of `std::shared_ptr<AdditionalTag>`.

kranar: ```to_text``` is an overloaded function that uses static polymorphism to determine what function to call. There is no reason why it can't be used with an ```EnumBox```.

Applications/Spire/Source/KeyBindings/TagNameBox.cpp
shaobinspire: I came up with `TransformListModel` class to preform the conversion from the source type in a list to another type, which functions like `TransformValueModel`.

[API](https://gist.github.com/shaobinspire/f2cc40742d806ec6029113e371e30812)

kranar: The API looks good. Focus on the other issues with this task and then this can be added as a new task in Asana.

Applications/Spire/Source/KeyBindings/TagNameBox.cpp
shaobinspire: I proposed adding a new method to `AdditionalTag` to create a tip.
 ```
 /**
   * Makes a tip displaying the information of this tag.
   * @param parent The parent of the tip.
   */
  virtual QWidget* make_info_tip(QWidget& parent) const = 0;
```

Comment (kranar) [reply Applications/Spire/Source/KeyBindings/TagNameBox.cpp:183]: That method looks too wide, can you narrow it down?

Comment (shaobinspire) [reply Applications/Spire/Source/KeyBindings/TagNameBox.cpp:183]: I can restrict it to return the `OrderFieldInfoTip` to display the tag information.
```
/**
  * Makes an OrderFieldInfoTip displaying the information of this tag.
  * @param parent The parent of the tip.
  */
 virtual OrderFieldInfoTip* make_info_tip(QWidget& parent) const = 0;
 ```

kranar: Why is it returning an ```OrderFieldInfoTip```? Why does it take a parent as a parameter? What is the utility of having every single derived class do this?

kranar: Why are these virtual? Virtual functions are intended to allow for polymorphic behavior, what polymorphic behavior, or any behavior are these methods supposed to have other than returning values?

This task has gone on for too long and I don't see much progress being made.

Please close it and work on something else.

kranar:
Pull requests are not suitable for discussion of design and implementation strategies. Use a pull request only when the work has reached a point where it is ready to be merged and needs one, or possibly two rounds of review for minor changes.

PR #1006, opened 2024-05-14
QuantityBox constructor 

shaobinspire:


kranar (changes requested):
I'm still seeing a lot of these QHash's all over the codebase.

kranar:
Resubmit the pull request once the whole issue has been addressed.

shaobinspire:
I updated the constructors of MoneyBox and PercentBox. Since the specs of Slider and Silder2D don't specify the default modifiers, I left their constructors unchanged.

Applications/Spire/Source/Ui/ColorCodePanel.cpp
```diff
@@ -16,20 +16,13 @@ using namespace Spire;
...
-    return QHash<Qt::KeyboardModifier, T>(
-      {{Qt::NoModifier, 1}, {Qt::AltModifier, 5}, {Qt::ControlModifier, 10},
-        {Qt::ShiftModifier, 20}});
-  }
-
   struct CustomIntegerBox : IntegerBox {
     using EditingFinishSignal = Signal<void (const optional<int>& value)>;
     mutable EditingFinishSignal m_editing_finish_signal;
 
     CustomIntegerBox(std::shared_ptr<OptionalIntegerModel> current,
```
kranar, line 23: What is this constructed used for? Can you remove it?

kranar (changes requested):

Applications/Spire/Source/Ui/ColorCodePanel.cpp
shaobinspire: I don’t think it can be removed. The CustomIntegerBox needs the model of current to construct.

kranar: It needs it for what? The constructor isn't using it for anything.

shaobinspire: It is used to create the RGB color boxes.
```
auto red_box = new CustomIntegerBox(std::move(red_model));
auto green_box = new CustomIntegerBox(std::move(green_model));
auto blue_box = new CustomIntegerBox(std::move(blue_model));
```

kranar: I don't see that happening. Please remove these constructors.

shaobinspire: I am confused. If this constructor is removed, the compiler will show the error of `cannot convert argument 1 from 'std::shared_ptr<Spire::ScalarValueModel<boost::optional<Group>>>' to 'const anonymous-namespace'::CustomIntegerBox &'`.  
My another guess is that you want to remove the custom `IntegerBox`, but I think the custom `IntegerBox` is necessary to implement the specs that the `IntegerBox` displaying nothing can restore to the valid value when it loses focus or when the user presses Enter.

kranar: A constructor definition is only needed if it performs a non-trivial function. If it's not needed then use a default or an inherited constructor:

https://en.cppreference.com/w/cpp/language/using_declaration#Inheriting_constructors

kranar (approved):
Looks much better.

PR #1016, opened 2024-05-30
Time and sales table view

shaobinspire:


kranar (changes requested):
Large pull requests will be unconditionally rejected. Break tasks down into small units that are suitable for review and resubmit.

PR #1017, opened 2024-05-31
Hide invisible table columns

shaobinspire:
When the TableHeaderItem is invisible, hide the column in which the header item is located by setting its width to 0.

### DIFF
--- a/Applications/Spire/Source/Ui/TableHeader.cpp
+++ b/Applications/Spire/Source/Ui/TableHeader.cpp
@@ -1,4 +1,5 @@
 #include "Spire/Ui/TableHeader.hpp"
+#include <boost/signals2/shared_connection_block.hpp>
 #include <QMouseEvent>
 #include "Spire/Spire/ArrayListModel.hpp"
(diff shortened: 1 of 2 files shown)


Applications/Spire/Source/Ui/TableHeader.cpp
```diff
@@ -81,6 +82,30 @@ connection TableHeader::connect_filter_signal(
...
+    if(index != -1) {
+      auto blocker = shared_connection_block(m_widths_connection);
+      m_widths->set(index, 0);
+    }
+  } else if(event->type() == QEvent::ShowToParent) {
+    auto index = find_item_index(watched);
+    if(index != -1) {
+      auto blocker = shared_connection_block(m_widths_connection);
+      m_widths->set(index, m_item_views[index]->width());
+    }
```
kranar, line 104: Factor this out, there's no need for this much repetition.

kranar (changes requested):

kranar (approved):
Looks great.

PR #1019, opened 2024-05-31
TimeAndSalesTableModel

shaobinspire:


Applications/Spire/Source/TimeAndSales/TimeAndSalesTableModel.cpp
```diff
@@ -0,0 +1,125 @@
...
+    } else if(column == TimeAndSalesTableModel::Column::PRICE) {
+      return time_and_sale.m_price;
+    } else if(column == TimeAndSalesTableModel::Column::SIZE) {
+      return time_and_sale.m_size;
+    } else if(column == TimeAndSalesTableModel::Column::MARKET) {
+      return time_and_sale.m_marketCenter;
+    } else if(column == TimeAndSalesTableModel::Column::CONDITION) {
+      return time_and_sale.m_condition;
+    }
+    return {};
```
kranar, line 22: Dangling reference.

Applications/Spire/Source/TimeAndSales/TimeAndSalesTableModel.cpp
```diff
@@ -0,0 +1,125 @@
...
+      auto& snapshot = result.Get();
+      for(auto i = snapshot.rbegin(); i != snapshot.rend(); ++i) {
+        m_entries.push(*i);
+      }
+      m_end_loading_signal();
+    });
+}
+
+void TimeAndSalesTableModel::on_update(const TimeAndSalesModel::Entry& entry) {
+  m_entries.insert(entry, 0);
```
kranar, line 106: This is an ```O(n)``` operation, way too slow.

kranar (changes requested):

Applications/Spire/Source/TimeAndSales/TimeAndSalesTableModel.cpp
```diff
@@ -95,31 +98,17 @@ void TimeAndSalesTableModel::load_snapshot(Queries::Sequence last, int count) {
...
-      m_transaction.end();
-    },
-    [&] (const ListModel<TimeAndSalesModel::Entry>::AddOperation& operation) {
-      m_transaction.push(TableModel::AddOperation(operation.m_index));
-    },
-    [&] (const ListModel<TimeAndSalesModel::Entry>::RemoveOperation&
-        operation) {
-      m_transaction.push(TableModel::RemoveOperation(operation.m_index));
-    });
+  m_entries.push_front(entry);
```
kranar, line 112: ```std::deque``` is appropriate when you need a double ended queue, that is you're pushing and popping from the front and back. It's not suitable for this task where only insertions to one end are performed. A dynamic array is the correct data structure for a list of elements, to preserve the ```O(1)``` guarantees ensure you only push and pop to the end of the array.

kranar (changes requested):

Applications/Spire/Source/TimeAndSales/TimeAndSalesTableModel.cpp
shaobinspire: When the model queries historical data, data is pushed to the back. When the model gets real-time data, data is inserted to the front.  Hence, I think this task requires inserting data into the front and back ends, not just one end.

kranar: Querying and inserting historical data is already an ```O(N)``` operation, there is nothing that can be done to reduce it's time complexity.

shaobinspire: I am confused. For this task, data can only be inserted to the front or pushed to the back, but not to the middle. `std::deuqe` has constant O(1) for inserting elements at the end or beginning, so I think it is suitable for this case.

kranar: Come up with a solution that uses a dynamic array instead of a std::deque. A deque is not the appropriate data structure for this use case, it's too slow.

Applications/Spire/Source/TimeAndSales/TimeAndSalesTableModel.cpp
```diff
@@ -0,0 +1,119 @@
...
+  return m_transaction.connect_operation_signal(slot);
+}
+
+void TimeAndSalesTableModel::load_snapshot(Queries::Sequence last, int count) {
+  m_begin_loading_signal();
+  m_promise = m_model->query_until(last, count).then(
+    [=] (auto&& result) {
+      auto& snapshot = result.Get();
+      auto size = get_row_size();
+      m_entries.insert(m_entries.begin(),
```
kranar, line 104: How do you prevent duplicate time and sales in the table? It's possible that you load a snapshot that contains time and sales that are already represented by this model.

kranar (changes requested):
Looks a lot better.

Applications/Spire/Source/TimeAndSales/TimeAndSalesTableModel.cpp
shaobinspire: I forgot that the array of storing time and sales is sorted, so I thought `std::unordered_set` might not be needed.

kranar (approved):
Good work.

PR #1024, opened 2024-06-04
Wrap the TableView by the ScrollBox

shaobinspire:


Applications/Spire/Source/Ui/TableView.cpp
```diff
@@ -77,6 +77,7 @@ TableView::TableView(
   enclose(*box_body, *m_header_view);
   auto box = new Box(box_body);
   box->setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Preferred);
+  link(*box, *m_header_view);
```
kranar, line 80: Don't link internal UI details.

Applications/Spire/Source/TimeAndSales/TimeAndSalesTableView.cpp
```diff
@@ -36,10 +36,7 @@ namespace {
       set(BackgroundColor(Qt::transparent));
     style.get(body_selector > CurrentColumn()).
       set(BackgroundColor(Qt::transparent));
-  }
-
-  auto apply_table_header_style(StyleSheet& style) {
-    style.get(Any()).
+    style.get(Any() > (is_a<TableHeader>() < is_a<Box>())).
```
kranar, line 39: Can you explain what this is trying to do? Why can't you style the ```TableView``` directly?

kranar (changes requested):
Looks much better!

Applications/Spire/Source/TimeAndSales/TimeAndSalesTableView.cpp
shaobinspire: I want to add a border line to the `TableHeader`. I don't think it can be achieved by styling the `TableView` directly, since there is no `Box` inside the `TableHeader` to style a border, and there is no style selector for the border in the `TableView.
image

kranar: If it's not in the spec then don't add it.

shaobinspire: Okay. It's not in the spec, but in the updated specs of the table header. I will remove it.

kranar: If it's in the table header spec then don't implement it in the time and sales. Doing this spreads out logic all of the codebase and makes it very hard to follow.

Things related to the table header should be implemented in TableHeader or TableHeaderItem, or whatever suitable abstraction is directly related to that functionality, as opposed to taking things like the table and smearing it's implementation everywhere.

kranar (approved):
Perfect.

PR #1026, opened 2024-06-05
Time and sales export

shaobinspire:


Applications/Spire/Source/TimeAndSales/TimeAndSalesWindow.cpp
```diff
@@ -59,6 +63,50 @@ namespace {
     }
     return height;
   }
+
+  struct ExportTimeAndSalesTableModel : TableModel {
```
kranar, line 67: Avoid writing hacky code to work around issues, instead come up with solutions that can be reused in the future to avoid repeating the same problems and also to help improve the codebase.

Instead of investing time writing an entire model that is just intended to be a hack to add table headers, instead write a function that takes a table and a list of header names and uses that to export. That function can be used by future components that export with a header.

Code like this makes future maintenance really hard and slows down development, making it harder and harder to add new features.

This is a lot of extra work for what should just be a simple function along the lines of (pseudocode, don't take it literally):

    void export(const TableModel& model, std::vector<QString> header, std::ostream& out) {
      for(auto i = 0; i < header.size(); ++i) {
        if(i != 0) {
          out << ',';
        }
        out << header[i];
      }
      out << '\n';
      export(model, out);
    }

You can even ask ChatGPT to write this function for you, or use Visual Studio Copilot to write it for you.

kranar (changes requested):
Looks good overall, just design/API changes.

kranar (approved):

PR #1028, opened 2024-06-05
Time and sales window

shaobinspire:


Applications/Spire/Include/Spire/Ui/TableView.hpp
```diff
@@ -70,6 +70,9 @@ namespace Spire {
       /** Returns the table body. */
       TableBody& get_body();
 
+      /** Returns the scroll box. */
+      ScrollBox& get_scroll_Box();
```
kranar, line 74: Fix case.

Applications/Spire/Source/TimeAndSales/TimeAndSalesTableView.cpp
```diff
@@ -64,6 +93,15 @@ namespace {
     return model;
   }
 
+   auto make_pull_indicator() {
+    auto spinner = new QMovie(":/Icons/spinner.gif", QByteArray());
```
kranar, line 97: Why is an empty ```QByteArray()``` being passed?

Applications/Spire/Source/TimeAndSales/TimeAndSalesTableView.cpp
```diff
@@ -140,5 +178,45 @@ TableView* Spire::make_time_and_sales_table_view(
         set(TextAlign(properties[i].m_alignment));
     });
   }
+  auto pull_indicator = make_pull_indicator();
+  pull_indicator->setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Fixed);
```
kranar, line 182: Keep construction logic together, in general keep code together instead of spreading it out. The ```make_pull_indicator``` should handle as much of the construction as possible, instead of having some of it constructed in there and then some of it constructed here.

kranar (changes requested):
Looks good overall. Minor fixes needed.

kranar (approved):

PR #1029, opened 2024-06-06
Cancel the properties window

shaobinspire:


Applications/Spire/Source/TimeAndSales/TimeAndSalesPropertiesWindow.cpp
```diff
@@ -211,6 +211,10 @@ const std::shared_ptr<TimeAndSalesPropertiesModel>&
   return m_model->m_properties_model;
 }
 
+void TimeAndSalesPropertiesWindow::hideEvent(QHideEvent*) {
+  m_initial_properties = m_model->m_properties_model->get();
```
kranar, line 215: What is special about hiding this window? What if the user closes the window? What if the window is hidden for other reasons?

PR #1037, opened 2024-06-24
DataType

shaobinspire:


Applications/Spire/Include/Spire/StyleParser/DataTypes/DataType.hpp
```diff
@@ -0,0 +1,29 @@
...
+  class DataType {
+    public:
+      virtual ~DataType() = default;
+
+      /** Returns the name. */
+      const std::string& get_name() const;
+
+      bool operator ==(const DataType& rhs) const;
+
+      bool operator !=(const DataType& rhs) const;
```
kranar, line 17: I think this can be defaulted or even removed.

Applications/Spire/Include/Spire/StyleParser/DataTypes/FloatType.hpp
```diff
@@ -0,0 +1,19 @@
+#ifndef SPIRE_FLOAT_TYPE_HPP
+#define SPIRE_FLOAT_TYPE_HPP
+#include "Spire/StyleParser/DataTypes/DataType.hpp"
+
+namespace Spire {
+
+  /** Represents the floating-point type. */
+  class FloatType : public DataType {
```
kranar, line 8: Can this be further inherited from?

Applications/Spire/Source/StyleParser/DataTypes/FloatType.cpp
```diff
@@ -0,0 +1,10 @@
+#include "Spire/StyleParser/DataTypes/FloatType.hpp"
+
+using namespace Spire;
+
+FloatType::FloatType()
+  : DataType("Float") {}
+
+bool FloatType::is_equal(const DataType& rhs) const {
+  return get_name() == rhs.get_name();
```
kranar, line 9: This does not look like a valid way to test for equality.

kranar (changes requested):

Applications/Spire/Include/Spire/StyleParser/DataTypes/DataType.hpp
shaobinspire: Yes, it can be removed since "The inequality operator is automatically generated by the compiler if operator== is defined."

Referring to https://en.cppreference.com/w/cpp/language/operators

Applications/Spire/Include/Spire/StyleParser/DataTypes/FloatType.hpp
shaobinspire: I think the percent type might inherit from it.

kranar: Avoid mixing types and values. A percentage is nothing more than a number divided by 100.

kranar (approved):

PR #1043, opened 2024-07-09
Selector parser

shaobinspire:


kranar (changes requested):
Add a unit tests for non-trivial selectors such as ```a and b or c``` and ```a or b and c```. The first should parse into ```OrSelector(AndSelector(a, b), c)``` and the second should parse as ```OrSelector(a, AndSelector(b, c))```.

shaobinspire:
> Add a unit tests for non-trivial selectors such as `a and b or c` and `a or b and c`. The first should parse into `OrSelector(AndSelector(a, b), c)` and the second should parse as `OrSelector(a, AndSelector(b, c))`.

I have added two additional unit tests. The second unit test is for `a and b or c` and the third unit test is for `a or b and c`.

kranar (approved):

PR #1054, opened 2024-07-22
FontStyleBox listItem taller than spec

shaobinspire:


Applications/Spire/Source/Ui/DropDownBox.cpp
```diff
@@ -412,6 +412,7 @@ void DropDownBox::make_drop_down_list() {
   auto list_view =
     new ListView(m_list, m_current, m_selection, m_item_builder, m_to_text);
   m_drop_down_list = new DropDownList(*list_view, *this);
+  link(*this, *m_drop_down_list);
```
kranar, line 415: Can you explain why this is being linked? ```DropDownList``` is already a child of ```DropDownBox```.

shaobinspire: I think that `DropDownList` is not is a direct child of `DropDownBox` since `DropDownList` is wrapped by `OverlayPanel` and `OverlayPanel` is the child of `DropDownBox`.

I have tried to use 
`Any() > is_a<DropDownList>() > is_a<ListView>() > is_a<ListItem>()`
and
`Any() > is_a<OverlayPanel>() > is_a<DropDownList>() > is_a<ListView>() > is_a<ListItem>()` ,
but both of them don't work. 
However, `Any() >> is_a<DropDownList>() > is_a<ListView>() > is_a<ListItem>()` can access `ListItems` within the `DropDownBox`. The descendant selector can cause a crash, so I add this link to let `DropDownBox` access `DropDownList` directly.

kranar: Prefer to style the ```ListItem``` from its body then. ```Any() < is_a<ListItem>()```

kranar (approved):
After going over this task I agree with you that this is the correct solution and I will approve it. Thanks for taking the time to implement it.

PR #1056, opened 2024-07-25
Slider2d crash

shaobinspire:
The crash is caused by performing style `match/unmatch` in the style signal handler, so I moved `match/unmatch` from it.

### DIFF
--- a/Applications/Spire/Source/Ui/Slider2D.cpp
+++ b/Applications/Spire/Source/Ui/Slider2D.cpp
@@ -431,9 +431,7 @@ void Slider2D::on_thumb_icon_style() {
     property.visit(
       [&] (const IconImage& image) {
         stylist.evaluate(image, [=] (auto image) {
-          if(m_thumb_image != image) {
-            m_thumb_image = std::move(image);
-          }
+          m_thumb_image = std::move(image);
         });
       });
   }
--- a/Applications/Spire/Source/Ui/Slider2D.cpp
+++ b/Applications/Spire/Source/Ui/Slider2D.cpp
@@ -253,6 +253,24 @@ connection Slider2D::connect_submit_signal(
   return m_submit_signal.connect(slot);
 }
 
+bool Slider2D::event(QEvent* event) {
+  if(event->type() == QEvent::LayoutRequest) {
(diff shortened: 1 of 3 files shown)


Applications/Spire/Source/Ui/Slider2D.cpp
```diff
@@ -392,52 +410,31 @@ void Slider2D::on_y_current(const Decimal& y) {
...
 void Slider2D::on_thumb_icon_style() {
   auto& stylist = find_stylist(*m_thumb->get_body());
-  auto has_update = std::make_shared<bool>(false);
   for(auto& property : stylist.get_computed_block()) {
     property.visit(
       [&] (const IconImage& image) {
         stylist.evaluate(image, [=] (auto image) {
-          if(m_thumb_image == image) {
-            return;
+          if(m_thumb_image != image) {
```
kranar, line 434: Why is a comparison being performed?

Applications/Spire/Source/Ui/Slider2D.cpp
```diff
@@ -253,6 +253,24 @@ connection Slider2D::connect_submit_signal(
   return m_submit_signal.connect(slot);
 }
 
+bool Slider2D::event(QEvent* event) {
+  if(event->type() == QEvent::LayoutRequest) {
+    auto& track_stylist = find_stylist(*m_track);
```
kranar, line 258: What is the relationship between a layout request and the styling? This looks very hacky.

kranar (changes requested):

Applications/Spire/Source/Ui/Slider2D.cpp
shaobinspire: Yeah, you are correct. There is no relationship between them. If the QTimer is not allowed to used in the style signal handler, I cannot find a way to avoid calling match/unmatch within the same event as the style update, except by using QEvent::LayoutRequest or QEvent::Paint, which are emitted outside the style update event.

shaobinspire:
The crash will be resolved by the [pull request](https://github.com/spiretrading/nexus/pull/1059).

PR #1060, opened 2024-08-30
HighlightBox crashes when opening HighlightPicker

shaobinspire:


kranar (approved):

Applications/Spire/Source/Ui/OverlayPanel.cpp
```diff
@@ -169,7 +169,6 @@ void OverlayPanel::position() {
       parent_geometry.bottomLeft());
     auto screen_geometry = parentWidget()->screen()->availableGeometry();
     auto panel_size = size();
-    windowHandle()->resize(panel_size);
```
kranar, line 172: What if ```windowHandle()``` is not ```nullptr```?

kranar (changes requested):

Applications/Spire/Source/Ui/OverlayPanel.cpp
shaobinspire: I didn't notice any difference after it was removed. I vaguely remember it being used to fix display issues when the OverlayPanel appeared on multiple screens, but I haven't encountered any problems since its removal.

PR #1065, opened 2024-09-06
TableHeader update

shaobinspire:


Applications/Spire/Source/Ui/TableHeader.cpp
```diff
@@ -79,6 +81,21 @@ connection TableHeader::connect_filter_signal(
   return m_filter_signal.connect(slot);
 }
 
+bool TableHeader::eventFilter(QObject* watched, QEvent* event) {
+  if(event->type() == QEvent::HideToParent) {
+    if(auto index = get_index(static_cast<TableHeaderItem*>(watched))) {
+      auto blocker = shared_connection_block(m_widths_connection);
+      m_widths->set(*index, 0);
```
kranar, line 88: Don't treat a hidden column as having a width of 0. This will end up breaking the persistence of widths, for example if a window is closed and re-opened and the widths need to be saved.

Instead what could be possible is to have a function that returns the effective width by testing of a header item is visible, and if it is return its width, otherwise return 0.

kranar (changes requested):

PR #1076, opened 2024-09-17
Time and sales properties window

shaobinspire:


Applications/Spire/Source/TimeAndSales/TimeAndSalesPropertiesWindowFactory.cpp
```diff
@@ -0,0 +1,21 @@
+#include "Spire/TimeAndSales/TimeAndSalesPropertiesWindowFactory.hpp"
+#include "Spire/Spire/LocalValueModel.hpp"
+
+using namespace Spire;
+
+TimeAndSalesPropertiesWindowFactory::TimeAndSalesPropertiesWindowFactory()
+  : TimeAndSalesPropertiesWindowFactory(
+      std::make_shared<LocalValueModel<TimeAndSalesProperties>>()) {}
```
kranar, line 8: Use the ```typedef``` instead.

Applications/Spire/Source/TimeAndSales/TimeAndSalesPropertiesWindowFactory.cpp
```diff
@@ -0,0 +1,21 @@
+#include "Spire/TimeAndSales/TimeAndSalesPropertiesWindowFactory.hpp"
+#include "Spire/Spire/LocalValueModel.hpp"
```
kranar, line 2: This include should not be needed.

Applications/Spire/Source/Ui/FontBox.cpp
```diff
@@ -117,6 +117,10 @@ const std::shared_ptr<ValueModel<QFont>>& FontBox::get_current() const {
   return m_current;
 }
 
+void FontBox::showEvent(QShowEvent* event) {
+  m_font_family_box->setFocus();
```
kranar, line 121: What is the reason for this? I am concerned that anytime this is hidden and shown it will steal focus from anything in the entire application.

kranar (changes requested):
Good start so far.

Applications/Spire/Source/Ui/FontBox.cpp
shaobinspire: It is used to focus on the FontFamilyBox when the FontBox is shown. Since the AdaptiveBox inside the FontBox changes the layout, the tab order gets updated, causing the FontBox to lose the default input focus. You are right. This approach may steal focus from other widgets when it is shown. I'll look into finding a different solution for this.

kranar (approved):
Good work!

PR #1087, opened 2024-10-04
Replace TableView

shaobinspire:


Applications/Spire/Include/Spire/Ui/TransitionView.hpp
```diff
@@ -37,6 +37,12 @@ namespace Spire {
        */
       void set_status(Status status);
 
+      /** Replaces the body widget.
+       *  @param body The new body widget to replace.
+       *  @return The old body widget to be replaced.
+       */
+      QWidget* replace_body(QWidget& body);
```
kranar, line 44: This has the potential to result in a memory leak. Make it ```set_body``` and have it delete the previous body.

kranar (changes requested):
Looks very good, thank you.

PR #1088, opened 2024-10-17
Technicals panel

shaobinspire:


Applications/Spire/Source/BookView/TechnicalsPanel.cpp
```diff
@@ -0,0 +1,208 @@
...
+  m_default_field->get_current()->set(
+    to_default_quantity(m_bid_quantity->get(), quantity));
+}
+
+void TechnicalsPanel::on_style() {
+  auto& stylist = find_stylist(*this);
+  for(auto& property : stylist.get_computed_block()) {
+    apply(property, m_geometry, stylist);
+  }
+  m_minimum_size_hint = none;
```
kranar, line 207: I believe when the size hint changes a call to ```updateGeometry``` is also needed. Verify if that's correct.

kranar (changes requested):
Looks very good.

Applications/Spire/Source/BookView/TechnicalsPanel.cpp
shaobinspire: Yes, I think it should be called. Although the official documentation states to call this function only when sizeHint() or sizePolicy() changes, other sources suggest calling it when minimumSizeHint() changes as well. To be safe, it's better to call it in this case.

kranar (approved):

PR #1089, opened 2024-10-18
BookView

shaobinspire:
The purpose of the commit is to remove all legacy code related to BookView from the repository. I think that it should not be merged into the master branch. Instead, it will serve as the base branch for BookView, and subsequent branches related to BookView will be merged into it.


kranar:
As you indicate, there's no need to merge this into master but it can be used for work on the updated book view.

PR #1090, opened 2024-10-23
BboBox

shaobinspire:


Applications/Spire/Source/BookView/BboBox.cpp
```diff
@@ -0,0 +1,135 @@
...
+      make_to_text_model<T>(make_field_value_model(quote, field),
+        [] (const auto& value) {
+          return to_text(value);
+        }));
+    label->setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Expanding);
+    return label;
+  }
+
+  auto get_gap_width(int font_size, int parent_width) {
+    return std::clamp(0.48276 * parent_width - 6.89 * font_size,
```
kranar, line 31: Remove magic numbers here and give each constant a name. As is I have no clue what these numbers are supposed to be.

Applications/Spire/Source/BookView/BboBox.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+void BboBox::update_gap_width() {
+  auto gap_width = get_gap_width(m_font_size, width());
+  m_gap1->changeSize(gap_width, 0, QSizePolicy::Fixed);
+  m_gap2->changeSize(gap_width, 0, QSizePolicy::Fixed);
+  m_body_layout->invalidate();
+}
+
+void BboBox::on_quote(const Quote& quote) {
+  if(quote.m_price <= Money(0)) {
```
kranar, line 99: Can you explain what is special about the price being 0 or negative?

Applications/Spire/Source/BookView/BboBox.cpp
```diff
@@ -0,0 +1,135 @@
...
+    update_style(*this, [] (auto& style) {
```
kranar, line 100: Can you use a declarative approach instead of constantly changing the style?

The preferred approach is to define all your styles in one spot, and then switch between them using a ```StateSelector```. ```DecimalBox``` has selectors for ```UpTick``` and ```DownTick```:

https://github.com/spiretrading/nexus/blob/68d7f419c23d1ad4d3305c41fbab84dcbcea205c/Applications/Spire/Include/Spire/Ui/DecimalBox.hpp#L29

Applications/Spire/Source/BookView/BboBox.cpp
shaobinspire: According to the spec, the MoneyLabel, SeparatorLabel, and QuantityLabel should become invisible when MoneyLabel.current is NULL. My understanding is that the quote is considered invalid (Money is NULL) if the price is zero or negative, as there are no zero or negative quote in the real world. Therefore, in such cases, these labels should be hidden.
image

kranar: Thanks, go ahead and remove it. There is no possible way for the quote to be invalid or null.

kranar (approved):

PR #1091, opened 2024-11-05
ColorBox potimization

shaobinspire:


Applications/Spire/Source/Ui/ColorBox.cpp
```diff
@@ -190,6 +184,17 @@ void ColorBox::on_current(const QColor& current) {
 }
 
 void ColorBox::show_color_picker() {
+  if(!m_color_picker) {
+    m_color_picker = new ColorPicker(m_current, *this);
+    m_color_picker->installEventFilter(this);
+    link(*this, *m_color_picker);
+    m_color_picker_panel = m_color_picker->window();
+    m_color_picker_panel->installEventFilter(this);
+    auto input_boxes = get_input_boxes(*m_color_picker);
```
kranar, line 193: Remove the variable ```input_boxes```.

Applications/Spire/Source/Ui/ColorBox.cpp
```diff
@@ -190,6 +184,17 @@ void ColorBox::on_current(const QColor& current) {
...
+    for(auto input_box : input_boxes) {
```
kranar, line 194: ```auto& input_box```

Applications/Spire/Include/Spire/Ui/ColorCodePanel.hpp
```diff
@@ -69,10 +68,13 @@ namespace Styles {
       DropDownBox* m_color_format_box;
       QStackedWidget* m_color_input;
       PercentBox* m_alpha_box;
+      bool m_is_alpha_visibile;
```
kranar, line 71: Typo.

Applications/Spire/Source/Ui/ColorCodePanel.cpp
```diff
@@ -525,3 +518,23 @@ void ColorCodePanel::on_mode_current(const optional<int>& current) {
...
   }
 }
+
+void ColorCodePanel::on_alpha_style() {
+  auto has_update = std::make_shared<bool>(false);
+  auto& stylist = find_stylist(*m_alpha_box);
+  if(auto visibility = Styles::find<Visibility>(stylist.get_computed_block())) {
+    stylist.evaluate(*visibility, [=] (auto visibility) {
+      *has_update = true;
+      if(visibility == Visibility::VISIBLE) {
```
kranar, line 528: m_is_alpha_visible = visibility == Visibility::VISIBLE;

Applications/Spire/Source/Ui/ColorBox.cpp
shaobinspire: Since `input_boxes` stores a list of pointers of input boxes, I don't think `auto` can lead to the unnecessary copy overhead. Furthermore, we don't need to modify the pointer itself. So what is the reason for changing `auto` to `auto&`?

kranar: Use ```auto``` when you need a copy of the value, for example you want to modify a copy of the value. If you just need to view an existing value then a reference is more constrained than a value.

PR #1094, opened 2024-11-13
HighlightBox optimization

shaobinspire:


kranar (approved):
Stuff like this really improves the user experience.

PR #1096, opened 2024-11-18
Market highlights TableView

shaobinspire:


Applications/Spire/Include/Spire/BookView/BookViewProperties.hpp
```diff
@@ -21,6 +24,80 @@ namespace Spire {
...
+
+      /** The color to highlight the market with. */
+      HighlightColor m_color;
+
+      /** The highlight level. */
+      MarketHighlightLevel m_level;
+    };
+
+    /** The number of the order state. */
+    static const auto ORDER_STATE_COUNT = 5;
```
kranar, line 87: ```ORDER_HIGHLIGHT_STATE_COUNT```

Declare it next to the enum it's counting to keep things local.

Applications/Spire/Source/BookView/MarketHighlightsTableView.cpp
```diff
@@ -0,0 +1,477 @@
...
+    TableHeaderItem::Order::UNORDERED, TableFilter::Filter::NONE);
+  table_builder.add_header_item(QObject::tr("Color"), QObject::tr("Clr"),
+    TableHeaderItem::Order::UNORDERED, TableFilter::Filter::NONE);
+  table_builder.set_item_builder(std::bind_front(make_item, std::move(table),
+    markets));
+  table_builder.set_comparator([=] (const AnyRef& left, int left_row,
+      const AnyRef& right, int right_row, int column) {
+    auto& left_market = any_cast<MarketCode>(left);
+    auto& right_market = any_cast<MarketCode>(right);
+    if(left_market.IsEmpty() || right_market.IsEmpty()) {
```
kranar, line 463: This violates the weak ordering property of a comparator. The empty market has to either be the smallest or the largest but it can't be both.

Applications/Spire/Source/Ui/ListView.cpp
```diff
@@ -452,9 +452,6 @@ void ListView::pre_remove_item(int index) {
     --item->m_index;
   }
   item->m_click_observer = none;
-  QTimer::singleShot(0, [item = std::move(item)] () mutable {
```
kranar, line 455: Any insight for why this was removed?

kranar (changes requested):
Looks great.

Applications/Spire/Source/Ui/ListView.cpp
shaobinspire: This code causes a crash when an item is removed. Since the item is a `std::unique_ptr` and has already been removed from `m_items`, it will automatically be released when the function exits. Therefore, I think assigning it to null in the timeout handler is redundant and unnecessary. Furthermore, the memory previously used by the item may be reused and assigned new values for other resource. When the timer expires, the reassigned memory might cause `is_initialized` of `m_click_observer` inside the item to be parsed to true. This could lead to an attempt to destroy the click observer again, resulting in a crash due to releasing an invalid resource.

Applications/Spire/Include/Spire/BookView/BookViewProperties.hpp
```diff
@@ -70,6 +70,9 @@ namespace Spire {
       REJECTED
     };
 
+    /** The number of the order state. */
+    static const auto ORDER_STATE_COUNT = 5;
```
kranar, line 74: The name must match exactly, ```ORDER_HIGHLIGHT_STATE_COUNT```.

kranar (approved):

PR #1103, opened 2025-01-27
Book view table view

shaobinspire:


Applications/Spire/Include/Spire/BookView/BookViewModel.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+  /** The model for the book view. */
+  class BookViewModel {
+    public:
+
+      /** Represents the user order. */
+      struct UserOrder {
+
+        /** The order destination. */
+        std::string m_destination;
```
kranar, line 16: ```std::string``` -> ```Nexus::Destination```

Applications/Spire/Include/Spire/BookView/BookViewTableModel.hpp
```diff
@@ -0,0 +1,26 @@
...
+  enum class BookViewColumns {
+
+    /** The MPID column. */
+    MPID,
+
+    /** The price column. */
+    PRICE,
+
+    /** The size column. */
+    SIZE,
```
kranar, line 18: Remove trailing comma.

kranar (changes requested):
This looks very good but it can't be submitted in one pull request.

Break this up into multiple branches, some of these are bug fixes that should be merged into master, while others are directly related to the book view. Pull out the bug fixes that should go into master and submit a separate pull request for those.

The reason we need to do it this way is so that the bug fixes can be released without depending on the entire book view as a whole. Other features can benefit from your bug fixes right away, while the book view as a whole won't get merged into master for a very long time (it will remain part of the preview branch).

Applications/Spire/Source/BookView/BookViewTableView.cpp
```diff
@@ -0,0 +1,1139 @@
...
+  const auto& get_price(const TableModel& table, int row) {
+    return table.get<Money>(row, static_cast<int>(BookViewColumns::PRICE));
+  }
+
+  const auto& get_size(const TableModel& table, int row) {
+    return table.get<Quantity>(row, static_cast<int>(BookViewColumns::SIZE));
+  }
+
+  auto is_order(const std::string& mpid) {
+    return mpid.front() == '@';
```
kranar, line 142: Undefined behavior if ```mpid.empty()```

kranar (changes requested):

kranar (approved):

PR #1128, opened 2025-03-12
RegionBox within KeyBindingsWindow navigation 

shaobinspire:


kranar (commented):
As a first attempt this is good, but unfortunately it is hacky since it makes too many assumptions specific to RegionBox. If any other TableView uses a RegionBox or any other component has a scrollbox then this issue will reappear.

For that reason I'd like you to look at a solution that can work for any component so we don't have to always worry about this issue.

Applications/Spire/Source/Ui/TableBody.cpp
```diff
@@ -1469,3 +1473,16 @@ void TableBody::on_widths_update(const ListModel<int>::Operation& operation) {
       }
     });
 }
+
+bool TableBody::on_key_press(QWidget& target, QKeyEvent& event) {
+  if(typeid(target) == typeid(ScrollBox)) {
```
kranar, line 1478: ```typeid``` can't test polymorphic types, instead ```dynamic_cast``` handles them.

We'll have to meet this Friday to go over this, I think this might need an update to the table specifications, but so far it's good.

kranar (approved):

PR #1130, opened 2025-03-14
Table view tab order

shaobinspire:


shaobinspire:
This update resolved [The TableView can’t navigate to the last column using the Tab key]( https://app.asana.com/1/1124513426775746/project/1198733135213649/task/1209662415204188?focus=true) and [Keybindings Cell[0][0] Focus Trap](https://app.asana.com/1/1124513426775746/project/1198733135213649/task/1207881833396298?focus=true).

Applications/Spire/Include/Spire/Ui/TableBody.hpp
```diff
@@ -129,6 +129,7 @@ namespace Styles {
       void paintEvent(QPaintEvent* event) override;
       void resizeEvent(QResizeEvent* event) override;
       void showEvent(QShowEvent* event) override;
+      void focusInEvent(QFocusEvent* event) override;
```
kranar, line 132: Alphabetical order.

Applications/Spire/Source/Ui/TableBody.cpp
```diff
@@ -142,7 +142,9 @@ struct TableBody::RowCover : Cover {
   void mount(int index) {
     auto& body = *static_cast<TableBody*>(parentWidget());
     for(auto i = 0; i != layout()->count(); ++i) {
-      get_item(i)->mount(*body.m_item_builder.mount(body.m_table, index, i));
+      auto item = body.m_item_builder.mount(body.m_table, index, i);
+      item->setFocusPolicy(Qt::NoFocus);
```
kranar, line 146: Can you explain why the focus policy on the mounted widget is turned off?

Applications/Spire/Source/Ui/TableBody.cpp
```diff
@@ -1227,42 +1246,72 @@ void TableBody::update_column_widths() {
 }
 
 bool TableBody::navigate_next() {
+  auto get_next_column =
+    [&] (const CurrentModel& current, int row, int column) {
+      ++column;
+      while(column < get_table()->get_column_size()) {
```
kranar, line 1252: Factor this out, there's no need for a ```break```.

Applications/Spire/Source/Ui/TableBody.cpp
```diff
@@ -1227,42 +1246,72 @@ void TableBody::update_column_widths() {
...
   } else {
     return false;
   }
   return focus_current(*get_current(), *this, Qt::TabFocusReason);
 }
 
 bool TableBody::navigate_previous() {
+  auto get_previous_column =
+    [] (const CurrentModel& current, int row, int column) {
+      --column;
```
kranar, line 1286: See above.

Applications/Spire/Source/Ui/TableView.cpp
```diff
@@ -164,6 +166,7 @@ TableView::TableView(
   m_body_style_connection = connect_style_signal(
     *m_body, std::bind_front(&TableView::on_body_style, this));
   setFocusProxy(m_body);
+  m_body->setFocusPolicy(Qt::StrongFocus);
```
kranar, line 169: Why is the focus policy set here?

Applications/Spire/Source/Ui/TableView.cpp
```diff
@@ -128,6 +128,7 @@ TableView::TableView(
   m_header_scroll_box->set(ScrollBox::DisplayPolicy::NEVER);
   m_header_scroll_box->setSizePolicy(
     QSizePolicy::Expanding, QSizePolicy::Fixed);
+  m_header_scroll_box->setFocusPolicy(Qt::NoFocus);
```
kranar, line 131: Can you describe why the scroll box's focus policy is set to no focus here? What makes TableView special that the focus policy has to be set here?

kranar (changes requested):

Applications/Spire/Source/Ui/TableView.cpp
shaobinspire: I don’t think ScrollBoxes within TableView need a strong focus policy. When navigating into or out of TableView using Tab or Shift+Tab, ScrollBox which has a strong focus policy (by default) can receive focus, but since it lacks a visual indicator for its focus state, this can be confusing for users. Moreover, TableView doesn’t require ScrollBox to achieve navigation function via the keyboard. Therefore, I set all ScrollBoxes to have no focus policy to avoid them getting focus.

Applications/Spire/Source/Ui/TableView.cpp
shaobinspire: Since the focus policy of ScrollBox is set to no focus policy and TableBody is its focus proxy, TableBody is also set to no focus policy. My intention is to clear ScrollBox’s foucs policy while retaining the focus policy of TableBody. Therefore, TableBody needs to be restored to a strong focus policy.

Applications/Spire/Source/Ui/TableBody.cpp
shaobinspire: Cell navigation inside TableBody does not rely on the cell's focus policy or Qt's tab order list. Instead, it is handled by the focusNextPrevChild function, which takes charge of navigation within TableBody. If a cell's focus policy is set to TabFocus or StrongFocus, Qt's default behavior may interfere with TableBody’s navigation, leading to an inconsistent tab order.

kranar: I am not sure I understand, perhaps you can help explain it to me. The item builder returns a ```QWidget``` called ```item```. ```item``` then has its focus policy set to ```NoFocus```. ```item``` is then mounted into a ```TableItem```. ```TableItem``` has a focus policy of ```StrongFocus```. When ```TableItem``` mounts ```item```, it also sets its focus proxy to ```item```, which then sets ```item```'s focus policy to ```StrongFocus```, which undoes the setting of ```item```'s focus policy to ```NoFocus```. This then makes setting the focus policy here redundant.

Did I make a mistake or misunderstand something?

Applications/Spire/Source/Ui/TableView.cpp
kranar: I am concerned about these inconsistent focus policies where ```A``` has a focus proxy of ```B```, but ```A```'s focus policy is not the same as ```B```'s focus policy. As far as I know this is an inconsistency and should never be the case.

Can you provide a specific use case where this is needed to avoid an issue?

Applications/Spire/Source/Ui/TableBody.cpp
shaobinspire: Good question. When `TableItem` mounts `item`, it doesn’t undo the the setting of `item`'s focus policy. This is because `TableItem::mount` only sets `item` as its focus proxy using`setFocusProxy(proxy)`, which does not automatically change the focus policy of `item`. 

I think a better solution is to set `TableItem` itself to `NoFocus` rather than `item`. I updated the code to set `TableItem` to `NoFocus` and adjust the focus policy when `TableItem::mount` a body, ensuring the consistency between TableItem and its body.

Applications/Spire/Source/Ui/TableView.cpp
shaobinspire: I couldn’t find a use case from standard `Qt` widgets, but `ChatGPT` provided a specific case with `QScrollArea`, where the scroll area itself might not need to receive focus, but its contained widget does. This is similar to our case.

However, since this is not the normal approach, we should avoid achieving the functionality this way. Instead, I updated the approach by setting the focus proxy of `TableView` to `ScrollBox` instead of `TableBody`. This ensures that the deepest focus proxy of `TableView` remains `TableBody`, while avoiding inconsistent focus policy settings.

Additionally, I updated `TableBody::focusNextPrevChild`, which now uses `QWidget::focusNextPrevChild` to determine which widget outside `TableView` should receive focus when navigating out of `TableView`.

kranar (approved):

PR #1131, opened 2025-03-20
Book view orders

shaobinspire:


Applications/Spire/Source/BookView/BookViewTableView.cpp
```diff
@@ -623,14 +629,14 @@ namespace {
         std::shared_ptr<RowTracker> row_tracker,
         std::shared_ptr<ValueModel<OrderVisibility>> visibility_property,
         std::shared_ptr<ValueModel<OrderHighlightArray>> highlight_properties,
-        bool is_ascending, QObject* timer_owner)
+        bool is_ascending, QObject& timer_owner)
```
kranar, line 632: What's a ```timer_owner```? What timer is this referring to? Can you use a ```QTimer```?

Applications/Spire/Source/BookView/BookViewTableView.cpp
```diff
@@ -990,7 +986,113 @@ namespace {
...
+
+    void transact(const std::function<void()>& transaction) override {
+      m_transaction.transact(transaction);
+    }
+
+    void on_book_quote_operation(
+        const ListModel<BookQuote>::Operation& operation) {
+      visit(operation,
+        [&] (const auto& operation) {
+          m_transaction.push(operation);
```
kranar, line 1046: The ```visit``` isn't doing anything.

Applications/Spire/Source/BookView/BookViewTableView.cpp
```diff
@@ -990,7 +986,113 @@ namespace {
...
+        [&] (const ListModel<UserOrder>::AddOperation& operation) {
+          m_order_quotes.insert(m_order_quotes.begin() + operation.m_index,
+            make_order_book_quote(m_user_orders->get(operation.m_index),
+              m_side));
+          m_transaction.push(AddOperation(
+            m_book_quotes->get_size() + operation.m_index));
+        },
+        [&] (const ListModel<UserOrder>::PreRemoveOperation& operation) {
+          QTimer::singleShot(ORDER_HIGHLIGHT_DELAY_MS, m_timer_owner,
+            [=, order = m_user_orders->get(operation.m_index)] {
```
kranar, line 1062: Consider the following series of operations that all happen over a time duration that is less than ```ORDER_HIGHLIGHT_DELAY_MS```:

    AddOperation[value = "Foo", m_index = 0];
    AddOperation[value = "Bar", m_index = 1];
    PreRemoveOperation[value = "Foo", m_index = 0];
    UpdateOperation[value = "Bar", m_index = 0];
    
This implementation will end up handling the ```UpdateOperation``` by updating ```[value = "Foo"]``` instead of the correct ```[value = "Bar"]```.

shaobinspire: This is the expected behavior. The purpose of the code is to delay the deletion of a quote so that the `TableView `can fully display animation.

```
AddOperation[value = "Foo", m_index = 0] -> Quote["Foo"] is added to the model
AddOperation[value = "Bar", m_index = 1] -> Quote["Bar"] is added to the model
PreRemoveOperation[value = "Foo", m_index = 0] -> Do nothing, waiting for the timeout to remove Quote["Foo"] from the model
UpdateOperation[value = "Bar", m_index = 0] -> Quote["Bar"] is updated
...
Timeout -> Find Quote["Foo"] based on MPID and price. If the quantity of Quote["Foo"] is less than or equal to 0, remove it from the model
```

The `UpdateOperation[value = "Bar", m_index = 0]` does not affect the deletion of `Quote["Foo"]`.

Another scenario may occur when an `AddOperation` for "Foo" is triggered during `ORDER_HIGHLIGHT_DELAY_MS`:

```
AddOperation[value = "Foo", m_index = 0] -> Quote["Foo"] is added to the model
AddOperation[value = "Bar", m_index = 1] -> Quote["Bar"] is added to the model
PreRemoveOperation[value = "Foo", m_index = 0] -> Do nothing, waiting for the timeout to remove Quote["Foo"] from the model
AddOperation[value = "Foo", m_index = 1] -> The quantity of Quote["Foo"] is updated since Quote["Foo"] has not been removed.
...
Timeout -> Find Quote["Foo"] based on MPID and price. Since the quantity of Quote["Foo"] is greater than 0, do not remove it from the model
```

In this case, when the timeout occurs, a newly added `"Foo"` has updated the quantity of `Quote["Foo"]`. If the updated quantity is greater than 0, `Quote["Foo"]` is not removed. The current code doesn't deal with this case, so I think it should be updated.

Applications/Spire/Source/BookView/BookViewTableView.cpp
shaobinspire: `timer_owner` is used to prevent handling table data if the table is deleted before the timer expires. I could replace it with `QTimer`, but that would be more verbose.

shaobinspire: Each terminal order status, such as order cancellation, requires a simulation display delay, so multiple timers may be needed when multiple rows are being cancelled. Therefore, I don't think a single `QTimer` can handle this case. `QTimer::singleShot` can easily create a timer delay for each cancellation.

Applications/Spire/Source/BookView/BookViewTableView.cpp
kranar: The ```visit``` is not needed, please remove it.

Applications/Spire/Source/BookView/BookViewTableView.cpp
kranar: Prefer writing code that is local, that is does not require looking at multiple disjoint parts of the codebase in order to understand. In this case rather than passing in a shared object whose ownership is unclear and could potentially be global, use a local object or have this class inherit from ```QObject```.

Applications/Spire/Source/BookView/BookViewTableView.cpp
kranar: This will update the wrong quote. In the original example the quote that needs to be updated is ```[value = "Bar"]```, not ```[value = "Foo"]```.

kranar (approved):

PR #1143, opened 2025-03-27
Region box scroll to end

shaobinspire:


Applications/Spire/Source/Ui/RegionBox.cpp
```diff
@@ -169,6 +169,7 @@ void RegionBox::on_current(const Region& region) {
       current->push(regions[i]);
       ++i;
     }
+    sort(*current);
```
kranar, line 172: In general try to minimize the number of operations that propagate outward from the system. Sorting ends up signaling on the order of ```O(nlogn)``` signals which everything keeping track of needs to record in order to remain synchronized. Instead do whatever work is needed to prepare an update, and then publish that update with the minimum number of operations.

kranar (approved):
There is an issue with this but we can fix it later due to time constraints.

PR #1156, opened 2025-04-11
Book view window

shaobinspire:


kranar:
There's a lot of good ideas in here but unfortunately the implementation is too complex and unstructured to be approved. It is no longer permissible to implement functionality by writing incredibly large, tightly integrated and complex chunks of unstructured code. It becomes too hard to maintain and keep track of or depend on.

Functionality must be written using composable units, each of which can be independently reasoned about, modified and tested. You are required to take the designs presented to you and identify the appropriate tools to take those designs, break them down into modular components and implement each component piece by piece, submitting them for review as you go.

It is simply not possible to review files that are thousands upon thousands of lines of code and it won't be possible for me or anyone else to work with this in the future if we need to add new features, fix bugs, or otherwise work with this.

With that said, there is good stuff in here and hopefully you'll be able to take some of these ideas and find a more suitable way to implement them in the future, but as is I can't accept this code.

PR #1185, opened 2025-05-14
DecimalBox autorepeat mouse press

shaobinspire:


kranar (commented):
The implementation looks good, resubmit after making feature changes.

Applications/Spire/Source/Ui/DecimalBox.cpp
```diff
@@ -529,6 +549,35 @@ void DecimalBox::keyPressEvent(QKeyEvent* event) {
   QWidget::keyPressEvent(event);
 }
 
+void DecimalBox::mousePressEvent(QMouseEvent* event) {
+  if(m_data && event->button() == Qt::LeftButton) {
+    event->accept();
+    if(m_data->m_up_button->geometry().contains(event->pos())) {
+      m_tick = TickIndicator::UP;
```
kranar, line 556: Can you explain why ```m_tick``` is set to ```TickIndicator::UP```? ```m_tick``` is intended strictly for style selection.

shaobinspire: I initially intended to use it as flags to indicate which button was being held during the timerEvent, but it turned out to be unnecessary. I'll remove it.

kranar (changes requested):
As usual there is a lot of good stuff in here but I want to move away from bundling all sorts of functionality that gets smeared all over piece by piece. It makes it too hard to reason about this.

If you can isolate the functionality and encapsulate it into its own component instead of spreading the functionality all over that would be great.

shaobinspire:
> As usual there is a lot of good stuff in here but I want to move away from bundling all sorts of functionality that gets smeared all over piece by piece. It makes it too hard to reason about this.
> 
> If you can isolate the functionality and encapsulate it into its own component instead of spreading the functionality all over that would be great.

I will try to encapsulate it into a component.

PR #1188, opened 2025-05-15
DecimalBox autorepeat mouse press

shaobinspire:
I have encapsulated the functionality into the internal component `StepButton`.

### DIFF
--- a/Applications/Spire/Source/Ui/DecimalBox.cpp
+++ b/Applications/Spire/Source/Ui/DecimalBox.cpp
@@ -19,6 +19,9 @@ using namespace Spire;
 using namespace Spire::Styles;
 
 namespace {
+  const auto REPEAT_DELAY_TIMEOUT_MS = 500;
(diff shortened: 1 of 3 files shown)


Applications/Spire/Source/Ui/DecimalBox.cpp
```diff
@@ -380,6 +356,108 @@ struct DecimalBox::DecimalToTextModel : TextModel {
...
+        reset();
+      }
+    } else {
+      QWidget::timerEvent(event);
+    }
+  }
+
+  void reset() {
+    if(is_match(*this, Press())) {
+      unmatch(*this, Press());
```
kranar, line 442: Unmatch unconditionally, it's faster than checking for a match and then removing it.

kranar (commented):
This is so much better, all the functionality is now properly encapsulated and no more non-local behavior.

PR #1203, opened 2025-06-26
Playback control panel

shaobinspire:


Applications/Spire/Source/Playback/PlaybackControlPanel.cpp
```diff
@@ -0,0 +1,425 @@
...
+      if(!value || get_minimum() && value < get_minimum() ||
+          value > get_maximum()) {
+        return QValidator::Invalid;
+      }
+      return QValidator::Acceptable;
+    }
+
+    QValidator::State set(const optional<date>& value) override {
+      m_state = test(value);
+      if(m_state == QValidator::Invalid) {
```
kranar, line 86: A model can never be in an invalid state. It can only be in an intermediate or valid state.

Applications/Spire/Source/Playback/PlaybackControlPanel.cpp
```diff
@@ -0,0 +1,425 @@
...
+    tooltip.set_label(tr("Pause (Space)"));
+    play();
+  } else {
+    tooltip.set_label(tr("Play (Space)"));
+    pause();
+  }
+}
+
+void PlaybackControlPanel::on_jump_to_start_click() {
+  QTimer::singleShot(0, this, [=] {
```
kranar, line 377: Why is this done asynchronously?

kranar (changes requested):
Looks very good.

PR #1301, opened 2026-04-20
SwitchButton

shaobinspire:


Applications/Spire/Source/Ui/SwitchButton.cpp
```diff
@@ -0,0 +1,214 @@
...
+using namespace boost::posix_time;
+using namespace boost::signals2;
+using namespace Spire;
+using namespace Spire::Styles;
+
+namespace {
+  using Switch = StateSelector<void, struct SwitchTag>;
+  using Track = StateSelector<void, struct TrackTag>;
+
+  const auto ANIMATION_DURATION_MS = 200;
```
kranar: Avoid encoding units into the name, and instead prefer encoding the units into the value directly.

PR #1305, opened 2026-04-21
OrderStatusListBox

shaobinspire:


Applications/Spire/Source/Ui/OrderStatusListBox.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include <map>
+#include "Spire/Spire/ArrayListModel.hpp"
+#include "Spire/Ui/CustomQtVariants.hpp"
+#include "Spire/Ui/ListView.hpp"
+
+using namespace Nexus;
+using namespace Spire;
+
+namespace {
+  template<typename T>
```
kranar: Why is this a template? How do you know it works correctly?

shaobinspire: For `OrderStatusListBox`, a single key may map to multiple values. For example, the key New may map to both `OrderStatus::New` and `OrderStatus::PENDING_NEW`.

To support this pattern, `MultiValueQueryModel` is designed as a reusable template for similar scenarios in the future. One option is to place it into a separate file and provide unit tests for it. Alternatively, a concrete class can be implemented specifically for `OrderStatus`.

shaobinspire: I have removed `MultiValueQueryModel` and selected a different approach for the `OrderStatus` query.

kranar (approved):

PR #1306, opened 2026-04-24
OrderStatusFilterPanel

shaobinspire:


Applications/Spire/Source/Ui/OrderStatusFilterPanel.cpp
```diff
@@ -0,0 +1,211 @@
...
+    OrderStatus::SUSPENDED, OrderStatus::STOPPED, OrderStatus::PENDING_CANCEL,
+    OrderStatus::CANCEL_REJECT};
+
+  constexpr auto TERMINAL_STATUSES = std::array{
+    OrderStatus::REJECTED, OrderStatus::EXPIRED, OrderStatus::CANCELED,
+    OrderStatus::FILLED, OrderStatus::DONE_FOR_DAY};
+}
+
+struct OrderStatusFilterPanel::PresetButtonContainer : QWidget {
+  std::shared_ptr<AssociativeValueModel<Preset>> m_preset;
```
kranar: Who is sharing this?

Applications/Spire/Source/Ui/TagBox.cpp
```diff
@@ -333,6 +333,16 @@ bool TagBox::event(QEvent* event) {
 void TagBox::changeEvent(QEvent* event) {
   if(event->type() == QEvent::EnabledChange) {
     update_tags_read_only();
+    for(auto i = 0; i < m_list_view->get_list()->get_size(); ++i) {
+      auto& item = *m_list_view->get_list_item(i);
+      if(item.is_mounted()) {
+        if(auto parent = item.parentWidget()) {
+          parent->resize(0, 0);
```
kranar, line 340: What is this code doing and can you find a way to implement this component without modifying ```TagBox```?

shaobinspire: This is a hack that forces the `ListView` to update its layout when the Tag changes its read-only styles. It attempts to fix the following issue:

> Changing tags does not consistently adjust the layout of TagComboBox. When switching between Live, Terminal, and Custom, the layout of TagComboBox does not always update correctly, causing some tags in Custom mode to become truncated with ellipses instead of reflowing to fit.
> 
> To reproduce:
> 
> 1. Set the width to 650px
> 2. Click Live
> 3. Click Terminal
> 4. Click Custom
> 5. The tags in Custom mode may become truncated with ellipses instead of reflowing to fit. If the issue does not occur, repeat steps 2–4 until it does.
> 6. Expected: The tags should consistently fit their contents and reflow properly inside TagComboBox.

The current `ListView` implementation cannot automatically adjust its layout when a `ListItem's` layout changes. Lines 402–409 use the same hack to update `ListView` layout when the Tag label content changes.

I spent some time investigating possible solutions, but all of them require modifying `ListView`. None of them perform well. One of the simplest approaches would be to expose an `update_layout()` API from `ListView` so that `TagBox` can manually trigger a layout update. However, I do not think this is an appropriate solution, as it goes against Qt’s layout design.

I think the fundamental solution would be to implement a custom `QLayout` and leverage Qt’s layout system to update layouts automatically. However, this would be a much larger change and should be addressed in a separate task.

kranar: Propose a solution to ```ListView``` itself on how the layout should be updated in response to changes to ```ListItem```s.

PR #1311, opened 2026-05-06
AccountBox

shaobinspire:


Applications/Spire/Source/UiViewer/StandardUiProfiles.cpp
```diff
@@ -227,6 +229,57 @@ namespace {
...
+    auto height = size.height() / 5.0;
+    for(auto& [row, col] : cells) {
+      painter.fillRect(QRectF(col * width, row * height, width, height), color);
+    }
+    return identicon;
+  }
+
+  std::shared_ptr<AccountQueryModel> populate_account_query_model() {
+    auto accounts = std::vector<AccountListItem::Account>();
+    accounts.push_back({make_identicon(QColor(0xB565BC),
```
kranar: How are people supposed to come up with these cells? Why are cells arbitrary series of numbers?

shaobinspire: I noticed there is an `identicon` in the portfolio specs, and I think it is intended to generate the identicon in the actual project. The cells here are generated using random numbers for demonstration purposes. If needed, I can create a function that generates the cells based on the account ID instead.

kranar: Sounds good.

shaobinspire: Updated the `identicon` generator in UiViewer to use the newly implemented `Spire::make_identicon`, which generates `identicons` based on the account ID instead of using arbitrary icons.

kranar (approved):

PR #1320, opened 2026-05-28
Split TagBox into AnyTagBox base and TextBox-backed TagBox.

shaobinspire:
Add a `AnyTagBox` to accept `AnyInputBox` as its input widget, while `TagBox` inherits from `AnyTagBox` accepts `TextBox` as the input. This update is used to implement `TagListBox`, which inherits from `AnyTagBox` and uses `KeyInputBox` as its input widget.

### DIFF
--- a/Applications/Spire/Include/Spire/Ui/TagBox.hpp
+++ b/Applications/Spire/Include/Spire/Ui/TagBox.hpp
@@ -42,10 +42,10 @@ namespace Spire {
       const AnyRef& get_submission() const;
 
       /** Returns the input box. */
-      AnyInputBox& get_input_box();
+      const AnyInputBox& get_input_box() const;
 
       /** Returns the input box. */
-      const AnyInputBox& get_input_box() const;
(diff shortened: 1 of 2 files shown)


kranar (changes requested):
API for ```AnyTagBox``` is rejected on the grounds that it does not satisfy the requirements of fan ```AnyInputBox```.

Applications/Spire/Include/Spire/Ui/TagBox.hpp
```diff
@@ -1,51 +1,61 @@
...
 
-      /** Sets the placeholder value. */
+      /** Returns the last submission. */
+      const AnyRef& get_submission() const;
+
+      /** Returns the input box. */
+      AnyInputBox& get_input_box();
+
+      /** Returns the input box. */
+      const AnyInputBox& get_input_box() const;
```
kranar: Put const before non-const/immutable before mutable.

kranar (changes requested):

PR #1322, opened 2026-06-01
KeyListBox

shaobinspire:
The component is implemented as part of the [`KeyFilterPanel`](https://app.asana.com/1/1124513426775746/project/1198733135213649/task/1204623371368721?focus=true).

### DIFF
--- a/Applications/Spire/Source/Ui/KeyListBox.cpp
+++ b/Applications/Spire/Source/Ui/KeyListBox.cpp
@@ -1,15 +1,26 @@
 #include "Spire/Ui/KeyListBox.hpp"
-#include <unordered_set>
-#include "Spire/Spire/ArrayListModel.hpp"
-#include "Spire/Spire/HashQtTypes.hpp"
+#include "Spire/Spire/ValidatedValueModel.hpp"
 
 using namespace boost::signals2;
 using namespace Spire;
 using namespace Spire::Styles;
 
 namespace {
-  auto make_key_input_box(std::shared_ptr<KeySequenceValueModel> current) {
-    auto key_input_box = new KeyInputBox(std::move(current));
+  auto make_key_input_box(std::shared_ptr<KeySequenceListModel> keys,
+      std::shared_ptr<KeySequenceValueModel> current) {
+    auto validated = make_validated_value_model(
+      [keys = std::move(keys)] (const QKeySequence& value) {
+        if(value.isEmpty()) {
+          return QValidator::Acceptable;
+        }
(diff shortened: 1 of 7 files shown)


Applications/Spire/Source/Ui/KeyListBox.cpp
```diff
@@ -0,0 +1,148 @@
...
+    update_style(*key_input_box, [] (auto& style) {
+      style.get(Any() || Hover() || Focus() || Disabled()).
+        set(BackgroundColor(QColor(Qt::transparent))).
+        set(border_size(0)).
+        set(padding(0));
+    });
+    return new AnyInputBox(*key_input_box);
+  }
+
+  struct UniqueKeySequenceListModel : KeySequenceListModel {
```
kranar: Is this just a copy and paste of something?

shaobinspire: Indeed, there was some duplicated logic in `UniqueKeySequenceListModel`. 
I updated the implementation to enforce key uniqueness through the input box's value model using `make_validated_value_model` and removed `UniqueKeySequenceListModel`.

However, this introduces a behavior change. When the user presses a key that is already present in the list, `keyPressEvent` now ignores the input, and the key sequence never appears in the input box. Previously, the key sequence would appear in the input box but would be silently rejected and removed when Enter was pressed.

The new behavior is consistent with `KeyInputBox`, where invalid keys are prevented from being entered in the first place.

Applications/Spire/Source/Ui/KeyListBox.cpp
```diff
@@ -0,0 +1,83 @@
...
+    [&] (const KeySequenceListModel::AddOperation&) {
+      update_prompt_visibility();
+    },
+    [&] (const KeySequenceListModel::RemoveOperation&) {
+      update_prompt_visibility();
+    });
+}
+
+void KeyListBox::update_prompt_visibility() {
+  auto visibility = get_keys()->get_size() == 0 ?
```
kranar: Don't use ternary operators.

Applications/Spire/Source/Ui/KeyListBox.cpp
```diff
@@ -0,0 +1,83 @@
...
+    Visibility::VISIBLE : Visibility::INVISIBLE;
+  update_style(get_input_box(), [visibility] (auto& style) {
```
kranar: Capture by value ```[=]```

kranar (changes requested):

PR #1326, opened 2026-06-05
AccountFilterPanel

shaobinspire:


Applications/Spire/Source/Ui/AccountFilterPanel.cpp
```diff
@@ -0,0 +1,52 @@
+#include "Spire/Ui/AccountFilterPanel.hpp"
+#include "Spire/Spire/ArrayListModel.hpp"
+#include "Spire/Spire/ListModel.hpp"
+
+using namespace boost::signals2;
+using namespace Spire;
+
+bool OpenFilterPanelAdaptor<AccountListBox>::is_empty(
+    AccountListBox& account_list_box) {
+  return account_list_box.get_current()->get_size() == 0;
```
kranar: None of this looks specific to an ```AccountListBox```. A specialization should only be written for types in which the type has something specific that needs to be customized.

Applications/Spire/Source/UiViewer/StandardUiProfiles.cpp
```diff
@@ -796,7 +797,15 @@ namespace {
         [=] (const auto& submission, auto mode) {
           auto items = QString();
           for(auto i = 0; i < submission->get_size(); ++i) {
-            items += to_text(submission->get(i)) + " ";
+            auto text = [&] {
+              if constexpr(std::is_same_v<TagListBox, AccountListBox>) {
+                return QString::fromStdString(
```
kranar: Why is this being special cased?

shaobinspire: I think these two questions can be answered together.

Since `AccountListItem::Account` does not have a `to_text` overload, `AccountListBox` cannot be implemented as a type alias like `TickerListBox`, `VenueListBox`, or `TradingGroupListBox` (They are all `using ... = TagComboBox<T>` aliases). Instead, it has to be implemented as a `QWidget` subclass, similar to `ScopeBox`.

Because `AccountListBox` is not a `TagComboBox<T>`, the existing `OpenFilterPanelAdaptor<TagComboBox<T>>` partial specialization can't be reused. As a result, `OpenFilterPanelAdaptor<AccountListBox>` had to be implemented even though there is no account-specific behavior. It also requires handling the submission output using `if constexpr` branch.

I propose adding a `to_text(AccountListItem::Account)` overload. This would allow `AccountListBox` to be implemented as a type alias and enable `AccountFilterPanel` to be built without specializing `OpenFilterPanelAdaptor`.

kranar: Okay add the ```to_text``` overload.

PR #1335, opened 2026-06-12
Reimplemented AccountListBox as a TagComboBox alias.

shaobinspire:
Since `AccountListItem::Account` now has a `to_text(AccountListItem::Account, ...)` overload, `AccountListBox` no longer requires a dedicated widget implementation and has been updated to a type alias with corresponding factory functions.

### DIFF
--- a/Applications/Spire/Include/Spire/Ui/CustomQtVariants.hpp
+++ b/Applications/Spire/Include/Spire/Ui/CustomQtVariants.hpp
@@ -21,7 +21,6 @@
 #include "Nexus/Definitions/Venue.hpp"
 #include "Spire/Canvas/Tasks/Task.hpp"
 #include "Spire/Spire/AnyRef.hpp"
-#include "Spire/Ui/AccountListItem.hpp"
 
 namespace Spire {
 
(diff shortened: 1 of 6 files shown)


Applications/Spire/Include/Spire/Ui/CustomQtVariants.hpp
```diff
@@ -121,6 +116,10 @@ namespace Spire {
   QString to_text(
     boost::posix_time::time_duration time, const QLocale& locale = QLocale());
 
+  /** Returns the text representation of an Account. */
+  QString to_text(const AccountListItem::Account& account,
```
kranar: Avoid spreading header files everywhere, it ends up bloating the dependency chain. Now anytime any change is made to ```AccountListItem```, almost every translation unit needs to get rebuilt even if it doesn't have anything to do with ```AccountListItem```. Structure dependencies to minimize coupling.

Applications/Spire/Source/Ui/CustomQtVariants.cpp
```diff
@@ -278,6 +279,11 @@ QString Spire::to_text(posix_time::time_duration time, const QLocale& locale) {
   return QString::fromStdString(to_simple_string(time));
 }
 
+QString Spire::to_text(
+    const AccountListItem::Account& account, const QLocale& locale) {
+  return QString::fromStdString(account.m_account.m_name);
```
kranar: Please keep things together, it's never acceptable to declare a function in one header file and to implement it in a completely unrelated source file that has nothing to do with the header file declaring the function.

Not only will this result in linker errors in situations where this translation unit is not used, it makes it incredibly hard to find definitions for things declared in a particular header file.

In general, avoid smearing the codebase with implementation details, keep code local and isolated.

